PR #222, opened 2026-02-21
[Customer Portal][BE]Update case creation and project responses with parentCase and chat support

cloby99:
- Purpose
This PR enhances case creation and related response payloads to support parent-child case relationships and chat references.

- Changes
Updated `CaseCreatePayload` to accept `parentCaseId` and `chatId` fields
Added related `parentCase` and `chat` fields to the case record
Updated `callRequestCreatePayload` response
Updated `project` response payload structure

## Release Notes

* **New Features**
  * Cases can now reference parent cases and related chats for better case organization and tracking.
  * Project responses now include detailed account information.
  * Call request reason field is now optional.


Rashmika998 (approved):
LGTM

PR #475, opened 2026-04-09
[Customer Portal][BE] Update openapi.yaml and call request update validation

Rashmika998:
## Description
- Update call request update validation.
- Update `openapi.yaml` file.

## Summary by CodeRabbit

* **New Features**
  * Allow specifying call request duration when updating a request.

* **Bug Fixes**
  * Validation now enforces duration > 0 when provided.
  * Improved validation for “customer rejected” state: utcTimes, duration and cancellation reason are treated as a single mutually‑exclusive group (single combined error).
  * API schema: removed empty-string defaults for some nullable account fields.

### DIFF
--- a/apps/customer-portal/backend/modules/entity/utils.bal
+++ b/apps/customer-portal/backend/modules/entity/utils.bal
@@ -88,6 +88,10 @@ public isolated function validateCallRequestUpdatePayload(CallRequestUpdatePaylo
     string? cancellationReason = payload?.cancellationReason;
     boolean hasUtcTimes = utcTimes !is () && utcTimes.length() > 0;
 
+    if durationInMinutes is int && durationInMinutes <= 0 {
+        return "Duration should be greater than 0.";
+    }
+
     if stateKey == PENDING_ON_WSO2 {
         if !hasUtcTimes && durationInMinutes is () {
             return "Either UTC times or duration should be provided when the status is Pending on WSO2.";
--- a/apps/customer-portal/backend/modules/entity/utils.bal
+++ b/apps/customer-portal/backend/modules/entity/utils.bal
@@ -88,23 +88,23 @@ public isolated function validateCallRequestUpdatePayload(CallRequestUpdatePaylo
     string? cancellationReason = payload?.cancellationReason;
     boolean hasUtcTimes = utcTimes !is () && utcTimes.length() > 0;
 
-    if durationInMinutes is int && durationInMinutes <= 0 {
-        return "Duration should be greater than 0.";
-    }
-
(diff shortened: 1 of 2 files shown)


apps/customer-portal/backend/openapi.yaml
```diff
@@ -2843,6 +2841,9 @@ components:
             is 2)
           items:
             $ref: '#/components/schemas/DateTime'
+        durationInMinutes:
+          type: integer
+          description: New duration in minutes for the call (mandatory when stateKey is 2)
```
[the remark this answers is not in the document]
Rashmika998: if the state key is 2, durationInMinutes is required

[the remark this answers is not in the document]
Rashmika998: Updated

apps/customer-portal/backend/modules/entity/utils.bal
```diff
@@ -92,15 +92,16 @@ public isolated function validateCallRequestUpdatePayload(CallRequestUpdatePaylo
         if !hasUtcTimes && durationInMinutes is () {
             return "Either UTC times or duration should be provided when the status is Pending on WSO2.";
         }
+        if durationInMinutes is int && durationInMinutes <= 0 {
```
cloby99, line 95:
```suggestion
        durationInMinutes <= 0 {
```
I believe the type validation for `durationInMinutes` is already handled by the `CallRequestUpdatePayload` record

apps/customer-portal/backend/openapi.yaml
```diff
@@ -2839,10 +2837,13 @@ components:
           description: Reason for the requested call cancellation
         utcTimes:
           type: array
-          description: New preferred UTC times for the call (mandatory when stateKey
-            is 2)
+          description: New preferred UTC times for the call (mandatory when stateKey is 2)
```
cloby99, line 2840:
```suggestion
          description: New preferred UTC times for the call
```

apps/customer-portal/backend/modules/entity/utils.bal
Rashmika998: We have to check whether the duration is <= 0 only when it's provided in the input(can be null). So need to check whether the type is `int` before checking <= 0.

`int? durationInMinutes = payload.durationInMinutes;`

apps/customer-portal/backend/openapi.yaml
Rashmika998: This is updated.

cloby99 (approved):

PR #533, opened 2026-04-21
[Customer Portal][BE] Add project features

Rashmika998:
## Summary
This PR introduces support for project-specific features by adding an endpoint to return enabled features for a given project.

## Changes
- Added endpoint to fetch features for a project (`GET /projects/{id}/features`)
- Implemented logic to resolve features based on project configuration/type
- Added response model for feature flags
- Integrated validation for project ID and access checks
- Updated service-layer mappings and serialization

## Functionality
The endpoint returns feature flags specific to a project, enabling:
- Dynamic enable/disable of frontend features
- Project-type-based capability control
- Reduced need for hardcoded feature toggles

## Reason
Different projects require different capabilities.  
Providing project-level feature flags enables:

- Fine-grained feature control
- Safer feature rollouts
- Improved UX by showing only relevant functionality
- Reduced frontend complexity

## Impact
- New endpoint added (non-breaking enhancement)
- No changes to existing APIs
- Enhances feature management per project

## Testing
- Verified feature flags returned for different project types
- Tested scenarios with no features / all features enabled
- Confirmed proper validation for invalid project IDs

## Summary by CodeRabbit

* **New Features**
  * Added project feature access configuration with granular read/write controls for service requests, SRA, change requests, engagements, updates, time logs, and deployments.
  * Added a new project-level "features" endpoint so clients can retrieve feature access details for a project.
* **Bug Fixes / Improvements**
  * Improved project filter handling to return clearer "not found" responses when a project is missing or inaccessible.

### DIFF
--- a/apps/customer-portal/backend/modules/entity/types.bal
+++ b/apps/customer-portal/backend/modules/entity/types.bal
@@ -561,10 +561,8 @@ public type SortBy record {|
 
 # Project feature access configuration.
 public type ProjectFeatures record {|
-    # Feature ID
-    IdString projectTypeId;
-    # Feature name
-    string projectType;
+    # Project type information
+    ReferenceTableItem projectType;
     # Severities available for the feature
     ChoiceListItem[] acceptedSeverityValues;
     # Indicates if service request write access is enabled
--- a/apps/customer-portal/backend/modules/entity/types.bal
+++ b/apps/customer-portal/backend/modules/entity/types.bal
@@ -559,6 +559,37 @@ public type SortBy record {|
     json...;
 |};
 
+# Project feature access configuration.
+public type ProjectFeatures record {|
+    # Feature ID
+    IdString projectTypeId;
+    # Feature name
+    string projectType;
+    # Severities available for the feature
+    ChoiceListItem[] acceptedSeverityValues;
+    # Indicates if service request write access is enabled
+    boolean hasServiceRequestWriteAccess;
+    # Indicates if service request read access is enabled
+    boolean hasServiceRequestReadAccess;
+    # Indicates if SRA write access is enabled
+    boolean hasSraWriteAccess;
+    # Indicates if SRA read access is enabled
+    boolean hasSraReadAccess;
+    # Indicates if change request read access is enabled
+    boolean hasChangeRequestReadAccess;
+    # Indicates if engagements read access is enabled
+    boolean hasEngagementsReadAccess;
+    # Indicates if updates read access is enabled
+    boolean hasUpdatesReadAccess;
+    # Indicates if time logs read access is enabled
+    boolean hasTimeLogsReadAccess;
+    # Indicates if deployment write access is enabled
+    boolean hasDeploymentWriteAccess;
+    # Indicates if deployment read access is enabled
+    boolean hasDeploymentReadAccess;
+    json...;
+|};
+
 # Project metadata response.
 public type ProjectMetadataResponse record {|
     # List of available case states (eg: Open, Closed, etc.)
@@ -587,6 +618,8 @@ public type ProjectMetadataResponse record {|
     ChoiceListItem[] engagementPaymentTypes;
     # Severity based allocation time mapping (severity ID to allocation time in minutes)
     map<int> severityBasedAllocationTime;
+    # Feature access configuration for the project
+    ProjectFeatures features;
     json...;
 |};
 
--- a/apps/customer-portal/backend/modules/types/types.bal
+++ b/apps/customer-portal/backend/modules/types/types.bal
@@ -308,8 +308,6 @@ public type ProjectFilterOptions record {|
     ReferenceItem[] engagementPaymentTypes;
     # Severity based allocation time mapping (severity ID to allocation time in minutes)
     map<int> severityBasedAllocationTime;
-    # Feature access configuration for the project
-    ProjectFeatures features;
 
 |};
 
--- a/apps/customer-portal/backend/modules/types/types.bal
+++ b/apps/customer-portal/backend/modules/types/types.bal
@@ -254,6 +254,32 @@ public type UpdatedUser record {|
     string timeZone?;
 |};
 
+# Project feature access configuration.
+public type ProjectFeatures record {|
+    # Severities available for the feature
+    ReferenceItem[] acceptedSeverityValues;
+    # Indicates if service request write access is enabled
+    boolean hasServiceRequestWriteAccess;
+    # Indicates if service request read access is enabled
+    boolean hasServiceRequestReadAccess;
+    # Indicates if SRA write access is enabled
+    boolean hasSraWriteAccess;
+    # Indicates if SRA read access is enabled
+    boolean hasSraReadAccess;
+    # Indicates if change request read access is enabled
+    boolean hasChangeRequestReadAccess;
+    # Indicates if engagements read access is enabled
+    boolean hasEngagementsReadAccess;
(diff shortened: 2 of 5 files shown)


apps/customer-portal/backend/openapi.yaml
```diff
@@ -4819,6 +4872,8 @@ components:
             format: int64
           description: Severity based allocation time mapping (severity ID to allocation
             time in minutes)
+        features:
+          $ref: '#/components/schemas/ProjectFeatures'
```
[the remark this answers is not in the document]
Rashmika998: In the `/projects/{id}/filters` response, features are also returned to the FE.

apps/customer-portal/backend/utils.bal
```diff
@@ -132,6 +132,9 @@ public isolated function getProjectFilters(entity:ProjectMetadataResponse projec
         select {id: item.id.toString(), label: item.label};
     types:ReferenceItem[] engagementPaymentTypes = from entity:ChoiceListItem item in projectMetadata.engagementPaymentTypes
         select {id: item.id.toString(), label: item.label};
+    types:ReferenceItem[] acceptedSeverityValues =
+        from entity:ChoiceListItem item in projectMetadata.features.acceptedSeverityValues
+    select {id: item.id.toString(), label: item.label};
```
shayanmalinda: **💡 Nitpick: Query expression `select` clause is misaligned**

The `select` keyword here is indented at the same level as the variable declaration (4 spaces), while every other query expression in this function indents `select` 4 spaces deeper — making clause boundaries visually clear. This inconsistency can make the code harder to scan at a glance.

Suggested fix: indent `select` to 8 spaces to match the surrounding style:
```ballerina
types:ReferenceItem[] acceptedSeverityValues =
    from entity:ChoiceListItem item in projectMetadata.features.acceptedSeverityValues
        select {id: item.id.toString(), label: item.label};
```
Running `bal format` before committing will catch and fix this automatically.

> 📖 **Reference ([Ballerina Best Practices](https://learn-ballerina.github.io)):**
> "Run `bal format` before committing"

---

apps/customer-portal/backend/modules/entity/types.bal
```diff
@@ -559,6 +559,35 @@ public type SortBy record {|
     json...;
 |};
 
+# Project feature access configuration.
+public type ProjectFeatures record {|
+    # Project type information
+    ReferenceTableItem projectType;
+    # Severities available for the feature
```
shayanmalinda, line 566: **🏗️ Architectural Question: `projectType` field is present in entity but dropped in the transformation**

The `ProjectFeatures` record in the entity module includes `projectType: ReferenceTableItem`, but this field is not mapped in `utils.bal` and does not appear in the `types:ProjectFeatures` record or the OpenAPI schema. If `projectType` carries meaningful data for the frontend, it would be silently discarded during the entity-to-types transformation.

No action needed if `projectType` is already available elsewhere in the response — but a short comment explaining why it is excluded from the output would help future readers understand the intent.

---

apps/customer-portal/backend/modules/types/types.bal
```diff
@@ -282,6 +308,9 @@ public type ProjectFilterOptions record {|
     ReferenceItem[] engagementPaymentTypes;
     # Severity based allocation time mapping (severity ID to allocation time in minutes)
     map<int> severityBasedAllocationTime;
+    # Feature access configuration for the project
+    ProjectFeatures features;
```
shayanmalinda: **🏗️ Architectural Question: `ProjectFeatures` embedded inside `ProjectFilterOptions`**

`ProjectFilterOptions` currently holds filter UI data — dropdown values like case states, severities, and engagement types. `ProjectFeatures` is access-control data — boolean flags governing which capabilities a user has for a given project. These two concepts serve different purposes, and mixing them into one response object could make the API harder to reason about as it grows (a client fetching filter options doesn't necessarily need feature flags, and vice versa).

The PR description mentions a new `GET /projects/{id}/features` endpoint. Is this endpoint returning a dedicated `ProjectFeatures` response, or is it returning the full `ProjectFilterOptions` object? If feature flags are bundled with filter options for a single round-trip, that may be a deliberate choice — but it would be worth a comment or API description note to make the reasoning clear.

---

apps/customer-portal/backend/utils.bal
Rashmika998: This is formatted by Ballerina formatter

apps/customer-portal/backend/modules/entity/types.bal
Rashmika998: It's not needed to return to the FE as type is already returned by a separate endpoint

shayanmalinda (dismissed):

apps/customer-portal/backend/modules/types/types.bal
Rashmika998: Yes, I'll change it

apps/customer-portal/backend/openapi.yaml
```diff
@@ -778,6 +778,36 @@ paths:
...
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ProjectFeatures'
+        "401":
+          description: Unauthorized
+        "500":
+          description: InternalServerError
+        "403":
+          description: Forbidden
+        "400":
```
shayanmalinda, line 809: **⚠️ Major: OpenAPI spec documents a 400 response that the implementation can never return**

The spec says this endpoint can return `400 BadRequest`, but the Ballerina function signature is `returns types:ProjectFeatures|http:Unauthorized|http:Forbidden|http:InternalServerError` — there is no `http:BadRequest` in the return type. This means the runtime cannot produce a 400, so the spec is lying to API consumers. Any client that handles 400 for bad project IDs (e.g. malformed ID format) will never see it — they'll get a 500 instead.

Suggested fix: Either add `http:BadRequest` to the return type and return it when the project ID is invalid, or remove the `"400"` entry from the spec to match what the code actually does.

---

apps/customer-portal/backend/service.bal
```diff
@@ -1414,6 +1414,55 @@ service http:InterceptableService / on new http:Listener(9090, listenerConf) {
...
+            if getStatusCode(projectMetadata) == http:STATUS_FORBIDDEN {
+                log:printWarn(string `User: ${userInfo.userId} is forbidden to access project features for project: ${
+                        id}`);
+                return <http:Forbidden>{
+                    body: {
+                        message: "You're not authorized to access the features for the selected project."
+                    }
+                };
+            }
+
```
shayanmalinda, line 1462: **⚠️ Major: Non-existent project returns 500 instead of 404**

The error handler explicitly checks for 401 and 403, but anything else — including a 404 from the entity layer when the project ID doesn't exist — falls through to the generic `InternalServerError` catch-all. A caller that passes a valid but non-existent project ID will receive a 500, which is misleading and makes it hard to distinguish "server broke" from "project not found".

Suggested fix: add a 404 check before the catch-all:
```ballerina
if getStatusCode(projectMetadata) == http:STATUS_NOT_FOUND {
    return <http:NotFound>{};
}
```
And add `http:NotFound` to the function return type and OpenAPI spec.

---

shayanmalinda: Not sure if this is valid. Better to check with the entity implementation.

Rashmika998: Added

shayanmalinda (approved):

PR #538, opened 2026-04-21
[Customer Portal][BE] Filter restricted change request states

Rashmika998:
## Description
- Filter restricted change request states as they don't want to show in the UI.

## Summary by CodeRabbit

* **New Features**
  * Added the ability to restrict specific change request states from appearing in project filters, providing greater control over which states are available to users.

### DIFF
--- a/apps/customer-portal/backend/utils.bal
+++ b/apps/customer-portal/backend/utils.bal
@@ -23,6 +23,7 @@ configurable int stateIdOpen = 1;
 configurable types:FeatureFlags featureFlags = {
     usageMetricsEnabled: true
 };
+configurable string[] restrictedChangeRequestStateIds = ["-3", "-4", "-5"];
 
 # Search cases for a given project.
 #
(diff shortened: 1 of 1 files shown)


apps/customer-portal/backend/utils.bal
```diff
@@ -23,6 +23,7 @@ configurable int stateIdOpen = 1;
 configurable types:FeatureFlags featureFlags = {
     usageMetricsEnabled: true
 };
+configurable string[] restrictedChangeRequestStateIds = ["-3", "-4", "-5"];
```
shayanmalinda, line 26: **📝 Nitpick: No documentation on what the restricted state IDs represent**

The default values `["-3", "-4", "-5"]` are magic numbers — it is not immediately clear which named states these IDs correspond to (e.g. "Cancelled", "Rejected", "Withdrawn"). A future maintainer looking at this will have to dig through the external system to figure out why exactly these three values are excluded. Adding a brief doc comment above this configurable would save that effort and make the intent explicit.

Suggested fix: Add a doc comment above the configurable, e.g.:
```ballerina
# State IDs that are not shown in the UI. Defaults exclude: Cancelled (-3), Rejected (-4), Withdrawn (-5).
configurable string[] restrictedChangeRequestStateIds = ["-3", "-4", "-5"];
```

---

shayanmalinda (approved):

PR #589, opened 2026-04-24
Add membership role support and user management types

agzaiyenth:
This pull request refactors and extends the user management types and logic to support more granular role handling for project contacts and memberships. The main changes include introducing explicit role fields (such as portal user and security contact), updating payloads and service interfaces to use roles instead of simple flags, and adding conversion utilities for mapping between user management service data and internal types.

**Role and Type Model Enhancements:**

* Added `isPortalUser` and `isCsAdmin` fields to `Contact`, `ContactOnboardPayload`, and related types, and introduced new types (`UserManagementContact`, `UserManagementOnBoardContactPayload`, `UserManagementMembership`, `UserManagementMembershipRolePayload`) to better model roles and payloads exchanged with the user management service. [[1]](diffhunk://#diff-373f28cd610162316937021515848b49fd6e883a9b4a5eea33d9ae9d98452a97R43-R44) [[2]](diffhunk://#diff-373f28cd610162316937021515848b49fd6e883a9b4a5eea33d9ae9d98452a97R54-R60) [[3]](diffhunk://#diff-373f28cd610162316937021515848b49fd6e883a9b4a5eea33d9ae9d98452a97R86-R107) [[4]](diffhunk://#diff-373f28cd610162316937021515848b49fd6e883a9b4a5eea33d9ae9d98452a97R116-R131) [[5]](diffhunk://#diff-373f28cd610162316937021515848b49fd6e883a9b4a5eea33d9ae9d98452a97L103-R158) [[6]](diffhunk://#diff-013ef672b53636e334800b67d6889966d840050136ee23ecabfa3662c5e08aeeR1049-R1064)

**API and Service Changes:**

* Changed the service and backend API to use `MembershipRolePayload` instead of `MembershipSecurityPayload`, and updated the PATCH endpoint and related service logic to handle roles rather than just a security flag. [[1]](diffhunk://#diff-eca07fe60f4c779acb6655f1aa59203532751ff527a16d51de371ef62ea3259eL3469-R3471) [[2]](diffhunk://#diff-eca07fe60f4c779acb6655f1aa59203532751ff527a16d51de371ef62ea3259eL3510-R3516) [[3]](diffhunk://#diff-f2fef0627c35b6c7b8d22030310a3f366114d276ecbac9c1d7651e284efc7fb4L72-R108)

**Role Handling Logic:**

* Added utility functions (`getRoles`, `toContact`, `toMembership`, `hasRole`) to map between boolean fields and role lists, and to convert user management service responses into internal types.

**Contact and Membership Creation/Update:**

* Updated contact creation and membership update logic to use the new role-based payloads and mapping functions, ensuring that role information is correctly passed to and from the user management service. [[1]](diffhunk://#diff-f2fef0627c35b6c7b8d22030310a3f366114d276ecbac9c1d7651e284efc7fb4L39-R53) [[2]](diffhunk://#diff-f2fef0627c35b6c7b8d22030310a3f366114d276ecbac9c1d7651e284efc7fb4L48-R63) [[3]](diffhunk://#diff-f2fef0627c35b6c7b8d22030310a3f366114d276ecbac9c1d7651e284efc7fb4L95-R118) [[4]](diffhunk://#diff-f2fef0627c35b6c7b8d22030310a3f366114d276ecbac9c1d7651e284efc7fb4L111-R135)

**Constants and Minor Additions:**

* Introduced constants for role names (`PORTAL_USER_ROLE`, `SECURITY_CONTACT_ROLE`) to avoid magic strings and improve maintainability.

These changes provide a more flexible and extensible way to manage user roles in project contacts and memberships, aligning the backend logic with the user management service’s expectations and supporting future role additions.

## Release Notes

* **New Features**
  * Extended contact management with portal user and CS admin role support.
  * Membership role assignment now supports granular role-based indicators for better access control.

* **Updates**
  * Contact creation and role updates now propagate additional role attributes.
  * Improved role management through enhanced payload structures for membership updates.

### DIFF
--- a/apps/customer-portal/backend/modules/user_management/user_management.bal
+++ b/apps/customer-portal/backend/modules/user_management/user_management.bal
@@ -18,12 +18,17 @@ import ballerina/log;
 
 public type CONFLICT_ERROR distinct error;
 
+const PORTAL_USER_ROLE = "Portal user";
+const SECURITY_CONTACT_ROLE = "Security Contact";
+
 # Get project contacts for the given project ID.
 #
 # + projectId - Salesforce ID of the project to get contacts for
 # + return - Array of contacts or error
 public isolated function getProjectContacts(string projectId) returns Contact[]|error {
-    return userManagementClient->/projects/[projectId]/contacts.get();
+    UserManagementContact[] contacts = check userManagementClient->/projects/[projectId]/contacts.get();
+    return from UserManagementContact contact in contacts
+        select toContact(contact);
 }
 
 # Create a new contact for the given project ID and payload.
@@ -36,7 +41,16 @@ public isolated function createProjectContact(string projectId, OnBoardContactPa
 
     string customError = "An error occurred while creating the contact for the project.";
 
-    http:Response userCreateResponse = check userManagementClient->/projects/[projectId]/contact.post(payload);
+    UserManagementOnBoardContactPayload userManagementPayload = {
+        contactEmail: payload.contactEmail,
+        adminEmail: payload.adminEmail,
+        contactFirstName: payload.contactFirstName,
+        contactLastName: payload.contactLastName,
+        isCsIntegrationUser: payload.isCsIntegrationUser,
+        role: getRoles(payload.isPortalUser, payload.isSecurityContact)
+    };
+
+    http:Response userCreateResponse = check userManagementClient->/projects/[projectId]/contact.post(userManagementPayload);
     if userCreateResponse.statusCode != http:STATUS_CREATED {
         json|error errBody = userCreateResponse.getJsonPayload();
         if errBody is error {
@@ -45,7 +59,8 @@ public isolated function createProjectContact(string projectId, OnBoardContactPa
         }
         return error(check errBody.message);
     }
-    return (check userCreateResponse.getJsonPayload()).cloneWithType();
+    UserManagementMembership membership = check (check userCreateResponse.getJsonPayload()).cloneWithType();
+    return toMembership(membership);
 }
 
 # Remove a contact from the given project ID using the contact's email and admin email.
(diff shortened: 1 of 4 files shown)


apps/customer-portal/backend/modules/user_management/user_management.bal
```diff
@@ -129,3 +153,52 @@ public isolated function validateProjectContact(ValidationPayload payload) retur
     log:printError(customError, info = errBody.toString());
     return error(check errBody.message);
 }
+
+isolated function getRoles(boolean isPortalUser, boolean isSecurityContact) returns string[] {
```
shayanmalinda, line 157: **🐛 Bug: `isCsAdmin` is accepted in the public payload but silently dropped**

`getRoles` only takes `isPortalUser` and `isSecurityContact` — `isCsAdmin` from both `MembershipRolePayload` and `OnBoardContactPayload` is never passed in, so any caller setting `isCsAdmin: true` will have it silently ignored. The same applies to the onboarding path: `UserManagementOnBoardContactPayload` has no `isCsAdmin` field, so it's dropped there too. If `isCsAdmin` is not intended to map to a role string, the field probably shouldn't be in the public payloads at all; if it is intended to affect roles, it needs to be handled here.

Suggested fix: Either remove `isCsAdmin` from `MembershipRolePayload` and `OnBoardContactPayload` if it has no effect, or add a constant and handle it in `getRoles`:
```ballerina
const CS_ADMIN_ROLE = "CS Admin"; // replace with the actual role string

isolated function getRoles(boolean isPortalUser, boolean isSecurityContact, boolean isCsAdmin) returns string[] {
    string[] roles = [];
    if isPortalUser { roles.push(PORTAL_USER_ROLE); }
    if isSecurityContact { roles.push(SECURITY_CONTACT_ROLE); }
    if isCsAdmin { roles.push(CS_ADMIN_ROLE); }
    return roles;
}
```

---

apps/customer-portal/backend/modules/user_management/user_management.bal
```diff
@@ -18,12 +18,17 @@ import ballerina/log;
 
 public type CONFLICT_ERROR distinct error;
 
+const PORTAL_USER_ROLE = "Portal user";
+const SECURITY_CONTACT_ROLE = "Security Contact";
```
shayanmalinda, line 22: **⚠️ Potential Bug: Inconsistent casing between role constants**

`PORTAL_USER_ROLE` is `"Portal user"` (lowercase 'u') while `SECURITY_CONTACT_ROLE` is `"Security Contact"` (both words capitalised). Since `hasRole` does an exact string comparison after trimming, if either of these doesn't exactly match what the user management service stores in the `role` field, it will always silently return `false` for that role. It's worth double-checking both strings against what the upstream service actually returns — a casing mismatch here would be hard to spot in testing.

Suggested fix: Verify both strings against the user management service's role values and ensure they match exactly.

---

shayanmalinda: @agzaiyenth hope the namings are correct?

v15a1 (approved):

PR #685, opened 2026-05-11
feat(customer-portal): restrict POST /users/groups to authorized group

Suwagath:
## Purpose
Restrict the add-users-to-group endpoint to users belonging to a configured group. Previously any authenticated user could invoke it.

## Approach
- Added `scimGroupManagementRole` to `AppRoles` configurable record
- Check `userInfo.groups` against the configured role using existing `checkRoles` utility
- Return 403 Forbidden if the user is not in the allowed group

## Summary by CodeRabbit

* **Chores**
  * Disabled the add users to groups functionality from the API.
  * Updated authorization roles configuration in the backend.

[Review Change Stack](https://app.coderabbit.ai/change-stack/wso2-open-operations/cs-tools/pull/685)


apps/customer-portal/backend/modules/authorization/types.bal
```diff
@@ -54,4 +54,6 @@ public type UserInfoPayload record {|
 public type AppRoles record {|
     # Role for an Admin
     string adminRole;
+    # Allowed group for SCIM group management
+    string scimGroupManagementRole;
```
Rashmika998: I think the name should be relevant to the functionality. Can you explain the use case?

Suwagath: This role restricts access to the POST /users/groups endpoint, which adds users to external groups via the SCIM operations service. Only users belonging to this configured group are authorized to invoke that API.

Since this endpoint could be invoked by many relevant services to customer management, I kept the name to be general to the actual service we are tapping into.

Happy to rename if you could suggest with one :)

apps/customer-portal/backend/service.bal
```diff
@@ -5611,6 +5611,16 @@ service http:InterceptableService / on new http:Listener(9090, listenerConf) {
             };
         }
 
+        string[] userGroups = userInfo.groups ?: [];
+        if !authorization:checkRoles([authorization:authorizedRoles.scimGroupManagementRole], userGroups) {
+            log:printWarn(string `User: ${userInfo.userId} is not authorized to manage group users.`);
```
Rashmika998:
```suggestion
            log:printWarn(string `User: ${userInfo.userId} is not authorized invoke group management endpoint.`);
```

Rashmika998 (approved):

PR #695, opened 2026-05-14
[Micro App] Inline Comment Image Rendering, Project Context Name Resolution, and Pie Chart Empty State

lithika-damnod:
This minor release fixes inline image rendering within comments, corrects project field naming, and improves the pie chart empty state.

### 1. Inline Comment Image Rendering
Inline images within comments are now fetched through an authenticated request channel. Previously, image requests were made without the necessary auth headers, causing them to fail to render in environments that require token-based access.

### 2. Project Context Name Resolution
The project field now correctly displays the name from the active project context rather than the name returned by the fetched operation response. This prevents a label mismatch when the two values differ across sessions.

### 3. Pie Chart Empty State
The pie chart widget's empty state circle now renders an "N/A" label when there is no data to display, making it clear to users that data is absent rather than still loading.

## Summary by CodeRabbit

* **Bug Fixes**
  * Select fields now handle missing options safely.
  * Reduced attachment refetches via longer-lived caching.

* **New Features / UI**
  * Embedded images in rich text now load attachments in-place.
  * Project is auto-selected and disabled on case creation.
  * Smaller, centered empty-state for pie charts; popover scrollbar behavior tuned.

* **Error Handling**
  * Attachment and PDF previews show a friendly error state.

* **Stability**
  * More reliable preview clicks and WebSocket teardown.

[Review Change Stack](https://app.coderabbit.ai/change-stack/wso2-open-operations/cs-tools/pull/695?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


apps/customer-portal/microapp/src/context/me/MeProvider.tsx
```diff
@@ -58,7 +57,8 @@ export default function MeProvider({ children }: { children: React.ReactNode })
     }
   }, [initialized, lastVisitedProject, setProjectId, navigate]);
 
-  if (meLoading || projectsLoading || isFeaturesLoading || isFeaturesError) return <LoadingFallback />;
+  if (meLoading || projectsLoading || (lastVisitedProject && isFeaturesError) || isFeaturesLoading)
+    return <LoadingFallback />;
```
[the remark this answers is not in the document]
lithika-damnod: This is intentional. The project-level feature set is required before rendering any project-scoped UI, since several views depend on those flags for correct initialization and access control.

Because of that, we intentionally block rendering not only while the feature request is in flight, but also when it fails. Rendering without the resolved feature state can lead to inconsistent project behavior, so the loading fallback here is acting as a safeguard until that dependency is available.

apps/customer-portal/microapp/src/pages/CaseDetailPage.tsx
```diff
@@ -320,39 +320,37 @@ function AttachmentCard({
...
-              {attachment.fileName}
-            </Typography>
-            <Typography variant="caption" color="text.secondary" noWrap>
-              {attachment.createdBy} · {fromNow(attachment.createdOn)}
-            </Typography>
-          </Stack>
-          <IconButton onClick={() => onPreview(attachment)}>
-            <ArrowUpRight size={pxToRem(18)} />
-          </IconButton>
+    <Card sx={{ p: 1.5 }} onClick={() => onPreview(attachment)}>
```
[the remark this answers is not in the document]
lithika-damnod: This interaction is currently only used in the mobile attachment view, where keyboard navigation isn’t part of the expected interaction pattern. That said, I’ve added `role="button"` to make the intent explicit.

I didn’t add `tabIndex` and `onKeyDown()` here since this is not intended to behave as a fully keyboard-interactive control in the current mobile context, and the existing UX scope doesn’t require desktop-style keyboard accessibility for this element.

apps/customer-portal/microapp/package.json
```diff
@@ -13,13 +13,11 @@
     "prettier:check": "prettier --check ."
   },
   "dependencies": {
-    "@cyntler/react-doc-viewer": "^1.17.1",
     "@emotion/react": "^11.14.0",
```
Rashmika998: As mentioned by security team, let's remove the ^ from all the dependencies
```suggestion
    "@emotion/react": "11.14.0",
```

apps/customer-portal/microapp/src/components/core/AppBar.tsx
```diff
@@ -115,11 +115,12 @@ export function AppBar() {
         {config.showProjectSelector && (
           <Button sx={{ justifyContent: "space-between", p: 0, mt: 2 }} onClick={openProjectSelector} disableRipple>
             <Stack direction="row" sx={{ alignItems: "center", flexGrow: 1, minWidth: 0, gap: 1 }}>
-              <Folder color={theme.palette.text.secondary} size={pxToRem(18)} />
+              <Folder color={theme.palette.text.secondary} size={pxToRem(18)} style={{ flexShrink: 0 }} />
               <Typography variant="body1" color="text.secondary" sx={{ textTransform: "initial" }} noWrap>
                 {projectName}
               </Typography>
             </Stack>
+
```
cloby99, line 123:
```suggestion
```

apps/customer-portal/microapp/src/components/shared/RichText.tsx
```diff
@@ -52,10 +54,13 @@ export const RichTextBase = styled(Box)(({ theme }) => ({
     fontFamily: "inherit !important",
     lineHeight: "inherit !important",
   },
+
```
cloby99, line 57:
```suggestion
```

Rashmika998 (approved):

PR #730, opened 2026-05-26
[CSM-Portal][BE] Add user search functionality

Rashmika998:
## Summary
- Adds SearchUsers method to the entity client in internal/entity/entity.go, proxying POST /users/search to the entity service
- Extends UsersHandler with an entityUserClient interface and a SearchUsers HTTP handler (POST /users/search), following the same pattern as CaseHandler
- Wires the entity client into NewUsersHandler in main.go and registers the route
- Documents the endpoint in `openapi.yaml` with UserSearchPayload, UserSearchResponse, and User schemas matching the entity service contract

## Notes
- The response currently proxies all fields from the entity service. Field filtering to a frontend-specific shape is left as a TODO, consistent with how case endpoints are handled today.
- `UserSearchPayload.pagination` reuses the existing Pagination schema (limit, offset).
- The entity service applies a default limit of 20 and a cap of 100 — validation is handled upstream; the portal layer adds no additional pagination constraints for now.

## Test plan
 - POST /users/search with a valid JWT returns 200 and a paginated user list
 - Request without a JWT returns 401
 - Request body exceeding 1 MiB returns 413
 - `searchQuery` filters results by firstName, lastName, userName, email
 - `userType` filter (`internal` / `customer`) scopes results correctly
 - `hasMore` is `true` when more pages exist

## Summary by CodeRabbit

* **New Features**
  * Added a new user search endpoint enabling authenticated users to search and discover other users with support for filtering by search query, user type, and pagination controls (limit, offset, hasMore).

[Review Change Stack](https://app.coderabbit.ai/change-stack/wso2-open-operations/cs-tools/pull/730?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/apps/csm-portal/backend/internal/handler/users.go
+++ b/apps/csm-portal/backend/internal/handler/users.go
@@ -35,14 +35,20 @@ type scimClient interface {
 	UpdateUserPhone(ctx context.Context, userID, mobile string) (*string, error)
 }
 
-// UsersHandler handles HTTP requests for user-related operations.
-type UsersHandler struct {
-	scim scimClient
+// entityUserClient abstracts the entity service user operations used by UsersHandler.
+type entityUserClient interface {
+	SearchUsers(ctx context.Context, body []byte) ([]byte, error)
 }
 
-// NewUsersHandler creates a UsersHandler backed by the given SCIM client.
-func NewUsersHandler(scim scimClient) *UsersHandler {
-	return &UsersHandler{scim: scim}
+// UsersHandler handles HTTP requests for user-related operations.
+type UsersHandler struct {
+	scim   scimClient
+	entity entityUserClient
+}
+
+// NewUsersHandler creates a UsersHandler backed by the given SCIM and entity clients.
+func NewUsersHandler(scim scimClient, entity entityUserClient) *UsersHandler {
+	return &UsersHandler{scim: scim, entity: entity}
 }
 
 // userMeResponse is the GET /users/me response shape.
@@ -150,4 +156,36 @@ func (h *UsersHandler) PatchMe(w http.ResponseWriter, r *http.Request) {
 	// TODO: update timeZone via entity once available.
 
 	writeJSONValue(w, http.StatusOK, resp)
+}
+
+// SearchUsers handles POST /users/search.
+func (h *UsersHandler) SearchUsers(w http.ResponseWriter, r *http.Request) {
+	user := middleware.UserInfoFromContext(r.Context())
+	if user == nil {
+		writeError(w, http.StatusUnauthorized, ErrMsgUnauthorized)
+		return
+	}
+
+	// TODO: Decode and validate the request body before forwarding.
+
+	r.Body = http.MaxBytesReader(w, r.Body, maxRequestBodyBytes)
+	body, err := io.ReadAll(r.Body)
+	if err != nil {
+		if _, ok := err.(*http.MaxBytesError); ok {
+			writeError(w, http.StatusRequestEntityTooLarge, ErrMsgTooLarge)
+			return
+		}
+		writeError(w, http.StatusBadRequest, errMsgReadBody)
+		return
+	}
+
+	result, err := h.entity.SearchUsers(r.Context(), body)
+	if err != nil {
+		slog.ErrorContext(r.Context(), "entity SearchUsers failed", "userID", user.UserID, "err", err)
+		mapUpstreamError(w, err, "Failed to search users.")
+		return
+	}
+
+	// TODO: Unmarshal result and filter to only the fields required by the frontend.
+	writeJSON(w, http.StatusOK, result)
 }
--- a/apps/csm-portal/backend/internal/handler/users.go
+++ b/apps/csm-portal/backend/internal/handler/users.go
@@ -166,8 +166,6 @@ func (h *UsersHandler) SearchUsers(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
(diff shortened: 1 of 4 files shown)


apps/csm-portal/backend/openapi.yaml
```diff
@@ -355,6 +394,63 @@ components:
           type: string
         # TODO: timeZone — requires entity
 
+    UserSearchPayload:
+      type: object
+      properties:
+        pagination:
+          $ref: '#/components/schemas/Pagination'
+        searchQuery:
```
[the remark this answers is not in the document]
Rashmika998: Updated

apps/csm-portal/backend/internal/handler/users.go
```diff
@@ -151,3 +157,35 @@ func (h *UsersHandler) PatchMe(w http.ResponseWriter, r *http.Request) {
...
+
+	result, err := h.entity.SearchUsers(r.Context(), body)
+	if err != nil {
+		slog.ErrorContext(r.Context(), "entity SearchUsers failed", "userID", user.UserID, "err", err)
+		mapUpstreamError(w, err, "Failed to search users.")
+		return
+	}
+
+	// TODO: Unmarshal result and filter to only the fields required by the frontend.
+	writeJSON(w, http.StatusOK, result)
```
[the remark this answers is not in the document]
Rashmika998: This is not needed for now

apps/csm-portal/backend/internal/handler/users.go
```diff
@@ -151,3 +157,35 @@ func (h *UsersHandler) PatchMe(w http.ResponseWriter, r *http.Request) {
 
 	writeJSONValue(w, http.StatusOK, resp)
 }
+
+// SearchUsers handles POST /users/search.
+func (h *UsersHandler) SearchUsers(w http.ResponseWriter, r *http.Request) {
```
kasunsiyambalapitiya, line 162: **[High] — Security**

`POST /users/search` is gated only by authentication, not authorization. Any valid JWT holder — including customer-type users — can call this endpoint and enumerate all users in the system.

Consider adding a role or user-type guard before processing the request:
```go
if user.UserType != "internal" {
    writeError(w, http.StatusForbidden, ErrMsgForbidden)
    return
}
```
If the intent is that customers can also search, document that explicitly and ensure the entity service enforces result scoping per caller identity.

apps/csm-portal/backend/internal/handler/users.go
```diff
@@ -151,3 +157,35 @@ func (h *UsersHandler) PatchMe(w http.ResponseWriter, r *http.Request) {
...
+	body, err := io.ReadAll(r.Body)
+	if err != nil {
+		if _, ok := err.(*http.MaxBytesError); ok {
+			writeError(w, http.StatusRequestEntityTooLarge, ErrMsgTooLarge)
+			return
+		}
+		writeError(w, http.StatusBadRequest, errMsgReadBody)
+		return
+	}
+
```
kasunsiyambalapitiya, line 179: **[Medium] — Error Handling**

The raw request body is forwarded to the entity service without JSON validation. If the client sends malformed JSON, the entity service returns a 4xx that `mapUpstreamError` may translate to a 500 from the portal's perspective, leaking an opaque error to the caller.

The TODO comment above acknowledges this — treat it as a **blocking** TODO before this endpoint ships to production. At minimum, `json.Valid(body)` should gate the forward:
```go
if !json.Valid(body) {
    writeError(w, http.StatusBadRequest, ErrMsgBadRequest)
    return
}
```

apps/csm-portal/backend/internal/handler/users.go
```diff
@@ -151,3 +157,35 @@ func (h *UsersHandler) PatchMe(w http.ResponseWriter, r *http.Request) {
...
+	result, err := h.entity.SearchUsers(r.Context(), body)
+	if err != nil {
+		slog.ErrorContext(r.Context(), "entity SearchUsers failed", "userID", user.UserID, "err", err)
+		mapUpstreamError(w, err, "Failed to search users.")
+		return
+	}
+
+	// TODO: Unmarshal result and filter to only the fields required by the frontend.
```
kasunsiyambalapitiya, line 194: **[Low] — Types**

`result` is `[]byte` (raw JSON from the entity service). If `writeJSON` calls `json.Marshal` internally, passing a `[]byte` value will base64-encode it rather than write it as raw JSON — the client would receive a JSON string instead of a JSON object.

Verify that `writeJSON` writes raw bytes directly (e.g., via `w.Write(result)`) rather than marshalling. If there is a sibling that explicitly handles raw bytes (e.g., `writeRawJSON`), prefer that for clarity.

apps/csm-portal/backend/internal/handler/users.go
Rashmika998: @kasunsiyambalapitiya This is fine as the app access is limited only for the internal users

apps/csm-portal/backend/internal/handler/users.go
Rashmika998: Fixed

apps/csm-portal/backend/internal/handler/users.go
Rashmika998: `writeJSON` writes the `[]byte` directly via `w.Write(data)` — there's no `json.Marshal` call involved, so no base64 encoding can occur. You can verify in `internal/handler/response.go:51-55`.

_Comment added by Claude on behalf of @Rashmika998_

cloby99 (approved):

PR #731, opened 2026-05-26
[Customer Entity] Add accounts and projects search endpoints

cloby99:
## Purpose
- Adds `POST /accounts/search` endpoint (mirrors the existing user search)
- Renames `entity_*` files to `user_*` for consistency
- Extracts shared request decoding and service error handling into reusable helpers
- Narrows user search to `username` and `email` only (drops `userType` filter and `first`/`last name` matching)
- Splits `000001_create_entities` migration into per-domain files: `000001_create_users`, `000002_create_accounts`, `000003_create_projects`

## Release Notes

* **New Features**
  * Added account search functionality with pagination and filtering support.

* **Bug Fixes**
  * Recovery middleware now returns JSON-formatted error responses for consistency with API standards.

* **Improvements**
  * Enhanced input validation for search queries to enforce length constraints.

[Review Change Stack](https://app.coderabbit.ai/change-stack/wso2-open-operations/cs-tools/pull/731?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


entity-service/internal/repository/account_repo.go
```diff
@@ -0,0 +1,118 @@
...
+
+// SearchAccounts implements AccountRepository.
+func (r *accountRepo) SearchAccounts(ctx context.Context, req domain.SearchAccountsRequest) ([]domain.Account, int, error) {
+	filterArgs := []any{}
+	argIdx := 1
+
+	where := "WHERE 1=1"
+
+	if req.SearchQuery != "" {
+		pattern := "%" + req.SearchQuery + "%"
```
kasunsiyambalapitiya: **[Medium] — Performance / Correctness**

`%` and `_` in `req.SearchQuery` are SQL LIKE metacharacters — they are not escaped before being wrapped in `"%" + req.SearchQuery + "%"`. This means:
- A query of `%` matches every account (full-table scan, O(n) response).
- A query of `50%` matches accounts whose name/sf_id contains `50` followed by anything.
- A query of `abc_def` matches `abc` + any single char + `def`, not the literal string.

Escape before wrapping:
```go
escaped := strings.NewReplacer(`\`, `\\`, `%`, `\%`, `_`, `\_`).Replace(req.SearchQuery)
pattern := "%" + escaped + "%"
```
And pass `ESCAPE '\'` to the `ILIKE` clause, or use `regexp_replace` in PostgreSQL.

entity-service/internal/repository/account_repo.go
```diff
@@ -0,0 +1,118 @@
...
+				&a.OwnerID, &a.TechnicalOwnerID,
+				&a.AgentEnabled, &a.KbReferencesEnabled,
+				&a.CreatedAt, &a.UpdatedAt,
+			); err != nil {
+				return fmt.Errorf("scan account: %w", err)
+			}
+			result = append(result, a)
+		}
+		if err := rows.Err(); err != nil {
+			return err
```
kasunsiyambalapitiya: **[Low] — Error Handling**

`rows.Err()` is not wrapped with context, unlike all other error paths in this function (which use `fmt.Errorf("...: %w", err)`).

```go
if err := rows.Err(); err != nil {
    return fmt.Errorf("iterate accounts: %w", err)
}
```

entity-service/internal/handler/account_handler.go
```diff
@@ -0,0 +1,80 @@
...
+	resp, err := h.svc.SearchAccounts(r.Context(), req)
+	if err != nil {
+		var ve *apierror.ValidationError
+		switch {
+		case errors.As(err, &ve):
+			apierror.WriteJSON(w, http.StatusBadRequest, ve.Error())
+		case errors.Is(err, context.DeadlineExceeded):
+			apierror.WriteJSON(w, http.StatusRequestTimeout, "request timeout")
+		case errors.Is(err, context.Canceled):
+			log.Printf("request canceled: %s %s", r.Method, r.URL.Path)
```
kasunsiyambalapitiya: **[Low] — Naming / Consistency**

The `context.Canceled` case logs before calling `w.WriteHeader(499)`. In the sibling `user_handler.go` (updated in this same PR), the order is reversed: `w.WriteHeader(499)` comes first, then `log.Printf`. Pick one order and apply it consistently across both handlers.

Preferring write-first matches the pattern used everywhere else in these handlers (write the response, then log).

entity-service/internal/middleware/recovery.go
```diff
@@ -17,18 +17,24 @@
...
 // Recovery is an HTTP middleware that catches any panic in a downstream handler,
-// logs it, and writes a 500 response so the server goroutine keeps running.
+// logs it, and writes a JSON 500 response so the server goroutine keeps running.
 func Recovery(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		defer func() {
 			if rec := recover(); rec != nil {
 				log.Printf("panic: %v", rec)
-				http.Error(w, "internal server error", http.StatusInternalServerError)
+				w.Header().Set("Content-Type", "application/json")
```
kasunsiyambalapitiya: **[Low] — HTTP Handlers**

If a handler panics *after* it has already called `w.WriteHeader` (e.g. mid-stream response), Go's `ResponseWriter` makes subsequent `w.Header().Set(...)` and `w.WriteHeader(500)` calls silent no-ops. The `json.NewEncoder(w).Encode(...)` call will then append the JSON error payload to a partially-written, possibly non-JSON response body, corrupting it.

The previous `http.Error` had the same limitation — this isn't a regression, just worth noting since this code was touched. A lightweight mitigation is a `responseStarted` flag on a wrapped `ResponseWriter`, or simply skipping the body write if headers are already sent.

For the common case (handlers that don't stream), the current behavior is correct.

Rashmika998 (approved):

PR #749, opened 2026-05-28
[CSM Portal] Add products and product versions search endpoints

Rashmika998:
## Summary

- Wires \`POST /products/search\` and \`POST /product/{id}/versions/search\` into the csm-portal backend, proxying to the entity service endpoints added in #746
- Adds \`SearchProducts\` and \`SearchProductVersions\` methods to the entity client (\`internal/entity/entity.go\`); \`SearchProductVersions\` accepts the product ID as a path segment
- Adds \`ProductHandler\` in \`internal/handler/products.go\` following the same gateway pattern as \`ProjectHandler\`; \`SearchProductVersions\` extracts \`{id}\` from the path and forwards it to the entity client
- Registers both routes in \`cmd/server/main.go\`
- Updates \`openapi.yaml\` with both new paths and their request/response schemas; \`productId\` is a path parameter on the versions endpoint, not a body field
- Adds full test coverage in \`products_test.go\` and mocks in \`helpers_test.go\`, matching the pattern used by all other handlers
- Fixes \`MaxBytesError\` detection to use \`errors.As\` instead of a direct type assertion (review feedback from @kasunsiyambalapitiya)

## Test plan

- [x] \`go test ./internal/handler/...\` passes locally
- [x] Pre-push hook ran all backend tests — all green
- [x] Verify \`POST /products/search\` proxies to entity service and returns product list
- [x] Verify \`POST /product/{id}/versions/search\` proxies to entity service and returns version list for the given product
- [x] Verify unauthenticated requests return 401
- [x] Verify missing \`{id}\` returns 400
- [x] Verify oversized / invalid JSON bodies return 413 / 400

### DIFF
--- a/apps/csm-portal/backend/internal/handler/products.go
+++ b/apps/csm-portal/backend/internal/handler/products.go
@@ -19,6 +19,7 @@ package handler
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"io"
 	"log/slog"
 	"net/http"
@@ -54,7 +55,8 @@ func (h *ProductHandler) SearchProducts(w http.ResponseWriter, r *http.Request) 
 	r.Body = http.MaxBytesReader(w, r.Body, maxRequestBodyBytes)
 	body, err := io.ReadAll(r.Body)
 	if err != nil {
-		if _, ok := err.(*http.MaxBytesError); ok {
+		var maxBytesErr *http.MaxBytesError
+		if errors.As(err, &maxBytesErr) {
 			writeError(w, http.StatusRequestEntityTooLarge, ErrMsgTooLarge)
 			return
 		}
@@ -91,7 +93,8 @@ func (h *ProductHandler) SearchProductVersions(w http.ResponseWriter, r *http.Re
 	r.Body = http.MaxBytesReader(w, r.Body, maxRequestBodyBytes)
 	body, err := io.ReadAll(r.Body)
 	if err != nil {
-		if _, ok := err.(*http.MaxBytesError); ok {
+		var maxBytesErr *http.MaxBytesError
+		if errors.As(err, &maxBytesErr) {
 			writeError(w, http.StatusRequestEntityTooLarge, ErrMsgTooLarge)
 			return
 		}
--- a/apps/csm-portal/backend/internal/handler/products.go
+++ b/apps/csm-portal/backend/internal/handler/products.go
@@ -30,7 +30,7 @@ import (
 // entityProductClient abstracts the entity service product operations used by ProductHandler.
 type entityProductClient interface {
 	SearchProducts(ctx context.Context, body []byte) ([]byte, error)
-	SearchProductVersions(ctx context.Context, body []byte) ([]byte, error)
+	SearchProductVersions(ctx context.Context, productID string, body []byte) ([]byte, error)
 }
 
 // ProductHandler handles HTTP requests for product operations, delegating to the
@@ -82,11 +82,17 @@ func (h *ProductHandler) SearchProducts(w http.ResponseWriter, r *http.Request) 
 	writeJSON(w, http.StatusOK, result)
 }
 
-// SearchProductVersions handles POST /product-versions/search.
+// SearchProductVersions handles POST /product/{id}/versions/search.
 func (h *ProductHandler) SearchProductVersions(w http.ResponseWriter, r *http.Request) {
 	user := middleware.UserInfoFromContext(r.Context())
 	if user == nil {
 		writeError(w, http.StatusUnauthorized, ErrMsgUnauthorized)
+		return
+	}
+
+	productID := r.PathValue("id")
+	if productID == "" {
+		writeError(w, http.StatusBadRequest, ErrMsgBadRequest)
 		return
 	}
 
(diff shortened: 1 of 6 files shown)


apps/csm-portal/backend/internal/handler/products.go
```diff
@@ -0,0 +1,118 @@
...
+	if user == nil {
+		writeError(w, http.StatusUnauthorized, ErrMsgUnauthorized)
+		return
+	}
+
+	r.Body = http.MaxBytesReader(w, r.Body, maxRequestBodyBytes)
+	body, err := io.ReadAll(r.Body)
+	if err != nil {
+		if _, ok := err.(*http.MaxBytesError); ok {
+			writeError(w, http.StatusRequestEntityTooLarge, ErrMsgTooLarge)
```
kasunsiyambalapitiya, line 60: **[Medium] — Error Handling**

Direct type assertion `err.(*http.MaxBytesError)` does not unwrap errors. Prefer `errors.As`:

```go
var maxBytesErr *http.MaxBytesError
if errors.As(err, &maxBytesErr) {
    writeError(w, http.StatusRequestEntityTooLarge, ErrMsgTooLarge)
    return
}
```

Add `"errors"` to the import block.

apps/csm-portal/backend/internal/handler/products.go
```diff
@@ -0,0 +1,118 @@
...
+	result, err := h.entity.SearchProducts(r.Context(), body)
+	if err != nil {
+		slog.ErrorContext(r.Context(), "entity SearchProducts failed", "userID", user.UserID, "err", err)
+		mapUpstreamError(w, err, "Failed to search products.")
+		return
+	}
+
+	// TODO: Unmarshal result and filter to only the fields required by the frontend.
+	writeJSON(w, http.StatusOK, result)
+}
```
kasunsiyambalapitiya, line 83: **[Low] — Types**

`result` is `[]byte` (raw JSON bytes). If `writeJSON` calls `json.Marshal`/`json.Encoder` internally, the bytes will be serialised as a base64 string rather than forwarded as JSON.

Verify that `writeJSON` special-cases `[]byte` as a raw write, or use `json.RawMessage` to be explicit:
```go
writeJSON(w, http.StatusOK, json.RawMessage(result))
```

apps/csm-portal/backend/internal/handler/products.go
```diff
@@ -0,0 +1,118 @@
...
+	if user == nil {
+		writeError(w, http.StatusUnauthorized, ErrMsgUnauthorized)
+		return
+	}
+
+	r.Body = http.MaxBytesReader(w, r.Body, maxRequestBodyBytes)
+	body, err := io.ReadAll(r.Body)
+	if err != nil {
+		if _, ok := err.(*http.MaxBytesError); ok {
+			writeError(w, http.StatusRequestEntityTooLarge, ErrMsgTooLarge)
```
kasunsiyambalapitiya, line 104: **[Medium] — Error Handling**

Same issue as `SearchProducts` line 58: use `errors.As` instead of a direct type assertion to correctly handle wrapped `*http.MaxBytesError`.

apps/csm-portal/backend/internal/handler/products.go
```diff
@@ -0,0 +1,118 @@
...
+	result, err := h.entity.SearchProductVersions(r.Context(), body)
+	if err != nil {
+		slog.ErrorContext(r.Context(), "entity SearchProductVersions failed", "userID", user.UserID, "err", err)
+		mapUpstreamError(w, err, "Failed to search product versions.")
+		return
+	}
+
+	// TODO: Unmarshal result and filter to only the fields required by the frontend.
+	writeJSON(w, http.StatusOK, result)
+}
```
kasunsiyambalapitiya, line 127: **[Low] — Types**

Same `writeJSON([]byte)` concern as line 81 — verify or use `json.RawMessage(result)`.

apps/csm-portal/backend/internal/handler/products.go
Rashmika998: Fixed — switched to `errors.As` for both `SearchProducts` and `SearchProductVersions`. Note: the same direct type assertion pattern exists in `accounts.go` and `projects.go` as well; a follow-up PR can align those.

apps/csm-portal/backend/internal/handler/products.go
Rashmika998: No change needed here. `writeJSON` in this codebase takes `[]byte` and calls `w.Write(data)` directly — it does not go through `json.Marshal` or `json.Encoder`, so there is no base64 encoding risk. The upstream entity service returns valid JSON bytes, which are written as-is to the response. `json.RawMessage` would only be needed if the bytes were being passed to a marshaller, which they are not.

apps/csm-portal/backend/internal/handler/products.go
Rashmika998: Fixed in the same commit as the `SearchProducts` fix above — both handlers now use `errors.As`.

apps/csm-portal/backend/internal/handler/products.go
Rashmika998: Same as the reply on line 83 — `writeJSON` writes `[]byte` raw via `w.Write(data)`, not through `json.Marshal`, so there is no base64 encoding issue. No change needed.

cloby99 (approved):

PR #781, opened 2026-06-01
[CSM Portal] Align POST /cases with entity-service CreateCase contract

Rashmika998:
## Summary

Aligns the csm-portal `POST /cases` handler and OpenAPI spec with the entity-service `CreateCase` endpoint introduced in #780.

- **`createdBy` injection** — the authenticated user's ID is now injected server-side before forwarding to the entity service; any caller-supplied `createdBy` is overridden, preventing identity spoofing
- **JSON validation** — request body is validated as valid JSON before processing (consistent with `SearchProjectCases`)
- **`CaseCreatePayload` schema updated** — removes legacy fields (`type`, `title`, integer `issueTypeKey`/`severityKey`, `relatedCaseId`, `conversationId`, `catalogId`, `catalogItemId`); replaces with the entity-service contract: `subject`, `description`, `priority` and `issueType` as string enums, all six fields required
- **`POST /cases` 201 response** — now references the `Case` response schema; also adds explicit 403/404/500 error bodies to match `mapUpstreamError` behavior

## Test plan

- [x] `TestCreateCase/requires_authenticated_user` — 401 when no user in context
- [x] `TestCreateCase/rejects_body_exceeding_1_MiB` — 413 on oversized body
- [x] `TestCreateCase/rejects_invalid_JSON_body` — 400 on malformed JSON
- [x] `TestCreateCase/injects_createdBy_from_auth_user_and_returns_201_with_response` — upstream receives correct `createdBy` from JWT, response body is forwarded
- [x] `TestCreateCase/auth_user_id_overrides_any_caller-supplied_createdBy` — spoofed `createdBy` in body is replaced by auth user ID
- [x] `TestCreateCase/upstream_errors_are_mapped_correctly` — 401/403/404/400/500 mapping for all upstream error codes

## Summary by CodeRabbit

* **Bug Fixes & Security**
  * Case creation now rejects non-JSON or invalid payloads with 400 and enforces the authenticated user as the case creator (unknown user yields 403).

* **Documentation**
  * API spec updated: POST /cases now returns a JSON Case on 201; error responses standardized to JSON; case-creation payload simplified (breaking change).

* **Tests**
  * Improved coverage for case creation, user-resolution, and error mappings.

### DIFF
--- a/apps/csm-portal/backend/internal/handler/cases.go
+++ b/apps/csm-portal/backend/internal/handler/cases.go
@@ -19,12 +19,60 @@ package handler
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"io"
 	"log/slog"
 	"net/http"
 
 	"github.com/wso2-open-operations/cs-tools/apps/csm-portal/backend/internal/middleware"
 )
+
+var errUserNotFound = errors.New("authenticated user not found in entity service")
+
+// resolveUserID looks up the entity-service UUID for the given email via users/search.
+// TODO: remove once DB-level auth propagates the entity UUID directly through the token.
+func resolveUserID(ctx context.Context, entity entityCaseClient, email string) (string, error) {
+	body, err := json.Marshal(map[string]any{
+		"searchQuery": email,
+		"pagination":  map[string]any{"limit": 1, "offset": 0},
+	})
+	if err != nil {
+		return "", err
+	}
+	resp, err := entity.SearchUsers(ctx, body)
+	if err != nil {
+		return "", err
+	}
+	var result struct {
+		Users []struct {
+			ID string `json:"id"`
+		} `json:"users"`
+	}
+	if err := json.Unmarshal(resp, &result); err != nil {
+		return "", err
+	}
+	if len(result.Users) == 0 {
+		return "", errUserNotFound
+	}
+	return result.Users[0].ID, nil
+}
+
+// injectCreatedBy merges createdBy into a JSON request body.
+func injectCreatedBy(body []byte, userID string) ([]byte, error) {
+	var m map[string]json.RawMessage
+	if err := json.Unmarshal(body, &m); err != nil {
+		return nil, err
+	}
+	if m == nil {
+		m = make(map[string]json.RawMessage)
+	}
+	id, err := json.Marshal(userID)
+	if err != nil {
+		return nil, err
+	}
+	m["createdBy"] = id
+	return json.Marshal(m)
+}
 
 // injectProjectID merges a project ID into a JSON request body as projectIds: [id].
 func injectProjectID(body []byte, projectID string) ([]byte, error) {
@@ -49,6 +97,7 @@ type entityCaseClient interface {
 	CreateCase(ctx context.Context, body []byte) ([]byte, error)
 	SearchCases(ctx context.Context, body []byte) ([]byte, error)
 	GetCase(ctx context.Context, caseID string) ([]byte, error)
+	SearchUsers(ctx context.Context, body []byte) ([]byte, error)
 }
 
 // CaseHandler handles HTTP requests for case operations, delegating to the
@@ -66,9 +115,10 @@ func NewCaseHandler(entity entityCaseClient) *CaseHandler {
 const maxRequestBodyBytes = 1 << 20
 
 // CreateCase handles POST /cases.
-// TODO: createdBy is currently supplied by the client. Once DB-level auth is
-// in place it will be injected server-side from the authenticated user and
-// removed from the request payload.
+// createdBy is resolved server-side by looking up the authenticated user's email
+// in the entity service and injecting the UUID into the request body.
+// TODO: remove the users/search lookup once DB-level auth propagates the entity
+// UUID directly through the token claims.
 func (h *CaseHandler) CreateCase(w http.ResponseWriter, r *http.Request) {
 	user := middleware.UserInfoFromContext(r.Context())
 	if user == nil {
@@ -92,9 +142,26 @@ func (h *CaseHandler) CreateCase(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	result, err := h.entity.CreateCase(r.Context(), body)
+	userID, err := resolveUserID(r.Context(), h.entity, user.Email)
 	if err != nil {
-		slog.ErrorContext(r.Context(), "entity CreateCase failed", "userID", user.UserID, "err", err)
+		if errors.Is(err, errUserNotFound) {
+			writeError(w, http.StatusUnauthorized, ErrMsgUnauthorized)
+			return
+		}
+		slog.ErrorContext(r.Context(), "resolveUserID failed", "email", user.Email, "err", err)
+		writeError(w, http.StatusInternalServerError, ErrMsgInternal)
+		return
+	}
+
+	entityBody, err := injectCreatedBy(body, userID)
+	if err != nil {
+		writeError(w, http.StatusBadRequest, ErrMsgBadRequest)
+		return
+	}
+
+	result, err := h.entity.CreateCase(r.Context(), entityBody)
+	if err != nil {
+		slog.ErrorContext(r.Context(), "entity CreateCase failed", "userID", userID, "err", err)
 		mapUpstreamError(w, err, "Failed to create case.")
 		return
 	}
--- a/apps/csm-portal/backend/internal/handler/cases.go
+++ b/apps/csm-portal/backend/internal/handler/cases.go
@@ -26,6 +26,24 @@ import (
 	"github.com/wso2-open-operations/cs-tools/apps/csm-portal/backend/internal/middleware"
 )
 
+// injectCreatedBy merges the authenticated user's ID into a JSON request body as createdBy.
+// Any caller-supplied createdBy is overridden to prevent spoofing.
+func injectCreatedBy(body []byte, userID string) ([]byte, error) {
+	var m map[string]json.RawMessage
+	if err := json.Unmarshal(body, &m); err != nil {
+		return nil, err
+	}
+	if m == nil {
+		m = make(map[string]json.RawMessage)
+	}
+	id, err := json.Marshal(userID)
+	if err != nil {
+		return nil, err
+	}
+	m["createdBy"] = id
+	return json.Marshal(m)
+}
+
 // injectProjectID merges a project ID into a JSON request body as projectIds: [id].
 func injectProjectID(body []byte, projectID string) ([]byte, error) {
 	var m map[string]json.RawMessage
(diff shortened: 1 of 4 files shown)


apps/csm-portal/backend/internal/handler/cases.go
```diff
@@ -19,13 +19,61 @@ package handler
...
+	}
+	var result struct {
+		Users []struct {
+			ID string `json:"id"`
+		} `json:"users"`
+	}
+	if err := json.Unmarshal(resp, &result); err != nil {
+		return "", err
+	}
+	if len(result.Users) == 0 {
```
shayanmalinda: **⚠️ Important: First search result is not verified to be an exact email match**

`SearchUsers` is called with a `searchQuery` string, which typically performs a prefix or fuzzy search rather than a deterministic lookup. Taking `result.Users[0].ID` without checking that the returned user's email exactly equals the requested email means a partial match could inject a *different* user's UUID as `createdBy`. For example, `alice@example.com` could match a result for `alice-admin@example.com` if the service orders by relevance.

Suggested fix: either verify the returned user's email field matches `email` before trusting the ID, or call a deterministic lookup endpoint (e.g. `GET /users?email=<exact>`) if the entity service exposes one. At minimum, assert the match in a unit test.

```go
if len(result.Users) == 0 || result.Users[0].Email != email {
    return "", errUserNotFound
}
```

(You'd need to add `Email string` to the anonymous struct above.)

---

apps/csm-portal/backend/internal/handler/cases.go
```diff
@@ -90,15 +137,35 @@ func (h *CaseHandler) CreateCase(w http.ResponseWriter, r *http.Request) {
...
+	if !json.Valid(body) {
+		writeError(w, http.StatusBadRequest, ErrMsgBadRequest)
+		return
+	}
+
+	userID, err := resolveUserID(r.Context(), h.entity, user.Email)
 	if err != nil {
-		slog.ErrorContext(r.Context(), "entity CreateCase failed", "userID", user.UserID, "err", err)
+		if errors.Is(err, errUserNotFound) {
+			writeError(w, http.StatusUnauthorized, ErrMsgUnauthorized)
```
shayanmalinda: **🛠️ Medium: `errUserNotFound` should map to `403 Forbidden`, not `401 Unauthorized`**

By this point the request has already passed the auth middleware — `user` is non-nil and has a valid token. The issue is that the entity service has no record for this authenticated user, which is an *authorisation* failure, not an authentication one. HTTP `401` means "you haven't identified yourself"; `403` means "I know who you are but you can't do this". Using `401` here can confuse clients that respond to `401` by retrying with a refreshed token (which will loop forever since the identity problem isn't the token).

Suggested fix: return `http.StatusForbidden` + `ErrMsgForbidden` for the `errUserNotFound` case.

> 📖 **Reference ([WSO2 REST API Design Guidelines](https://wso2.com/whitepapers/wso2-rest-apis-design-guidelines/)):**
> "`401` unauthenticated, `403` forbidden"

---

apps/csm-portal/backend/internal/handler/cases.go
```diff
@@ -90,15 +137,35 @@ func (h *CaseHandler) CreateCase(w http.ResponseWriter, r *http.Request) {
...
+	result, err := h.entity.CreateCase(r.Context(), entityBody)
+	if err != nil {
+		slog.ErrorContext(r.Context(), "entity CreateCase failed", "userID", userID, "err", err)
 		mapUpstreamError(w, err, "Failed to create case.")
 		return
 	}
 
-	// TODO: Unmarshal result into a typed CaseCreateResponse and transform it
-	// to the CSM portal response shape before writing.
 	writeJSON(w, http.StatusCreated, result)
```
shayanmalinda, line 178: **🛠️ Medium: `201 Created` response is missing the `Location` header**

The REST guidelines require a `POST` that creates a resource to return a `Location` response header pointing at the new resource URI (e.g. `/cases/{id}`). The entity service response likely includes an `id` field that could be used to build this URI.

Suggested fix: unmarshal enough of `result` to extract the new case ID, then set the header before writing the body:

```go
var created struct { ID string `json:"id"` }
if err := json.Unmarshal(result, &created); err == nil && created.ID != "" {
    w.Header().Set("Location", "/cases/"+created.ID)
}
writeJSON(w, http.StatusCreated, result)
```

> 📖 **Reference ([WSO2 REST API Design Guidelines](https://wso2.com/whitepapers/wso2-rest-apis-design-guidelines/)):**
> "`201` created (+ Location)"

---

apps/csm-portal/backend/internal/handler/cases.go
```diff
@@ -90,15 +137,35 @@ func (h *CaseHandler) CreateCase(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	result, err := h.entity.CreateCase(r.Context(), body)
+	if !json.Valid(body) {
+		writeError(w, http.StatusBadRequest, ErrMsgBadRequest)
+		return
+	}
+
+	userID, err := resolveUserID(r.Context(), h.entity, user.Email)
```
shayanmalinda, line 147: **📝 Note: Extra round-trip on every POST /cases**

`resolveUserID` makes a synchronous `SearchUsers` call before every `CreateCase`, doubling the entity-service round-trips per request. The TODO comment covers the intention to remove this once JWT claims include the entity UUID, which is the right long-term fix. Worth tracking as a performance concern in the backlog: under load, two serial entity-service calls per case creation will roughly halve the effective throughput of this endpoint.

No action required on this PR — just flagging it so the TODO doesn't get forgotten.

---

apps/csm-portal/backend/internal/handler/cases_test.go
```diff
@@ -71,42 +79,89 @@ func TestCreateCase(t *testing.T) {
...
-		if string(capturedBody) != reqPayload {
-			t.Errorf("upstream received body %q, want %q", capturedBody, reqPayload)
+
+		var sent map[string]json.RawMessage
+		if err := json.Unmarshal(capturedBody, &sent); err != nil {
+			t.Fatalf("upstream received invalid JSON: %v", err)
+		}
+		var gotID string
+		if err := json.Unmarshal(sent["createdBy"], &gotID); err != nil || gotID != resolvedUserID {
+			t.Errorf("upstream createdBy = %q, want %q", gotID, resolvedUserID)
```
shayanmalinda, line 115: **📝 Nit: Test doesn't assert original payload fields are preserved after injection**

The test verifies that `createdBy` is set in the body sent upstream, which is great. It doesn't assert that the original fields (`projectId`, `deploymentId`, etc.) are still present after `injectCreatedBy` processes the body. If `injectCreatedBy` were accidentally changed to return only the injected field, this test would still pass.

Suggested addition:

```go
var gotProjectID string
if err := json.Unmarshal(sent["projectId"], &gotProjectID); err != nil || gotProjectID != "proj-1" {
    t.Errorf("upstream projectId = %q, want \"proj-1\"", gotProjectID)
}
```

---

apps/csm-portal/backend/openapi.yaml
```diff
@@ -43,10 +47,30 @@ paths:
...
+                $ref: '#/components/schemas/ErrorPayload'
+        "404":
+          description: NotFound
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorPayload'
         "413":
           description: RequestEntityTooLarge
         "500":
```
shayanmalinda, line 72: **📝 Nit: `413 RequestEntityTooLarge` is missing its response body schema**

All other error responses (`400`, `401`, `403`, `404`, `500`) now declare an `application/json` body with `$ref: '#/components/schemas/ErrorPayload'`, but `413` has no `content` block. The implementation does write a JSON error body for `413` (via `writeError`), so the spec is inconsistent.

Suggested fix:

```yaml
"413":
  description: RequestEntityTooLarge
  content:
    application/json:
      schema:
        $ref: '#/components/schemas/ErrorPayload'
```

---

apps/csm-portal/backend/internal/handler/cases.go
Rashmika998: Good catch. Fixed — added an `Email` field to the anonymous struct and now verify `result.Users[0].Email == email` before trusting the ID. Also added a dedicated test case for the email-mismatch scenario (`returns 403 when search result email does not match token email`).

apps/csm-portal/backend/internal/handler/cases.go
Rashmika998: Agreed — the token is valid at this point so 401 is wrong. Changed to `http.StatusForbidden` + `ErrMsgForbidden` for the `errUserNotFound` case.

apps/csm-portal/backend/internal/handler/cases.go
Rashmika998: Fixed — extracting the `id` from the entity response and setting `Location: /cases/{id}` before writing the body.

apps/csm-portal/backend/internal/handler/cases.go
Rashmika998: Acknowledged — tracked the concern. The TODO comment is intentional and the lookup will be removed once the entity UUID is available directly in the JWT claims.

apps/csm-portal/backend/internal/handler/cases_test.go
Rashmika998: Good suggestion. Added an assertion for `projectId` in the injection test to confirm original fields survive the round-trip through `injectCreatedBy`.

apps/csm-portal/backend/openapi.yaml
Rashmika998: Fixed — added the `content: application/json: schema: $ref: ErrorPayload` block to all 11 `413` responses across the spec for consistency.

apps/csm-portal/backend/internal/handler/cases.go
Rashmika998: This part will be removed once we have the db set up to get the user uuid from the email.

apps/csm-portal/backend/internal/handler/cases.go
```diff
@@ -19,13 +19,63 @@ package handler
...
+	}
+	if err := json.Unmarshal(resp, &result); err != nil {
+		return "", err
+	}
+	// Verify exact email match — SearchUsers may perform fuzzy/prefix search.
+	if len(result.Users) == 0 || result.Users[0].Email != email {
+		return "", errUserNotFound
+	}
+	return result.Users[0].ID, nil
+}
```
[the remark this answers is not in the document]
Rashmika998: User search is added temporary and will be removed once the db is set to get user uuid from the email

apps/csm-portal/backend/internal/handler/helpers_test.go
```diff
@@ -111,6 +112,13 @@ func (m *mockEntityCaseClient) GetCase(ctx context.Context, caseID string) ([]by
 	return []byte(`{}`), nil
 }
 
+func (m *mockEntityCaseClient) SearchUsers(ctx context.Context, body []byte) ([]byte, error) {
+	if m.searchUsersFn != nil {
+		return m.searchUsersFn(ctx, body)
+	}
+	return []byte(`{"users":[{"id":"user-123"}],"total":1}`), nil
+}
```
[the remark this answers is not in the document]
Rashmika998: User search is added temporary and will be removed once the db is set to get user uuid from the email

shayanmalinda (approved):

PR #791, opened 2026-06-03
[Customer Entity] Implement Case Comments Search Endpoint

cloby99:
## Summary
- Adds paginated search for comments on a specific case, ordered by `created_at` `DESC`
- Wires SearchCaseComments through domain → repository → service → handler following existing case patterns; COUNT and SELECT run concurrently on separate pool connections

## Summary by CodeRabbit

* **New Features**
  * Users can now search and retrieve case comments with pagination support. New capability enables filtering and browsing comments associated with specific cases with customizable limits and offsets for efficient data navigation.

### DIFF
--- a/entity-service/internal/domain/entity.go
+++ b/entity-service/internal/domain/entity.go
@@ -458,3 +458,20 @@ type CreateCaseCommentRequest struct {
 	Body        string      `json:"body"`
 	CreatedBy   string      `json:"createdBy"`
 }
+
+// SearchCaseCommentsRequest is the input for listing comments on a case.
+// CaseID is populated from the URL path parameter and is not part of the JSON body.
+type SearchCaseCommentsRequest struct {
+	CaseID     string     `json:"-"`
+	Pagination Pagination `json:"pagination"`
+}
+
+// SearchCaseCommentsResponse is the paginated result of a case comment search.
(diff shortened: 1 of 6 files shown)


Rashmika998 (approved):

entity-service/internal/domain/entity.go
```diff
@@ -458,3 +458,20 @@ type CreateCaseCommentRequest struct {
 	Body        string      `json:"body"`
 	CreatedBy   string      `json:"createdBy"`
 }
+
+// SearchCaseCommentsRequest is the input for listing comments on a case.
+// CaseID is populated from the URL path parameter and is not part of the JSON body.
+type SearchCaseCommentsRequest struct {
+	CaseID     string     `json:"-"`
```
v15a1, line 465: By ``json:"-"`` we are explicitly asking Golang to Ignore the field entirely. We can technically remove the entire param if it's not needed.

cloby99: The field is intentionally kept. json:"-" prevents it from being read from the request body since the case ID already comes from the URL path, having it in both would be ambiguous. The field is set in the handler via r.PathValue("id") after JSON decoding and passed through to the repository from there.

PR #814, opened 2026-06-05
feat(csm-portal): shared foundation for feature build-out

rksk:
Base layer of the stacked CSM-portal case-management series; later PRs (starting with #815) build on top of it. Pure infrastructure — nothing  is wired into the running app yet, so there is no user-visible change and no route change.

  ### What's here
  - **Auth-aware backend client** (`useBackendApi`) — one thin wrapper per HTTP verb (`get`/`post`/`patch`/`postEmpty`) over the existing
  auth-aware fetch, pairing the configured backend base URL with JSON (de)serialization. Errors surface as `BackendApiError` carrying the HTTP
  status and optional payload. `GET` resolves 404/204 to `null`. **Known gap:** the write verbs do not yet handle 204/empty-body responses — a
  follow-up before any endpoint that returns 204.
  - **Backend types** — `Be`-prefixed request/response types. **These currently mirror an earlier flat case shape and are out of sync with the
  v2 `openapi.yaml` (`CaseView`/`CaseSearchView` return nested `EntityRef` objects, `UserRef` createdBy, and `nextStates`). Regenerating from
  the spec is tracked as a follow-up and lands with #815.**
  - **Runtime mock-vs-backend toggle** (`MockModeProvider`/`useMockMode`, `hydrateMockModeFromStorage`) — persisted to `localStorage` under
  `csm.useMocks`. **Defined here but not yet mounted; #815 wires the provider into the tree and calls `hydrateMockModeFromStorage()` from
  `main.tsx`.**
  - **Display utilities** — `dateTime.ts` (timezone-aware backend-timestamp formatting) and `formatBytes.ts` (B/KB/MB; no GB tier yet).
  - **API constants** — React Query/mutation keys for the CSM features in this stack. **Note:** several keys are ported verbatim from
  customer-portal and are not used by this stack; the auth/token constants are staged for a later `useAuthApiClient` retry change and are not
  consumed yet.
  - **Path aliases** — adds `@api` (resolves today) plus feature-scoped aliases (`@case-details*`, `@update-cards`, `@time-tracking`,
  `@deployments`) whose target directories are introduced by later PRs in the stack. `vite.config.ts` and `tsconfig.app.json` kept in sync.

  ### Test plan
  - [ ] `pnpm build`
  - [ ] `pnpm lint`
  - [ ] `pnpm test`

  Key corrections vs the current text: removed the false claim that the path aliases are @api/@constants/@context (the latter two pre-exist) and disclosed the dead targets; flagged the types are stale vs v2; stated the mock provider isn't mounted; stopped implying the API constants are consolidated/consumed; left the test boxes unchecked since they're unverified.


rksk:
@coderabbitai review

Rashmika998 (approved):

PR #820, opened 2026-06-08
[CSM Portal] Rename next_states to nextStates in GET /cases/{id} response

Rashmika998:
## Summary

- Renames the computed field `next_states` → `nextStates` in the `GET /cases/{id}` response to align with the camelCase convention used by all other portal-owned response fields
- Updates `state.go`, `cases.go`, `cases_test.go`, `openapi.yaml`, and `CLAUDE.md`

## Test plan

- [x] All `TestGetCase` subtests pass, including `nextStates reflects current state` for all 6 states

## Summary by CodeRabbit

* **API Changes**
  * Updated case response fields: `next_states` is now `nextStates`

### DIFF
--- a/apps/csm-portal/backend/internal/handler/cases_test.go
+++ b/apps/csm-portal/backend/internal/handler/cases_test.go
@@ -620,7 +620,7 @@ func TestGetCase(t *testing.T) {
 		assertContentType(t, w, "application/json")
 	})
 
-	t.Run("passes case ID to upstream and returns 200 with next_states injected", func(t *testing.T) {
+	t.Run("passes case ID to upstream and returns 200 with nextStates injected", func(t *testing.T) {
 		var capturedID string
 		client := &mockEntityCaseClient{
 			getCaseFn: func(_ context.Context, caseID string) ([]byte, error) {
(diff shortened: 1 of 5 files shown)


cloby99 (dismissed):

shayanmalinda (dismissed):

apps/csm-portal/backend/internal/handler/cases_test.go
v15a1: Can't we define a structure/type for this? Although the `ns, ok := resp["nextStates"].([]any)` prevent panics/crashes, it's not type safe which can result in unexpected results especially if the response structure changes, like it has.

v15a1 (approved):

apps/csm-portal/backend/internal/handler/cases_test.go
Rashmika998: Fixed

Rashmika998: Fixed in fa54268 — introduced a local `getCaseResp` struct decoded via `decodeJSON[getCaseResp]`. Both subtests now assert directly against `resp.ID` and `resp.NextStates []string`, eliminating the `.([]any)` type assertion entirely.

PR #862, opened 2026-06-15
[CSM Portal][BE] fix(handler): improve UUID validation error message and skip for SN IDs

Rashmika998:
## Summary

- Add `ErrMsgInvalidUUID = "Invalid UUID format."` constant so UUID format rejections return a specific, actionable message instead of the generic `"Invalid request payload."`
- Skip UUID format validation on `GetCase` and `PatchCase` when the `x-user-id-token` header is present — ServiceNow IDs do not follow UUID format and would otherwise be incorrectly rejected on the ServiceNow-backed data source
- Update handler tests to assert the new `ErrMsgInvalidUUID` message

## Test plan

- [x] `GET /cases/{id}` with a non-UUID ID and no `x-user-id-token` → 400 `"Invalid UUID format."`
- [x] `PATCH /cases/{id}` with a non-UUID ID and no `x-user-id-token` → 400 `"Invalid UUID format."`
- [x] `GET /cases/{id}` with a non-UUID ID and `x-user-id-token` present → request forwarded to entity service (SN ID accepted)
- [x] `PATCH /cases/{id}` with a non-UUID ID and `x-user-id-token` present → request forwarded to entity service (SN ID accepted)

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced error messaging: Case API endpoints now return specific "Invalid UUID format" error messages instead of generic bad request responses when case ID validation fails.

* **New Features**
  * Improved case ID handling: Case API validation now supports alternative case ID formats when authentication is present.

### DIFF
--- a/apps/csm-portal/backend/internal/handler/cases.go
+++ b/apps/csm-portal/backend/internal/handler/cases.go
@@ -314,8 +314,8 @@ func (h *CaseHandler) PatchCase(w http.ResponseWriter, r *http.Request) {
 		writeError(w, http.StatusBadRequest, "Case ID cannot be empty!")
 		return
 	}
-	if !uuidRe.MatchString(caseID) {
-		writeError(w, http.StatusBadRequest, ErrMsgBadRequest)
+	if r.Header.Get("x-user-id-token") == "" && !uuidRe.MatchString(caseID) {
+		writeError(w, http.StatusBadRequest, ErrMsgInvalidUUID)
 		return
 	}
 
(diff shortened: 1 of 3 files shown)


Rashmika998:
Both findings addressed in the latest commit:

1. **Whitespace-only header bypass** — wrapped both checks with `strings.TrimSpace(r.Header.Get("x-user-id-token")) == ""` so a header containing only whitespace is treated as absent.
2. **Missing bypass test coverage** — added `"allows non-UUID case ID when x-user-id-token is present"` test cases for both `PatchCase` and `GetCase`, verifying the request reaches the upstream with the non-UUID ID intact.

cloby99 (approved):

PR #871, opened 2026-06-16
[CSM Portal] case-create project search, deployed-product labels, dashboard count cells

rksk:
## What

Frontend-only changes to the CSM portal.

### Case create page
- **Project field is now a type-ahead search.** Previously the page paged through the entire project catalogue (hundreds of records) on mount before the form was usable. It now uses a new `AsyncProjectSelect` that queries the backend as the user types (debounced, first page of matches only), mirroring the existing async project filter on the case list.
- **Deployed-product labels fixed.** The deployed-product records returned by the backend already embed resolved `product` and `version` objects, but the option builder was reading flat fields that don't exist in that shape, so every option collapsed to a generic `Product` label. It now reads the embedded objects and renders `"{product name} {version}"` directly. This also removes the obsolete secondary lookups (paging `/products/search` and a per-product versions search), so the dropdown populates from the single deployed-products call.

### Case list page
- The filter panel now starts **expanded** by default instead of collapsed.

### Dashboard
- In the **Cases by severity and state** matrix, `0` cells now use the same size, weight, and link treatment as non-zero counts, so the grid reads consistently instead of muting the zeros.

## Notes
- The FE `BeDeployedProduct` type was corrected to match the nested shape the backend actually returns.
- Verified with `pnpm lint` and `pnpm build`.

## Summary by CodeRabbit

* **New Features**
  * Added an async Project selector with debounced, backend-backed search and stable selection during results updates.
  * Improved deployed product selection using embedded product/version data for smoother loading.

* **UI/UX Improvements**
  * Case filters panel now opens by default on the CSM Cases page.
  * Case count values are always shown as clickable links.
  * Case metadata band now highlights “Project type” and updates the shown fact set.

* **Data & Display Updates**
  * Customer tier values and related labels/colors now support only **basic** and **enterprise**.
  * Case detail customer context is updated to use embedded project account information with safe fallbacks.

### DIFF
--- a/apps/csm-portal/webapp/src/api/backend/types.ts
+++ b/apps/csm-portal/webapp/src/api/backend/types.ts
@@ -113,10 +113,21 @@ export interface BeDeployedProductRef {
 }
 
 /**
+ * Account summary embedded in the CaseView. `type` is the account/support tier
+ * (e.g. "Enterprise"); it is a free-form string, not the PG `basic|enterprise`
+ * enum, because ServiceNow-sourced cases pass their support tier through as-is.
+ */
+export interface BeCaseAccountRef {
+  id: string;
+  name?: string;
+  type?: string;
+}
+
+/**
  * `GET /cases/{id}` response — the rich CaseView. Unlike {@link BeCase} (the
  * flat create/legacy shape), it embeds the related entities as objects, so the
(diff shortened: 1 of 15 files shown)


rksk:
@coderabbitai review

Rashmika998 (approved):

PR #876, opened 2026-06-17
[CSM Portal] align case-detail with the backend: comments, creator & assignee

rksk:
Aligns the case-detail page with the real backend `CaseView` / comments payloads. Frontend only. (Folds in what was previously #877.)

## 1. Comments load failure

The case-detail page showed **"Could not load comments…"** despite a valid backend response. The FE typed each comment as `{ commentType, body, createdBy: string, createdAt }` and ran `htmlFromPlain(comment.body)`; the real shape is `{ type, content, createdBy: {…}, createdOn }`, so `body` was undefined, the mapper threw, and the whole query rejected.

- `BeCaseComment` retyped: `type`, `content` (rich-text HTML), `createdOn`, and a `createdBy` author object (`id`/`firstName`/`lastName`/`fullName`).
- `uiCommentFromBe` reads the new fields, derives the author name (`fullName` → `firstName+lastName` → `id`), and passes `content` through as HTML (the bubble already DOMPurify-sanitises it). Removed the obsolete plain-text wrapping.
- Create payload aligned to `{ type, content }`, sending the editor HTML as-is.
- Mapper unit tests updated.

## 2. Creator & assignee

- `createdBy` was read from `createdBy.displayName`; the payload uses `name`, so the reporter silently fell back to the raw email. `BeUserRef.displayName` renamed to `name`; reporter now uses `createdBy.name` with an email fallback when empty (unhydrated users return `name: ""`).
- Assignee was hardcoded to "Unassigned". Added `assignedEngineer` (`BeEntityRef | null`) to `BeCaseView`; assignee now reads `assignedEngineer.name`, falling back to "Unassigned" when null.

Verified with `pnpm lint`, `pnpm exec tsc -b`, and `pnpm test` (84 passing).

## Notes for reviewers
- The **comment read** path is confirmed against the live response. The **comment create** path is aligned to the same model for consistency (the old request stripped HTML and used `commentType`/`body`, which can't match the new read shape) — worth a real post-a-comment test against the deployed backend to confirm the create request contract.
- `assigneeIsMe` stays `false`: the "is me" highlight needs the signed-in user's entity id to compare against `assignedEngineer.id`, which this mapper doesn't have. Follow-up.

### DIFF
--- a/apps/csm-portal/webapp/src/api/backend/mappers.test.ts
+++ b/apps/csm-portal/webapp/src/api/backend/mappers.test.ts
@@ -93,50 +93,55 @@ describe("uiCommentFromBe", () => {
   const base: BeCaseComment = {
     id: "c1",
     caseId: "case1",
-    commentType: "comment",
-    body: "hello",
-    createdBy: "user-123",
-    createdAt: "2026-06-01T10:00:00Z",
+    type: "comment",
+    content: "<p>hello</p>",
+    createdBy: {
+      id: "user@wso2.com",
+      firstName: "Ada",
+      lastName: "Lovelace",
+      fullName: "Ada Lovelace ⓦ",
+    },
+    createdOn: "2026-06-01T10:00:00Z",
   };
 
   it("wraps a public comment as a wso2_engineer, non-internal bubble", () => {
     const ui = uiCommentFromBe(base);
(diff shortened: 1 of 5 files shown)


rksk:
@coderabbitai review

Rashmika998 (approved):

PR #877, opened 2026-06-17
[CSM Portal] align case-detail creator & assignee with the CaseView payload

rksk:
## Problem

On the case-detail page:
- **Creator** was read from `createdBy.displayName`, but the CaseView embeds the reporter as `{ id, name, userId, email }`. With no `displayName`, the reporter silently fell back to the raw email.
- **Assignee** was hardcoded to "Unassigned" — the CaseView now carries `assignedEngineer`.

CaseView payload (relevant fields):

```json
"assignedEngineer": { "id": "94a1b01b-…", "name": "John Doe ⓦ" },
"createdBy": { "id": "", "name": "", "userId": "", "email": "john@abc.com" }
```

## Changes (frontend only)

- `BeUserRef.displayName` renamed to **`name`** to match the payload (only the case-detail mapper read it).
- Added **`assignedEngineer`** (`BeEntityRef | null`) to `BeCaseView`.
- **Reporter** now uses `createdBy.name`, falling back to the email when `name` is empty (unhydrated users return `name: ""`).
- **Assignee** now uses `assignedEngineer.name`, falling back to "Unassigned" when null.

Verified with `pnpm lint`, `pnpm exec tsc -b`, and `pnpm test` (84 passing).

## Note

`assigneeIsMe` stays `false` for now — the "is me" highlight needs the signed-in user's entity id to compare against `assignedEngineer.id`, which this mapper doesn't have. Left as a follow-up.

## Summary by CodeRabbit

* **Bug Fixes**
  * Case detail views now accurately display assigned CS engineers instead of showing all cases as unassigned, improving visibility into case ownership and team allocation
  * Standardized how user information is referenced throughout the case management interface

### DIFF
--- a/apps/csm-portal/webapp/src/api/backend/types.ts
+++ b/apps/csm-portal/webapp/src/api/backend/types.ts
@@ -95,7 +95,7 @@ export interface BeCase {
 /** A referenced user, as embedded in case views (not just an id string). */
 export interface BeUserRef {
   id: string;
-  displayName?: string;
+  name?: string;
   userId?: string;
   email?: string;
 }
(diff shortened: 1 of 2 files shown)


rksk:
Folding these creator/assignee changes into #876, which now carries both the comment-shape fix and the case-detail creator/assignee alignment. Closing in favour of that PR.

PR #878, opened 2026-06-17
[CSM Portal] UI improvements: case-id search, action gating, comment render & dashboard reconciliation

rksk:
## What

A batch of CSM-portal UI improvements:

1. **Global search by case id.** The Ctrl/Cmd-K quick-nav now resolves a CS/WSO2 case id (or any subject text) to real cases via `POST /cases/search`, and Enter jumps to the case. Previously it matched only in-memory pinned/recent/page entries. Debounced, gated to ≥2 chars, capped at 8 hits, and disabled while the palette is closed.
2. **Gate unimplemented actions.** Every case-detail *More actions* item except **Copy case link** is disabled, and the **Time tracking** tab is disabled, until their backends are wired up — so the menu/tabs still show the roadmap without exposing dead actions.
3. **Fix the just-posted comment rendering.** The comment-create response is a thin ack (no body, type, or author display name), so the appended entry rendered as "Unknown" with an empty body. The list is now refetched on success so the new comment hydrates from `comments/search`.
4. **Reconcile the dashboard pies with the case matrix.** The severity/state matrix tracks active cases (closed excluded), but the two composition pies counted the whole population, so per-severity totals disagreed with the matrix (the gap was the closed cases). Both pies are now scoped to the same active states as the matrix, all three totals reconcile, and the closed count is shown as a caption rather than a slice.

## Why

Items 1–3 remove confusing/dead UI in the case experience. Item 4 fixes a real numeric inconsistency on the dashboard (e.g. the S0 severity-pie total no longer disagrees with the matrix's S0 total).

## Notes

- Frontend only. A few items surfaced backend follow-ups (echo the full comment on create; a `work_note` create rejection; folding the retired `reopened` state into `waiting_on_wso2` at the data source) — these are tracked separately and are out of scope here.
- The dashboard still shows a `Reopened` bucket; it will be retired on the FE once the data source stops emitting that state, to avoid orphaning any case still in it.

## Test

`pnpm build`, `pnpm lint`, and `pnpm test` (84 tests) all pass.

## Summary by CodeRabbit

* **New Features**
  * Added quick case search functionality to the quick navigation, allowing users to search cases by ID or subject with debounced input and real-time results.

* **Improvements**
  * Dashboard case composition now separately tracks closed cases and displays explanatory messaging that charts and matrix show active cases only.
  * Time tracking tab is now disabled in the case detail view.
  * Menu actions can now be individually disabled while remaining visible.

### DIFF
--- a/apps/csm-portal/webapp/src/features/csm-cases/api/useCsmCaseComments.ts
+++ b/apps/csm-portal/webapp/src/features/csm-cases/api/useCsmCaseComments.ts
@@ -135,11 +135,17 @@ export function usePostCsmCaseComment(): UseMutationResult<
       >(`/cases/${encodeURIComponent(input.caseId)}/comments`, payload);
       return uiCommentFromBe(created);
     },
-    onSuccess: (newComment, variables) => {
-      queryClient.setQueryData<CsmCaseComment[] | undefined>(
-        [ApiQueryKeys.CSM_CASE_COMMENTS, variables.caseId],
-        (prev) => (prev ? [...prev, newComment] : [newComment]),
-      );
+    onSuccess: (_newComment, variables) => {
+      // The BE create response is a thin ack — it echoes only {id, createdOn,
+      // createdBy(email string)}, not the author's display name, the rendered
+      // content, or the comment type. Appending that partial object renders the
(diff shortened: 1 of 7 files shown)


rksk:
@coderabbitai review

Rashmika998 (approved):

PR #884, opened 2026-06-17
[Customer Entity] Nest case search filters, fix totalRecords mapping, and forward sortBy to ServiceNow

cloby99:
## Summary
- Restructures SearchCasesRequest to nest all filter fields (searchQuery, projectIds, deploymentIds, deployedProductIds, stateKeys, priorityKeys, issueTypeKeys) under a filters object, with sortBy and pagination at the top level
- Fixes snCasesResponse to map totalRecords from the correct JSON field returned by the SN /cases/search endpoint, so total and hasMore in the response are accurate
- Adds sortBy forwarding to ServiceNow, mapping domain sort fields (created_at → createdOn, updated_at → updatedOn) to SN's expected field names

@coderabbitai ignore

## Summary by CodeRabbit

* **Refactor**
  * Updated the case search API request format by grouping optional filter criteria (search query, project/deployment/deployed product IDs, and state/priority/issue-type keys) into a dedicated `filters` object.
  * Kept `sortBy` and `pagination` at the top level while aligning behavior across the service and repository.
  * Added support for optional ServiceNow case sorting and refreshed the OpenAPI schema to match the new request shape.

PR #886, opened 2026-06-17
[CSM Portal][BE] feat(cases): update case search payload to nest filters under filters object

Rashmika998:
## Summary
- Restructures `CaseSearchPayload` in `openapi.yaml` to nest all filter fields under a `filters` object, with `sortBy` and `pagination` at the top level — aligns the CSM Portal BFF contract with the entity-service restructure from PR #884
- Raises `CreateCase` body cap from 1 MiB (`maxRequestBodyBytes`) to 10 MiB (`maxCaseBodyBytes`) to match the comment endpoint
- Gates comment creation on case state: `CreateCaseComment` fetches the case and rejects with 409 if `state != work_in_progress` or `workState != ongoing` (refs digiops-cs#2093)
- Adds `workState` field to the `Case` schema and documents the 409 response on `POST /cases/{id}/comments`

## Test plan
- [x] `TestSearchCases` — updated to use nested `filters` payload shape
- [x] `TestCreateCase` — updated to assert 10 MiB cap
- [x] `TestCreateCaseComment` — three new guard tests (wrong state, `on_hold` work state, absent work state); existing tests updated with `getCaseFn`
- [x] Pre-push hook ran full backend test suite — all green

### DIFF
--- a/apps/csm-portal/backend/internal/handler/cases.go
+++ b/apps/csm-portal/backend/internal/handler/cases.go
@@ -68,6 +68,9 @@ func NewCaseHandler(entity entityCaseClient) *CaseHandler {
 // maxRequestBodyBytes caps incoming request bodies at 1 MiB to prevent memory DoS.
 const maxRequestBodyBytes = 1 << 20
 
+// maxCaseBodyBytes caps case-create bodies at 10 MiB to accommodate rich descriptions.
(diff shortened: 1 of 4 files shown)


cloby99 (approved):

cloby99:
@coderabbitai review

PR #898, opened 2026-06-20
[CSM Portal][BE] feat(tracing): rename X-Correlation-ID to X-CSM-Correlation-ID

Rashmika998:
## Summary

- Renames the correlation ID header from `X-Correlation-ID` to `X-CSM-Correlation-ID` across the frontend, csm-portal backend, and entity-service
- The Choreo API gateway reserves and overwrites `X-Correlation-ID` for its own internal tracing, causing the frontend-originated ID to be replaced before it reaches the backend; using a CSM-specific header name bypasses this interception
- The rename is applied consistently in all six source locations plus the two CLAUDE.md files

## Files changed

| Component | File |
|-----------|------|
| Frontend constant & test | `webapp/src/utils/correlationId.ts`, `correlationId.test.ts` |
| Frontend API client comment | `webapp/src/api/backend/client.ts` |
| CSM Portal backend middleware | `backend/internal/middleware/correlation.go` |
| CSM Portal entity client | `backend/internal/entity/client.go` |
| Entity-service middleware | `entity-service/internal/middleware/correlationid.go` |
| Docs | `apps/csm-portal/backend/CLAUDE.md`, `entity-service/CLAUDE.md` |

## Test plan

- [x] Confirm `X-CSM-Correlation-ID` is added to the Choreo CORS `Access-Control-Allow-Headers` list (replacing or alongside `x-correlation-id`)
- [x] Verify frontend console logs and backend logs show the same correlation ID for a given request in staging
- [x] Run backend unit tests (`make test` in `apps/csm-portal/backend` and `entity-service`)

## Summary by CodeRabbit

* **Chores**
  * Updated request correlation ID header naming convention from `X-Correlation-ID` to `X-CSM-Correlation-ID` for consistency across backend and frontend services.

### DIFF
--- a/entity-service/internal/middleware/correlationid.go
+++ b/entity-service/internal/middleware/correlationid.go
@@ -25,11 +25,11 @@ import (
 
 // correlationIDHeader is the HTTP header used to propagate the correlation ID
 // across service boundaries.
-const correlationIDHeader = "X-Correlation-ID"
+const correlationIDHeader = "X-CSM-Correlation-ID"
 
 type correlationIDKey struct{}
 
-// CorrelationID is an HTTP middleware that reads the X-Correlation-ID request
+// CorrelationID is an HTTP middleware that reads the X-CSM-Correlation-ID request
 // header (forwarded by the portal BFF) or generates a UUID v4 if absent.
 // The ID is stored in the request context for logging and echoed in the
 // response header so callers can reference it in support requests.
--- a/apps/csm-portal/backend/CLAUDE.md
+++ b/apps/csm-portal/backend/CLAUDE.md
@@ -6,7 +6,7 @@ Go HTTP server (`net/http`, Go 1.22+) that acts as a backend-for-frontend (BFF) 
 
 `CorrelationID → Auth → Logger → Mux`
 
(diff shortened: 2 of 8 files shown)


entity-service/internal/middleware/correlationid.go
```diff
@@ -25,11 +25,11 @@ import (
 
 // correlationIDHeader is the HTTP header used to propagate the correlation ID
 // across service boundaries.
-const correlationIDHeader = "X-Correlation-ID"
+const correlationIDHeader = "X-CSM-Correlation-ID"
```
cloby99, line 28: Is there a specific reason for adding `CSM` to the correlation header name? 
Would it make sense to use a more generic header name, since `entity-service` isn't CSM-specific?

Rashmika998: As mentioned in the PR description Choreo overrides the header and generates a new correlation ID according to claude. We have to change the name to another to fix it.

Regarding the name, since this is for CSM portal it should have an identification for it. entity service may receive more headers in future (for ex: correlation header related to the customer portal). In that case entity-service will be used by 2 services. so we need a way to differentiate the logs of two services separately. So it's better to have this header as `x-csm-correlation-id`

cloby99 (approved):

PR #917, opened 2026-06-23
[Customer Portal] fix: re-authenticate expired Asgardeo session instead of showing error page

rksk:
## Problem

The auth-aware fetch wrapper (`useAuthApiClient` → `authFetch`) caught the Asgardeo unauthenticated error (`SPA-AUTH_CLIENT-VM-IV02`) and **blindly retried the same call once**, on the assumption a concurrent caller may have refreshed the token.

When nothing had, the retry replayed the same dead token and failed identically. The error then propagated up and rendered as a generic **"500 / something went wrong on our side"** page with a Retry button that loops, misdiagnosing an **auth failure as a server failure**.

## Fix

The fix lives entirely in the auth client, so every caller inherits it. Feature hooks and error-page components are untouched.

- **Classify the error.** Extracted `isTokenExpiredError(error)` (checks `error?.code === ASGARDEO_UNAUTHENTICATED_CODE`). Any other error (network, real backend 5xx) is rethrown untouched so existing error handling / per-component error UIs + manual retry still work.
- **Recovery ladder** when the error is the unauthenticated code:
  - **Retry the original request once.** `getIdToken()` only reads the stored token, but the `AsgardeoProvider` runs a periodic background refresh (`periodicTokenRefresh`), so a concurrent refresh may have re-minted the token by the time the retry runs.
  - If the retry **still** returns the unauthenticated code, the session is gone → redirect to a full `signIn()`.
  - If the retry fails for a **non-auth** reason (e.g. a transient network blip on the second attempt), rethrow it rather than bouncing the user to sign-in.
- **Never fall through to an error page on the redirect.** The sign-in path returns a never-resolving promise so callers don't render an error page while the browser navigates away.
- **Single-flight the redirect** at module scope (`signInInFlight`) so concurrent auth failures trigger exactly one `signIn()`.

`attemptFetch` (token fetch + header build + fetch) is unchanged.

### Design note: why no explicit silent sign-in

An earlier revision called `signInSilently()` to silently re-mint tokens before the retry. Removed: it relies on a hidden auth iframe (third-party cookies), which can hang and needed a timeout guard, and it was redundant given the provider's periodic background refresh. The simpler retry-then-redirect ladder degrades cleanly: if nothing refreshed the token, the retry fails and we redirect to sign-in (which is the correct outcome for a dead session).

Verified at the source (`@asgardeo/browser@0.7.9`): `getIdToken()` delegates to the core client, which reads the stored token; token refresh is a separate operation. So the retry only recovers when a background/periodic refresh has already updated the stored token, otherwise it redirects.

## Testing

- `tsc -b` (typecheck): clean
- `eslint`: clean
- `vitest`: 7 passing — POST header build, missing-token throw, non-auth rethrow, retry-recovers, non-auth-error-on-retry rethrow, redirect-when-still-unauthenticated, and single-flight redirect across concurrent failures.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved authentication token recovery to properly distinguish between authentication-related errors and other failures.
  * Enhanced handling of concurrent authentication failures to prevent duplicate sign-in redirects.
  * Users are now correctly redirected to sign-in when automatic token recovery fails.

### DIFF
--- a/apps/customer-portal/webapp/src/hooks/__tests__/useAuthApiClient.test.ts
+++ b/apps/customer-portal/webapp/src/hooks/__tests__/useAuthApiClient.test.ts
@@ -64,14 +64,6 @@ describe("useAuthApiClient", () => {
     expect(headers.get("Content-Type")).toBe("application/json");
   });
 
-  it("throws when ID token is missing", async () => {
-    getIdTokenMock.mockResolvedValueOnce(undefined);
-    const { result } = renderHook(() => useAuthApiClient());
-    await expect(result.current("https://api.test")).rejects.toThrow(
-      "Unable to retrieve ID token",
-    );
-  });
-
   it("rethrows non-auth errors without attempting recovery", async () => {
     getIdTokenMock.mockRejectedValueOnce(new Error("network down"));
     const { result } = renderHook(() => useAuthApiClient());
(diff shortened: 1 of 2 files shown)


rksk:
@coderabbitai review

Rashmika998 (approved):

PR #928, opened 2026-06-24
[Customer Portal Web App] feat: usage overview environment breakdown and metrics improvements

SasmithaDilshan:
Description:
  ## Summary
  
  - **Environment breakdown product cards**: Fixed product/instance counts showing zero by falling back to `product.id`/`product.label` when `deployedProduct` is null
  - **Row header counts**: Switch to deployment-level instance data for accurate product/instance/core counts in the collapsed row header; usages fetched lazily on first expand to avoid N
  concurrent API calls on mount
  - **Product count stability**: Product count stays consistent between collapsed and expanded states — falls back to project-level count until usages are cached, then uses deployment-level
  data thereafter
  - **Product count correctness**: Count distinct products from both instances and usages so usage-only products (0 instances) are included in the count
  - **Overview stat cards**: Wire `useGetProjectUsageStats` current values into the Environments/Products/Instances stat cards; derive min/avg/max from time-series metrics
  - **Asgardeo**: Upgrade `@asgardeo/react` from `0.25.5` → `0.25.6`
  
  ## Test plan

  - [ ] Expand an environment accordion — product cards show with correct names, instance counts, and core counts
  - [ ] Collapse the accordion — product count and instance count in the row header match the expanded view
  - [ ] Verify no continuous loading on page load (usages only fire for the currently expanded row)
  - [ ] Overview stat cards (Environments, Products, Instances) show correct current values
  - [ ] Asgardeo auth flow works as before (login, token refresh, logout)

## Summary by CodeRabbit

* **New Features**
  * Usage metrics now show richer summaries with current, average, minimum, and maximum values across several cards and breakdowns.
  * Environment and product usage views now include expanded instance and core summary metrics for clearer comparison.

* **Bug Fixes**
  * Improved how usage charts and summary cards pick the most relevant current value from recent data.
  * Updated organization usage trends to use the correct metric source.
  * Refined loading and fallback behavior so usage panels display more reliably when data is still arriving.

### DIFF
--- a/apps/customer-portal/webapp/pnpm-workspace.yaml
+++ b/apps/customer-portal/webapp/pnpm-workspace.yaml
@@ -1,3 +1,6 @@
 allowBuilds:
   core-js: true
   esbuild: true
+minimumReleaseAgeExclude:
+  - '@asgardeo/browser@0.7.10'
+  - '@asgardeo/react@0.25.6'
--- a/apps/customer-portal/webapp/package.json
+++ b/apps/customer-portal/webapp/package.json
@@ -11,7 +11,7 @@
     "test": "vitest"
   },
   "dependencies": {
-    "@asgardeo/react": "0.25.5",
+    "@asgardeo/react": "0.25.6",
     "@asgardeo/react-router": "2.0.0",
     "@lexical/code": "0.40.0",
     "@lexical/html": "0.40.0",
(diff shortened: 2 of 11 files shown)


apps/customer-portal/webapp/pnpm-workspace.yaml
```diff
@@ -1,3 +1,6 @@
 allowBuilds:
   core-js: true
   esbuild: true
+minimumReleaseAgeExclude:
```
Rashmika998, line 4: Let's remove this may be tomorrow as this is only needed for choreo build for today.

Rashmika998 (approved):

PR #931, opened 2026-06-24
[CSM Portal] remove runtime mock-data mode and its dead filters

rksk:
## What

Removes the runtime **mock-data mode** (the header Mock/Live toggle) and the entire seeded-mock layer, so the app always talks to the live backend.

Removed:
- The header mock toggle, its context provider, and boot-time hydration.
- `isMockMode()` and its branches in every data hook.
- The seeded cases / comments / dashboard mock modules and the mock-only client-side cases filter.
- The build-time mock config flags and the local-gateway mode's mock pinning.

With the mock data gone, this also drops the cases-filter controls that only ever worked against it and have **no live `/cases/search` support**:
- the My-ABT / All-customers scope toggle, and the **SLA** and **product** filters;
- the orphaned, never-rendered `MyQueueSection` / `SlaAtRiskSection` dashboard widgets built around those filters.

The **assignee filter is kept** (left disabled with a "coming soon" note); it is wired to the live backend in the stacked follow-up PR. The cases filter keeps **search, severity, state, and project** — all backed by the search API.

## Merge order

This PR is independent of the case-filters PR and is intended to **merge first**. The case-type + assignee work stacks on top of it.

## Testing

- `pnpm lint`, `pnpm test`, `pnpm build`, `tsc -b` all green.
- Bundle shrinks ~35 KB (seeded mock data dropped).


rksk:
@coderabbitai review

dilshanfardil (approved):

shayanmalinda (approved):

PR #941, opened 2026-06-24
[CSM Portal] stop calling missing /csm/dashboard endpoint

rksk:
## What

The CSM dashboard fired `GET /csm/dashboard?scope=…` on every load, but the BFF has **no such endpoint** — so it was a guaranteed 404 (and left the header stuck on "Loading engineer…").

- **Disable the query** (`DASHBOARD_ENDPOINT_READY = false`) in `useGetCsmDashboard` so it never fires. The fetch logic is kept — flip the flag to re-enable in one line once the BE ships `/csm/dashboard`.
- Header engineer-overview line degrades to a neutral **"Engineer overview"** subtitle instead of a perpetual "Loading engineer…"; drops the now-unused `isError` prop.
- The live **severity-by-state matrix** (`CaseCountsMatrix`, backed by `/cases/search`) and the composition charts are unaffected.

Independent of the open CSM PRs; mergeable on its own.

## Testing

`pnpm lint`, `pnpm test`, `pnpm build`, `tsc -b` all green.

### DIFF
--- a/apps/csm-portal/webapp/src/features/csm-dashboard/api/useGetCsmDashboard.ts
+++ b/apps/csm-portal/webapp/src/features/csm-dashboard/api/useGetCsmDashboard.ts
@@ -23,10 +23,15 @@ import type {
 } from "@features/csm-dashboard/types/abtDashboard";
 
 /**
- * Loads the CSM ABT dashboard payload.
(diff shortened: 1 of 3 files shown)


rksk:
Folded into #937 (the dashboard /csm/dashboard 404 fix is now commit `e68717471` there). Closing this in favour of the consolidated PR.

PR #973, opened 2026-06-29
[Customer Entity] Add filters, sort, and ServiceNow support to POST /users/search

cloby99:
## Summary

- Restructures `POST /users/search` request from a flat `searchQuery` field to a nested `filters` object (searchQuery, roles, userNames, emails, active) with an optional `sortBy` (field, order)
- Adds ServiceNow data source support: when `DataSource == ServiceNow`, the endpoint is backed by Choreo `/users/search` and returns `SearchSNUsersResponse` with `SNUser` items (name, timeZone, active, roles) instead of the postgres `User` shape
- Postgres data source continues to work as before, with additional `userNames` and `emails` filter support added to the repository layer

## Changes

| Layer | Change |
|---|---|
| Domain | New `UserRole`, `UserSortField`, `UserSortOrder` types; `SearchUsersFilters`, `UserSortBy`; `SNUser`, `SearchSNUsersResponse` |
| Service | `normalizeUserPagination` (default 10, max 50); `SNUserService` interface; `snUserService` backed by Choreo |
| Repository | Postgres `SearchUsers` extended with `userNames` and `emails` filters |
| Handler | `SNUserHandler` added alongside existing postgres `UserHandler` |
| Routes | `POST /users/search` conditionally routes to SN or postgres handler based on `DataSource` |
| OpenAPI | `SearchUsersRequest`, `UserPagination`, `SearchUsersFilters`, `UserSortBy`, `SNUser`, `SearchSNUsersResponse` schemas added/updated |

## Test plan

- [ ] `POST /users/search` with postgres data source returns existing `User` shape with `firstName`, `lastName`, `userType`
- [ ] `POST /users/search` with ServiceNow data source returns `SNUser` shape with `name`, `timeZone`, `active`, `roles`
- [ ] `roles` filter with an invalid value returns `400`
- [ ] `sortBy.field` with an invalid value returns `400`
- [ ] `pagination.limit > 50` returns `400`
- [ ] `userNames` and `emails` filters correctly narrow postgres results
- [ ] Missing `x-user-id-token` header on SN data source returns `401`

## Summary by CodeRabbit

* **New Features**
  * Expanded user search to support filters, sorting, and updated pagination defaults.
  * Added support for ServiceNow-backed user search results alongside the existing user list.

* **Bug Fixes**
  * Search requests now handle additional criteria like roles, names, emails, and active status more consistently.
  * Pagination limits and defaults were tightened for safer, more predictable results.

### DIFF
--- a/entity-service/internal/domain/entity.go
+++ b/entity-service/internal/domain/entity.go
@@ -50,21 +50,68 @@ type User struct {
 	UpdatedOn time.Time `json:"updatedOn"`
 }
 
+// UserRole represents a ServiceNow role that can be assigned to a user.
+type UserRole string
+
+const (
+	UserRoleInternal      UserRole = "internal"
+	UserRoleAgent         UserRole = "agent"
+	UserRoleAdmin         UserRole = "admin"
+	UserRoleCommenter     UserRole = "commenter"
+	UserRoleExternal      UserRole = "external"
+	UserRoleCustomer      UserRole = "customer"
+	UserRoleCustomerAdmin UserRole = "customer_admin"
+	UserRolePartner       UserRole = "partner"
+	UserRolePartnerAdmin  UserRole = "partner_admin"
+)
+
+// UserSortField enumerates the columns by which user search results may be ordered.
+type UserSortField string
+
+const (
+	UserSortFieldName      UserSortField = "name"
+	UserSortFieldCreatedOn UserSortField = "createdOn"
+	UserSortFieldUpdatedOn UserSortField = "updatedOn"
+)
+
+// UserSortOrder is the direction of a user search sort.
+type UserSortOrder string
+
+const (
+	UserSortOrderAsc  UserSortOrder = "asc"
+	UserSortOrderDesc UserSortOrder = "desc"
+)
+
 // Pagination controls which page of results is returned.
-// Limit defaults to 20 and is capped at 100 by the service layer.
+// Limit defaults to 10 and is capped at 50 by the service layer.
 type Pagination struct {
 	Limit  int `json:"limit"`
 	Offset int `json:"offset"`
 }
 
-// SearchUsersRequest is the input for a user search operation.
-// SearchQuery is matched case-insensitively against username and email.
-type SearchUsersRequest struct {
-	Pagination  Pagination `json:"pagination"`
+// SearchUsersFilters holds the optional filter criteria for a user search.
+type SearchUsersFilters struct {
 	SearchQuery string     `json:"searchQuery"`
+	Roles       []UserRole `json:"roles"`
+	UserNames   []string   `json:"userNames"`
+	Emails      []string   `json:"emails"`
+	Active      *bool      `json:"active"`
 }
 
-// SearchUsersResponse is the paginated result of a user search.
+// UserSortBy specifies the sort field and direction for a user search.
+type UserSortBy struct {
+	Field UserSortField `json:"field"`
+	Order UserSortOrder `json:"order"`
+}
+
+// SearchUsersRequest is the input for a user search operation.
+type SearchUsersRequest struct {
+	Pagination Pagination         `json:"pagination"`
+	Filters    SearchUsersFilters `json:"filters"`
+	SortBy     UserSortBy         `json:"sortBy"`
+}
+
+// SearchUsersResponse is the paginated result of a postgres user search.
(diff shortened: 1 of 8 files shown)


Rashmika998 (approved):

entity-service/openapi.yaml
```diff
@@ -31,13 +35,15 @@ paths:
               $ref: '#/components/schemas/SearchUsersRequest'
       responses:
         "200":
-          description: Users matching the search query.
+          description: Users matching the supplied filters.
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/SearchUsersResponse'
+                oneOf:
```
shayanmalinda, line 42: **🟡 Medium: `oneOf` response schemas overlap, so postgres responses match both branches**

The two schemas in this `oneOf` aren't mutually exclusive. `SNUser` declares no `required` properties and doesn't set `additionalProperties: false`, so a postgres `User` object (which just carries extra fields like `firstName`) also validates as an `SNUser`. As a result a postgres response — `users, total, limit, offset, hasMore` — satisfies **both** `SearchUsersResponse` and `SearchSNUsersResponse`. `oneOf` means "valid against exactly one", so strict validators and client/code generators will reject otherwise-correct postgres responses as ambiguous.

Suggested fix: the response shape is fixed per deployment (selected by `DataSource`) and the two payloads can't be told apart by content, so `oneOf` is the wrong construct here. Either use `anyOf` (matches one-or-more), or make the branches disjoint, e.g. add `additionalProperties: false` and a distinguishing `required` field to each schema so exactly one ever matches.

---

entity-service/internal/domain/entity.go
```diff
@@ -50,21 +50,68 @@ type User struct {
...
+	Order UserSortOrder `json:"order"`
+}
+
 // SearchUsersRequest is the input for a user search operation.
-// SearchQuery is matched case-insensitively against username and email.
 type SearchUsersRequest struct {
-	Pagination  Pagination `json:"pagination"`
-	SearchQuery string     `json:"searchQuery"`
+	Pagination Pagination         `json:"pagination"`
+	Filters    SearchUsersFilters `json:"filters"`
```
shayanmalinda, line 110: **❓ Question: moving `searchQuery` under `filters` is a breaking change to the request contract**

This restructures the request so `searchQuery` is now nested inside `filters` instead of being a top-level field. Because the handler decodes with `DisallowUnknownFields()`, any existing caller still sending the old flat shape (`{"searchQuery": "...", "pagination": {...}}`) will now get a `400 unknown field searchQuery` instead of having it silently ignored. Failing loudly is the safe behaviour, but it is still a breaking change to a `v2` API. Could you confirm every current caller of `POST /users/search` is being updated in lockstep? (I tried to locate consumers via code search but this repo isn't reliably indexed, so I couldn't verify the impact myself.)

Suggested fix: if consumers can't all migrate at once, consider accepting the legacy top-level `searchQuery` for a deprecation window, or bump the API major version to signal the break.

> 📖 **Reference ([WSO2 REST API Design Guidelines — Versioning](https://wso2.com/whitepapers/wso2-rest-apis-design-guidelines/)):**
> "Major bump = breaking change; minor bump = backward-compatible addition"

---

entity-service/internal/service/user_service.go
```diff
@@ -80,12 +97,27 @@ func NewUserService(repo repository.UserRepository) UserService {
 
 // SearchUsers implements UserService.
 func (s *userService) SearchUsers(ctx context.Context, req domain.SearchUsersRequest) (domain.SearchUsersResponse, error) {
-	if err := normalizePagination(&req.Pagination); err != nil {
+	if err := normalizeUserPagination(&req.Pagination); err != nil {
```
shayanmalinda, line 100: **❓ Question: was lowering the postgres pagination cap (max 100→50, default 20→10) intended for the existing path?**

This switches the postgres user search from `normalizePagination` (default 20, max 100) to `normalizeUserPagination` (default 10, max 50). For the already-shipped postgres `/users/search`, that's a behaviour change: a client that previously requested a `limit` between 51 and 100 will now receive `400 limit cannot exceed 50`, and clients relying on the old default will get 10 results instead of 20. If the goal of this PR was to add ServiceNow support, it may be safer to leave the postgres limits as they were and apply the new 10/50 values only to the SN path, so existing consumers aren't affected.

---

PR #974, opened 2026-06-29
[CSM][BE] feat: update POST /users/search with filters, sortBy, and SNUser response

Rashmika998:
## Summary

Implements the CSM portal backend side of entity-service PR #973.

- **`UserSearchPayload`** restructured: flat `searchQuery` moved into a nested `filters` object; new `sortBy` added; pagination now uses `UserPagination` (default 10, max 50 instead of 20/100)
- **`UserSearchFilters`**: `searchQuery`, `roles` (SN-only, enum), `userNames`, `emails`, `active` (SN-only)
- **`UserSortBy`**: `field` (`name`/`createdOn`/`updatedOn`) + `order` (`asc`/`desc`); ServiceNow data source only
- **`SNUser`** + **`SNUserSearchResponse`** schemas added for the ServiceNow data source response shape
- **`/users/search` 200 response** updated to `oneOf` (`UserSearchResponse` | `SNUserSearchResponse`)
- Added missing `403` response to `/users/search`
- Stale TODO comments removed from `SearchUsers` handler (all validation is handled in the entity service)
- `README.md` updated

The handler itself remains a passthrough — request body is forwarded to the entity service unchanged and the raw response is returned.

## Entity service PR
wso2-open-operations/cs-tools#973

## Test plan

- [x] `POST /users/search` with `{}` body forwards to entity service and returns response
- [x] `POST /users/search` with `{ "filters": { "searchQuery": "alice" } }` forwards correctly
- [x] Build passes, all existing handler tests pass

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

## Summary by CodeRabbit

* **Documentation**
  * Expanded the user search API documentation with supported filters, sorting options, and clearer response details.
  * Clarified that search results may return different user shapes depending on the data source.
* **New Features**
  * Updated the API contract to support source-specific search responses, including ServiceNow user results.

### DIFF
--- a/apps/csm-portal/backend/openapi.yaml
+++ b/apps/csm-portal/backend/openapi.yaml
@@ -612,7 +612,11 @@ paths:
 
   /users/search:
     post:
-      summary: Search users.
+      summary: Search users with optional filters, sort, and pagination.
+      description: >
+        Returns a paginated list of users. When the entity-service data source is
+        ServiceNow the response uses SNUser (name, timeZone, active, roles); for
+        the postgres data source it uses User (firstName, lastName, userType).
       operationId: postUsersSearch
       requestBody:
         description: User search payload
@@ -623,19 +627,27 @@ paths:
         required: true
       responses:
         "200":
-          description: Ok
+          description: Users matching the supplied filters.
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/UserSearchResponse'
+                oneOf:
+                  - $ref: '#/components/schemas/UserSearchResponse'
+                  - $ref: '#/components/schemas/SNUserSearchResponse'
         "400":
-          description: BadRequest
+          description: Bad request — invalid filter value, sort field, or pagination limit.
           content:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorPayload'
         "401":
           description: Unauthorized
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorPayload'
+        "403":
+          description: Forbidden
           content:
             application/json:
               schema:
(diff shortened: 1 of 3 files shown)


apps/csm-portal/backend/openapi.yaml
```diff
@@ -623,13 +627,15 @@ paths:
         required: true
       responses:
         "200":
-          description: Ok
+          description: Users matching the supplied filters.
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/UserSearchResponse'
+                oneOf:
```
shayanmalinda, line 634: **🟠 Concern: `oneOf` here is ambiguous because the two schemas overlap**

With `oneOf`, a response must validate against *exactly one* of the listed schemas. The problem is that `SNUserSearchResponse`'s fields (`users`, `total`, `limit`, `offset`) are a subset of `UserSearchResponse`'s (`users`, `total`, `limit`, `offset`, `hasMore`), and neither schema marks any field as `required`. So a real ServiceNow response like `{"users": [...], "total": 5, "limit": 10, "offset": 0}` validates against *both* branches — which a strict validator rejects, and code generators can't use to pick a type.

Note that a bare `discriminator:` keyword won't fix this on its own, since there's no field in the body (e.g. `dataSource`) that says which shape it is. Two workable options:

Suggested fix: make the two schemas mutually exclusive, e.g. require a distinguishing field on the postgres shape —
```yaml
UserSearchResponse:
  type: object
  required: [hasMore]   # SNUser response has no hasMore
  additionalProperties: false
  ...
```
or add an explicit discriminating field (e.g. `source: postgres | servicenow`) to both response bodies and use it as a `discriminator`. Either makes the union unambiguous. Worth confirming whether anything actually runs strict validation/codegen against this spec before deciding how far to go.

---

apps/csm-portal/backend/openapi.yaml
```diff
@@ -2768,18 +2780,72 @@ components:
...
-                limit:
-                  default: 20
-                  maximum: 100
+          $ref: '#/components/schemas/UserPagination'
+        filters:
+          $ref: '#/components/schemas/UserSearchFilters'
+        sortBy:
+          $ref: '#/components/schemas/UserSortBy'
+
+    UserPagination:
```
shayanmalinda, line 2789: **🔵 Question: could this reuse the existing `Pagination` schema?**

The new `UserPagination` duplicates `offset`/`limit` from the existing `Pagination` schema — the only real difference is `maximum: 50` (the `default: 10` already matches `Pagination`). Before this PR, `UserSearchPayload` composed it via `allOf: [Pagination, { limit override }]`, and the sibling `AccountSearchPayload` still uses that pattern.

Suggested fix: to stay consistent with `AccountSearchPayload` and avoid duplicating the base definition, consider —
```yaml
pagination:
  allOf:
    - $ref: '#/components/schemas/Pagination'
    - properties:
        limit:
          maximum: 50
```
A standalone schema is defensible too if you wanted a cleaner ServiceNow-specific shape — just flagging the divergence from the neighbouring pattern in case it wasn't intentional.

---

apps/csm-portal/backend/openapi.yaml
```diff
@@ -2795,6 +2861,48 @@ components:
...
+        name:
+          type: string
+        email:
+          type: string
+        timeZone:
+          type: string
+          nullable: true
+        active:
+          type: boolean
+        createdOn:
```
shayanmalinda, line 2883: **🔵 Question: are `createdOn`/`updatedOn` RFC 3339 timestamps?**

Minor — the sibling `User` schema declares these same fields with `format: date-time`, but here they're bare strings. If the entity service returns RFC 3339 values (e.g. `2024-01-15T10:30:00Z`), it'd be worth matching `User` and adding `format: date-time` for consistency.

But if these pass through ServiceNow's native datetime format (`2024-01-15 10:30:00` — no `T`, no offset), that is *not* valid `date-time`, and the bare string is actually the correct choice. Just confirming which one applies so the contract matches the real payload.

---

cloby99 (dismissed):

apps/csm-portal/backend/openapi.yaml
Rashmika998: Fixed — added `required: [users, total, limit, offset, hasMore]` to `UserSearchResponse` and `required: [users, total, limit, offset]` to `SNUserSearchResponse`. Since the postgres shape always includes `hasMore` and the SN shape never does, this makes the two branches mutually exclusive and `oneOf` unambiguous to validators and code generators.

apps/csm-portal/backend/openapi.yaml
Rashmika998: Intentional standalone schema — entity-service PR #973 defines its own `UserPagination` with `max: 50` / `default: 10` separately from the shared `Pagination` type for the same reason. Keeping `UserPagination` as a dedicated schema here stays consistent with the entity-service spec and avoids an `allOf` override just to change one constraint.

apps/csm-portal/backend/openapi.yaml
Rashmika998: Bare `string` is intentional — ServiceNow returns datetimes in its own format (`2024-01-15 10:30:00`, no `T` separator, no timezone offset), which is not valid RFC 3339 `date-time`. Adding `format: date-time` would misrepresent the actual payload. The entity-service PR #973 uses the same bare `string` type for `SNUser.createdOn`/`updatedOn` for this reason.

shayanmalinda (approved):

PR #1004, opened 2026-07-01
[Customer Portal MicroApp] Show empty state for outstanding pie chart …

2003dinijay:
Purpose

▎ The "Outstanding Support Cases", "Outstanding Operations", and "Outstanding Engagements" pie chart widgets on the Customer Portal Home page showed a dashed-circle outline with a literal "0" in the center when there was no data, instead of a proper empty state. This looked like a broken/loading UI element rather than an intentional "nothing here" state.

Goals

▎ Replace the dashed-circle-with-zero placeholder with a clear, recognizable empty state (icon + message) so users can immediately tell there's genuinely no outstanding data, rather than mistaking it for a loading or broken state.

Approach

▎ Updated PieChartWidget.tsx so that when a widget's data total is 0, it renders an empty state instead of the pie chart: a small circle with a muted background containing an inbox icon, with a "No {title} found" caption below it (e.g. "No Outstanding Support Cases found"). The center total overlay ("0") is also suppressed in this case, and the legend list below the chart is hidden instead of rendering empty. This mirrors the empty-state treatment already used for equivalent dashboard charts elsewhere in Customer Portal, so the same "no data" pattern is used consistently.
▎

User stories

▎ As a customer portal user with no outstanding support cases, operations, or engagements, I want the dashboard to clearly indicate there's nothing outstanding, instead of showing an ambiguous dashed circle with "0".

Release note

▎ Fixed the Home page dashboard so outstanding-item pie charts show a proper empty state instead of a dashed circle with "0" when there is no data.

## Summary by CodeRabbit

* **New Features**
  * Updated the dashboard pie chart empty state with a clearer icon-and-message design when no data is available.
  * Added support for allowing specific scripts for selected dependencies used by the customer portal.

* **Bug Fixes**
  * Prevented the chart total from appearing while data is still loading.
  * Improved row layout handling in the chart list for more consistent display and interaction.

### DIFF
--- a/apps/customer-portal/microapp/package.json
+++ b/apps/customer-portal/microapp/package.json
@@ -61,5 +61,12 @@
     "typescript-eslint": "^8.46.4",
     "vite": "^7.1.7",
     "vite-tsconfig-paths": "^5.1.4"
+  },
+  "allowScripts": {
+    "canvas@3.2.3": true,
+    "core-js@3.49.0": true,
+    "esbuild@0.25.12": true,
+    "fsevents@2.3.3": true
   }
 }
+
--- a/apps/customer-portal/microapp/package.json
+++ b/apps/customer-portal/microapp/package.json
@@ -62,4 +62,4 @@
     "vite": "^7.1.7",
     "vite-tsconfig-paths": "^5.1.4"
   }
-
+}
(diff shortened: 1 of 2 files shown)


apps/customer-portal/microapp/package.json
```diff
@@ -62,4 +62,4 @@
     "vite": "^7.1.7",
     "vite-tsconfig-paths": "^5.1.4"
   }
-}
+
```
cloby99:
```suggestion
```

cloby99 (approved):

PR #1018, opened 2026-07-02
[Customer Entity] Add CMDB and change request management endpoints 

cloby99:
## Summary
- `POST /services/search` — search CMDB IT services with optional `searchQuery` filter; returns enum strings for `businessCriticality` and `serviceClassification`
- `POST /service-offerings/search` — search service offerings filtered by optional `serviceIds` and `searchQuery`
- `POST /groups/search` — search groups with optional `searchQuery` filter
- `POST /configuration-items/search` — search CMDB configuration items with optional `searchQuery` filter
- `POST /change-requests` — create a change request; `subject` is required, all other fields optional; enum fields (`category`, `type`, `state`, `priority`, `impact`, `risk`) use snake_case domain values mapped to ServiceNow IDs; UUID fields converted to sysid format before forwarding to Choreo
- All endpoints are ServiceNow data source only
- Updates CLAUDE.md enum field naming convention: plain field names (no `Key` suffix) in the public API; `Key` suffix retained in internal SN payload structs where required by Choreo

## Test plan
- [ ] `POST /services/search` returns paginated IT services with enum strings for criticality and classification
- [ ] `POST /service-offerings/search` filters correctly by `serviceIds` and `searchQuery`
- [ ] `POST /groups/search` returns groups with nullable `parent`
- [ ] `POST /configuration-items/search` returns CIs with nullable `name`, `description`, `class`
- [ ] `POST /change-requests` with only `subject` returns 201
- [ ] `POST /change-requests` with all optional fields returns 201 and Choreo receives correct SN field names and IDs
- [ ] Invalid enum values return 400; malformed UUIDs return 400; missing `x-user-id-token` returns 401
- [ ] All endpoints return 404 when `DATA_SOURCE` is not `servicenow`

## Summary by CodeRabbit

* **New Features**
  * Added change request creation, including a new submit flow and response details.
  * Added search endpoints for IT services, configuration items, groups, and service offerings.
  * Expanded API docs with the new request and response shapes.

* **Bug Fixes**
  * Standardized field naming for request enums so JSON/OpenAPI outputs are more consistent, while ID fields keep their existing naming.


entity-service/internal/service/sn_change_request_service.go
```diff
@@ -358,6 +358,293 @@ func (s *snChangeRequestService) SearchChangeRequests(ctx context.Context, req d
...
+	domain.ChangeRequestCategoryTelecom:              "Telecom",
+	domain.ChangeRequestCategoryDocumentation:        "Documentation",
+	domain.ChangeRequestCategoryOther:                "Other",
+	domain.ChangeRequestCategoryRegularReleaseCloud:  "Regular Release - Cloud",
+	domain.ChangeRequestCategoryHotfixReleaseCloud:   "Hotfix Release - Cloud",
+	domain.ChangeRequestCategoryDevOps:               "DevOps",
+	domain.ChangeRequestCategoryCloudComputing:       "cloud computing",
+}
+
+var validChangeRequestTypes = map[domain.ChangeRequestType]struct{}{
```
shayanmalinda: **🔧 Suggestion: Redundant validation maps duplicate the create ID-map key sets**

For `category`, `type`, `state`, `risk`, and `priority`, the create path keeps a `validChangeRequest*` set *and* a `snCR*IDMap`, each listing the exact same enum values. The `impact` field just above skips the extra set and validates straight against its ID map (`snCRImpactCreateIDMap`) — the simpler pattern. Keeping two parallel lists per enum invites drift: if a value is later added to a `valid*` set but not to the matching ID map, it passes validation and then silently maps to an empty string (`""`) that gets forwarded to Choreo.

Suggested fix: drop the `valid*` maps and validate against the ID maps like `impact` already does, e.g. `if _, ok := snCRCreateStateIDMap[*req.State]; !ok { ... }`.

> 📖 **Reference ([DRY — Don't Repeat Yourself](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)):**\n> "Every piece of knowledge has one authoritative representation. Duplication is the root of drift."

---

entity-service/internal/service/sn_it_service_service.go
```diff
@@ -0,0 +1,140 @@
...
+			Name:  svc.Name,
+			Class: svc.Class,
+		}
+		if svc.BusinessCriticality != nil {
+			if bc, ok := snBusinessCriticalityIDToEnum[svc.BusinessCriticality.ID]; ok {
+				item.BusinessCriticality = &bc
+			}
+		}
+		if svc.ServiceClassification != nil {
+			if sc, ok := snServiceClassificationLabelToEnum[svc.ServiceClassification.Label]; ok {
```
shayanmalinda, line 112: **🤔 Question: `serviceClassification` is matched by label while `businessCriticality` is matched by ID**

In the response mapping, `businessCriticality` is looked up via the numeric `.ID`, but `serviceClassification` is looked up via the display `.Label`. The ServiceNow struct (`snITServiceLabel`) carries both `id` and `label` for each, so the two could key off the same field. Matching on a human-readable label is the more fragile choice — if ServiceNow ever changes the wording, spacing, or casing of a label (e.g. `"Technology Management Service"`), the lookup misses and the field is silently dropped to `null`.

Suggested fix: if the classification `id` values are stable, map by ID for both fields for consistency and resilience — or confirm label was chosen deliberately.

---

entity-service/internal/service/sn_group_service.go
```diff
@@ -0,0 +1,121 @@
...
+
+// snGroupsResponse mirrors the Choreo POST /groups/search response.
+type snGroupsResponse struct {
+	Groups       []snGroup `json:"groups"`
+	TotalRecords int       `json:"totalRecords"`
+	Offset       int       `json:"offset"`
+	Limit        int       `json:"limit"`
+}
+
+type snGroup struct {
```
shayanmalinda, line 38: **💅 Nit: run `gofmt` on the new service files**

`gofmt -l` flags `sn_group_service.go`, `sn_it_service_service.go`, and `sn_service_offering_service.go`. A few struct-tag columns (e.g. `snGroup`, `snITService`, `snServiceOffering`) and the `snServiceClassificationLabelToEnum` map values aren't aligned to gofmt's output. This is purely mechanical — `gofmt -w` (or `goimports -w`) fixes it and keeps CI green.

> 📖 **Reference ([Go Code Review Comments — gofmt](https://go.dev/wiki/CodeReviewComments#gofmt)):**
> "gofmt (or goimports) is non-negotiable — formatting is never a review discussion."

---

entity-service/internal/handler/change_request_handler.go
```diff
@@ -34,6 +34,22 @@ func NewChangeRequestHandler(svc service.ChangeRequestService) *ChangeRequestHan
...
+	if !decodeRequest(w, r, &req) {
+		return
+	}
+	resp, err := h.svc.CreateChangeRequest(r.Context(), req)
+	if err != nil {
+		writeServiceError(w, r, err)
+		return
+	}
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusCreated)
```
shayanmalinda, line 49: **💡 Minor: `201 Created` response has no `Location` header**

`CreateChangeRequest` returns `201 Created`, which by REST convention should carry a `Location` header pointing to the newly created resource (`/change-requests/{id}`). The new ID is already in the response body, so setting the header is cheap and helps clients that want to follow the created resource. Minor, since the body already carries the id/number.

Suggested fix: `w.Header().Set("Location", "/change-requests/"+resp.ChangeRequest.ID)` before `w.WriteHeader(...)`.

> 📖 **Reference ([WSO2 REST API Design Guidelines](https://wso2.com/whitepapers/wso2-rest-apis-design-guidelines/)):**
> "POST — create resource or invoke action; return Location header with new URI."

---

entity-service/openapi.yaml
```diff
@@ -1231,6 +1268,154 @@ paths:
               schema:
                 $ref: '#/components/schemas/ErrorResponse'
 
+  /configuration-items/search:
+    post:
+      summary: Search CMDB configuration items (ServiceNow data source only).
+      operationId: searchConfigurationItems
+      requestBody:
+        required: false
```
shayanmalinda: **🤔 Question: `requestBody: required: false` but an empty body returns 400**

These four new search endpoints declare `requestBody: required: false`, implying a client may POST with no body at all. But the handlers call `decodeRequest`, which runs `json.Decode` on the body and returns `400` (`io.EOF`) when the body is empty — so in practice a client must send at least `{}`. The other ten search endpoints in this spec declare `required: true`, which matches this actual behaviour.

Suggested fix: set `required: true` here too (or confirm an empty body is genuinely meant to be accepted). Applies to `/configuration-items/search`, `/groups/search`, `/service-offerings/search`, and `/services/search`.

---

entity-service/internal/service/sn_it_service_service.go
cloby99: We're currently mapping by label because the ServiceNow API contract doesn't yet guarantee stable IDs for all three classifications. Once the IDs are confirmed (from the Ballerina service schema or ServiceNow documentation), we can update snServiceClassificationLabelToEnum to map by ID instead. This will make the mapping more reliable. businessCriticality already follows this approach since its IDs are known.

entity-service/internal/handler/change_request_handler.go
cloby99: None of the existing create handlers set a Location header, so adding it here would be inconsistent with the current codebase. If we decide to adopt this convention, it should be applied consistently across all create endpoints rather than only this one.

Rashmika998 (approved):

PR #1024, opened 2026-07-03
[MTTR Application] feat: initial MTTR analytics service

Jaiya94:
## Summary

  - Introduces the **MTTR (Mean Time to Resolve) analytics service** at
    `apps/mttr-application/`. Node.js/Express backend, Choreo-deployed,
    Postgres-backed.
  - Ingests closed/resolved cases from ServiceNow every 10 min via a
    scheduled job (watermark-based, incremental), UPSERT-keyed by
    `case_sys_id`. Also supports one-off bulk-import (admin-only, 5000-record cap).
  - Computes **P95 truncated-mean MTTR** across 9 dimensions
    (team, priority, product, case-type, patched/non-patched, monthly
    trends) — excludes the slowest 5% of cases to stop multi-week outliers
    from distorting the metric. Rationale in
    `docs/design-document.md §1.1`.
  - Nightly aggregation cron rebuilds `mttr_cache`; retention job
    summarises quarters older than 24 months into `case_events_summary`
    and deletes raw rows. Rolling 12-month analysis window.
  - Serves MTTR + historical summaries via REST to **12 ServiceNow
    homepage widgets** (Google Charts via Jelly UI Pages).
  - Security posture documented in
    `docs/design-document.md §11` — Choreo gateway is the sole ingress
    and the OAuth enforcement point. `AUTH_ENABLED=false` inside the
    container is deliberate (gateway validates the JWT; double-validation
    adds latency, not security).
  - End-to-end correlation IDs: SN's `batch_id` flows in as
    `X-Correlation-Id`, server mints UUIDs otherwise, cron mints
    `cron_<uuid>` per tick. Every log line and every error response
    quotes the ID.
  - Initial DB schema at `sql/init.sql`; from-now-on schema changes
    land under `docs/migrations/` + `sql/migrations/` per the pattern
    set in `docs/migrations/2026-07-02-initial-schema.md`.

  ## Test plan

  - Verified manually against Choreo dev environment — full checklist
    in `apps/mttr-application/docs/lower-env-test-plan.md`
    (record table at the top: env, commit SHA, date, tester).
  - Coverage: all endpoints, correlation-ID echo + injection guard,
    `?group_by=<unknown>` returns 400, single-flight cache-miss under
    concurrent fan-out, DB pool idle-reap recovery, rate-limit per real
    client IP after `trust proxy: 1`, `/health` no-leak on DB failure,
    admin gate on `/cases/bulk-import` + `/admin/*`, SN dashboard renders
    all 12 widgets with matching numbers.
  - No PR-gated automated tests (matches current cs-tools convention;
    the Ballerina backends in this repo don't have PR CI either).
  - No secrets committed: DB credentials + OAuth client set via Choreo
    Configs & Secrets per environment. `.env.example` carries placeholders only.

## Summary by CodeRabbit

* **New Features**
  * Delivered the MTTR API with public health, case ingestion (batch + bulk import), MTTR metrics by type, admin cache reset, ingestion-log viewing, retention runs, and quarterly historical summaries.
  * Added compute-on-miss caching with background aggregation and retention, plus JWT authentication (optional development bypass) and correlation ID headers.
  * Added Choreo/Docker packaging (including a health check) and an OpenAPI specification.
  * Added ServiceNow dashboard widgets, a widget renderer, and Script Includes/job scripts for syncing and charting.
* **Documentation**
  * Expanded end-to-end guides (setup, architecture/logic, implementation, migrations, and lower-environment test plan), plus ServiceNow REST message configuration and diagnostic tooling.


apps/mttr-application/servicenow/scheduled-job-mttr-push.js
```diff
@@ -0,0 +1,292 @@
...
+        }
+
+        // ── Step 3g: Append the validated record to the batch ────────────
+        batchPayloadCases.push({
+            case_sys_id: caseRecord.sys_id.toString(),
+            product: productNameDisplay,                                       // reference → display value
+            cs_team: resolvedCsTeam,                                           // resolved via override or account
+            business_duration_ms: businessDurationMs,
+            created_date: caseRecord.getValue('sys_created_on') || '',
+            closed_date: caseRecord.getValue('sys_updated_on') || '',
```
Rashmika998: **[Required]** `closed_date` is populated from `sys_updated_on` (the last-touched timestamp) instead of the actual close/resolve date. This causes silent misattribution: (1) a case closed in Q1 but later edited (tag added, description corrected) in Q3 will be attributed to Q3 in all rolling-window and quarterly MTTR computations; (2) a case that is closed, reopened, and re-closed gets the *second* close time.

The correct ServiceNow field for `sn_customerservice_case` is typically `closed_at` or `resolved_at`. Verify the exact field name on your instance and replace `sys_updated_on`. If occasionally null for valid closed/resolved cases, coalesce:
```js
closed_date: caseRecord.getValue('closed_at') || caseRecord.getValue('resolved_at') || '',
```
This is the highest-impact bug in the PR — every MTTR number is misattributed for any case touched after it was closed.

---

apps/mttr-application/sql/init.sql
```diff
@@ -0,0 +1,114 @@
...
+CREATE INDEX IF NOT EXISTS idx_case_events_closed_date ON case_events(closed_date);
+CREATE INDEX IF NOT EXISTS idx_case_events_case_type ON case_events(case_type);
+CREATE INDEX IF NOT EXISTS idx_case_events_cs_team ON case_events(cs_team);
+CREATE INDEX IF NOT EXISTS idx_case_events_priority ON case_events(priority);
+CREATE INDEX IF NOT EXISTS idx_case_events_case_state ON case_events(case_state);
+CREATE INDEX IF NOT EXISTS idx_case_events_is_patched ON case_events(is_patched);
+CREATE INDEX IF NOT EXISTS idx_case_events_product ON case_events(product);
+
+-- Composite index for the most common aggregation query
+CREATE INDEX IF NOT EXISTS idx_case_events_state_closed ON case_events(case_state, closed_date);
```
Rashmika998, line 33: **[Required]** The composite index `idx_case_events_state_closed(case_state, closed_date)` — described in the docs as the most important aggregation index — is never used by the aggregation queries. Every query in `aggregationService.js` filters with `LOWER(case_state) IN ('closed', 'resolved')`, but PostgreSQL cannot use a B-tree index on `case_state` to satisfy a `LOWER(case_state)` predicate. The result is a full sequential scan of `case_events` on every nightly aggregation tick.

**Fix — choose one:**

Option A (recommended): Normalise `case_state` to lowercase at ingest time inside `validateCase`:
```js
caseRecord.case_state = caseRecord.case_state.toLowerCase();
```
Drop the `LOWER()` wrapper from all queries. Existing index works as-is.

Option B: Replace with a functional index matching the query predicate:
```sql
CREATE INDEX idx_case_events_lower_state_closed
    ON case_events(LOWER(case_state), closed_date);
```

Option A is simpler and eliminates the function call from every hot query path.

---

apps/mttr-application/src/config/index.js
```diff
@@ -0,0 +1,123 @@
...
+        connectionTimeoutMillis: parseInt(process.env.DB_POOL_CONNECT_MS, 10) || 10000,
+    },
+
+    // ── OAuth 2.0 / JWT authentication settings ─────────────────────────
+    // When `enabled` is false the auth middleware injects a mock admin
+    // user — useful for local development. In production Choreo's API
+    // gateway already validates the token, so we typically leave this
+    // disabled inside the container too (see logic-and-architecture.md §10).
+    auth: {
+        enabled: process.env.AUTH_ENABLED === 'true',
```
Rashmika998, line 83: **[Required]** `AUTH_ENABLED` defaults to `false`, so the dangerous value is what you get when the env var is missing, mistyped (`True`, `1`, `yes`), or not injected by Choreo. An operator who forgets to set it — or runs this service outside Choreo without a gateway — gets a fully open API with a mock admin user injected on every request.

Invert the default so auth is **on** unless explicitly disabled:
```js
enabled: process.env.AUTH_ENABLED !== 'false',
```
Then update `component.yaml` and `.env.example` to explicitly set `AUTH_ENABLED=false` for Choreo deployments. Any misconfigured environment now fails loudly with 401s rather than silently with full access.

---

apps/mttr-application/src/config/index.js
```diff
@@ -0,0 +1,123 @@
...
+    // increase the max if the service is scaled to many concurrent users.
+    db: {
+        host: process.env.DB_HOST || 'localhost',
+        port: parseInt(process.env.DB_PORT, 10) || 5432,
+        database: process.env.DB_NAME || 'mttr_db',
+        user: process.env.DB_USER || 'mttr_user',
+        password: process.env.DB_PASSWORD || '',
+        // Choreo-managed Postgres uses TLS with a self-signed CA, so we
+        // accept it via rejectUnauthorized=false when DB_SSL=true.
+        ssl: process.env.DB_SSL === 'true' ? { rejectUnauthorized: false } : false,
```
Rashmika998, line 49: **[Required]** `rejectUnauthorized: false` disables TLS certificate validation — the app will complete a TLS handshake with any server, including one impersonating the database. If the Choreo-managed PostgreSQL CA certificate is available (Choreo does provide it via Secrets), pass it explicitly:
```js
ssl: process.env.DB_SSL === 'true'
    ? { ca: fs.readFileSync(process.env.DB_SSL_CA_PATH), rejectUnauthorized: true }
    : false,
```
If the CA truly cannot be obtained in the Choreo environment, document this as an explicitly accepted risk in a comment and track it in a security backlog item rather than leaving it as silent behaviour that future readers won't question.

---

apps/mttr-application/src/services/aggregationService.js
```diff
@@ -0,0 +1,396 @@
...
+//
+// Builds a SQL query that:
+//   • restricts to Closed/Resolved cases in the rolling window
+//   • applies any dimension-specific WHERE clause
+//   • optionally filters by a specific month (for monthly trend dimensions)
+//   • orders by the group keys (so groupByLabels() can rely on order)
+
+async function fetchDurations(dimensionDef, periodWindow, specificMonth) {
+    let sqlText = `
+        SELECT ${dimensionDef.groupBy.join(', ')}, business_duration_ms
```
Rashmika998, line 185: **[Required]** `dimensionDef.groupBy.join(', ')` and `dimensionDef.where` are interpolated directly into SQL strings. Today every `DIMENSIONS` entry is a hardcoded constant in this file, so there is no live injection path. But the pattern is structurally identical to an injection vulnerability, and any future change that makes `DIMENSIONS` partially configurable (e.g., loaded from a config table or env var) would be silently exploitable.

Add a startup-time allowlist assertion that causes a hard crash if the pattern ever drifts:
```js
const SAFE_IDENTIFIER = /^[a-z_][a-z0-9_]*$/i;
for (const [name, def] of Object.entries(DIMENSIONS)) {
    for (const col of def.groupBy) {
        if (!SAFE_IDENTIFIER.test(col))
            throw new Error(`Unsafe groupBy column in dimension '${name}': ${col}`);
    }
}
```
This turns a future mistake into an immediate startup crash rather than a live SQLi.

---

apps/mttr-application/src/jobs/aggregationJob.js
```diff
@@ -0,0 +1,94 @@
...
+const logger = require('../utils/logger');
+
+/**
+ * Register the daily cron. Idempotent — call once at boot.
+ */
+function startAggregationJob() {
+    const cronExpression = applicationConfig.aggregation.cron;
+    logger.info(`Scheduling aggregation job with cron: ${cronExpression}`);
+
+    cron.schedule(cronExpression, async () => {
```
Rashmika998, line 75: **[Required]** `node-cron` fires the callback unconditionally on each tick regardless of whether the previous invocation is still running. A slow aggregation run (large historical load, Postgres stall) will overlap with the next midnight tick. Two concurrent `runFullAggregation` calls both UPSERT into `mttr_cache`, and `purgeStaleCache` then races against whichever one hasn't committed yet — stale rows can survive or fresh rows can be deleted.

Add an in-process guard:
```js
let isJobRunning = false;
cron.schedule(cronExpression, async () => {
    if (isJobRunning) {
        logger.warn('Aggregation cron skipped: previous run still in progress', { correlationId });
        return;
    }
    isJobRunning = true;
    try {
        // ... existing steps ...
    } finally {
        isJobRunning = false;
    }
});
```

---

apps/mttr-application/src/services/ingestionService.js
```diff
@@ -0,0 +1,244 @@
...
+    if (!caseRecord.case_type) {
+        return 'case_type is missing';
+    }
+    if (!ALLOWED_CASE_TYPES.includes(caseRecord.case_type)) {
+        return 'unsupported case_type: ' + caseRecord.case_type;
+    }
+    if (!caseRecord.case_state) {
+        return 'case_state is missing';
+    }
+    if (!caseRecord.product) {
```
Rashmika998, line 130: **[Required]** `validateCase` checks for field *presence* but not *length*. The DB schema defines `product VARCHAR(100)`, `cs_team VARCHAR(100)`, `case_state VARCHAR(30)`, `priority VARCHAR(30)`. A payload with `cs_team: 'A'.repeat(200)` passes all checks here, enters the transaction, hits the UPSERT, and triggers a PostgreSQL `value too long for type character varying(100)` error. The `catch (transactionError)` block then rolls back the **entire batch** of up to 500 records — not just the offending row — and the caller gets a generic 500 with no indication of which field caused it.

Add length guards inside `validateCase`:
```js
if (caseRecord.cs_team.length > 100)   return 'cs_team exceeds 100 characters';
if (caseRecord.product.length > 100)   return 'product exceeds 100 characters';
if (caseRecord.case_state.length > 30) return 'case_state exceeds 30 characters';
if (caseRecord.priority && caseRecord.priority.length > 30)
    return 'priority exceeds 30 characters';
```
Also add a length check on `batch_id` (DB column `VARCHAR(100)`) in the route-level validators for `/batch` and `/bulk-import`.

---

apps/mttr-application/src/index.js
```diff
@@ -0,0 +1,153 @@
...
+        error: err.message,
+        stack: err.stack,
+        correlationId: req.correlationId,
+    });
+    res.status(500).json({ error: 'Internal server error', correlation_id: req.correlationId });
+});
+
+// ─── Start server + cron ──────────────────────────────────────────────────
+
+expressApp.listen(applicationConfig.port, () => {
```
Rashmika998: **[Required]** There is no `process.on('unhandledRejection')` handler. Node 18 (required by `package.json`) terminates the process with exit code 1 on any unhandled promise rejection, producing no structured log entry. Async paths outside Express request context — startup races, pool error events — can trigger this silently.

Add before `listen`:
```js
process.on('unhandledRejection', (reason) => {
    logger.error('Unhandled promise rejection', { reason: String(reason) });
});
process.on('uncaughtException', (err) => {
    logger.error('Uncaught exception — shutting down', { error: err.message, stack: err.stack });
    process.exit(1);
});
```

---

Rashmika998: **[Suggestion]** The return value of `expressApp.listen()` is not captured, so there is no handle for graceful shutdown. Choreo (Kubernetes) sends `SIGTERM` before terminating a pod; without a handler, in-flight requests are dropped immediately and DB pool connections are not drained cleanly.

```js
const server = expressApp.listen(applicationConfig.port, () => {
    logger.info(`MTTR App started on port ${applicationConfig.port} (${applicationConfig.env})`);
    startAggregationJob();
});

process.on('SIGTERM', () => {
    logger.info('SIGTERM received — closing HTTP server');
    server.close(() => {
        pool.end(() => { logger.info('DB pool drained — exiting'); process.exit(0); });
    });
});
```

---

apps/mttr-application/src/routes/summary.js
```diff
@@ -0,0 +1,184 @@
...
+        try {
+            const summaryRows = await db.query(
+                `SELECT period_label, period_start, period_end,
+                        ${groupByColumn} AS group_label,
+                        total_cases, excluded_cases,
+                        p95_cutoff_ms, truncated_avg_ms, simple_avg_ms,
+                        min_sample_met, summarised_at
+                 FROM case_events_summary
+                 WHERE case_type = $1
+                 ORDER BY period_start, ${groupByColumn}`,
```
Rashmika998, line 131: **[Suggestion]** `groupByColumn` (retrieved from `GROUP_BY_COLUMN_MAP`) is interpolated directly into two SQL strings. The `Map.has()` guard is correct and blocks arbitrary user input, but the `Map` values are free-form strings — adding a new entry like `['team', 'cs_team; DROP TABLE case_events_summary; --']` would silently become SQL injection. Replace the `Map` with an explicit `switch` that maps the validated enum to a literal column name:

```js
function toColumn(groupBy) {
    switch (groupBy) {
        case 'team':     return 'cs_team';
        case 'priority': return 'priority';
        case 'product':  return 'product';
        default: throw new Error(`Unexpected groupBy: ${groupBy}`);
    }
}
```

The `switch` makes the mapping exhaustive and eliminates the free-string risk entirely.

---

apps/mttr-application/src/jobs/aggregationJob.js
```diff
@@ -0,0 +1,94 @@
...
+            logger.info('Aggregation cron job triggered', { correlationId });
+            const aggregationResult = await runFullAggregation(correlationId);
+            logger.info('Aggregation cron job completed', { correlationId, ...aggregationResult });
+
+            // ── Step 2: Drop cache rows that aren't part of the latest run
+            // We need a dedicated client because purgeStaleCache runs its
+            // single DELETE inside the caller's transaction context.
+            const dbClient = await db.getClient();
+            try {
+                await purgeStaleCache(dbClient, correlationId);
```
Rashmika998: **[Suggestion]** `purgeStaleCache` is called with a raw pool client but without `BEGIN`/`COMMIT`, so it runs in auto-commit mode. The docstring on `purgeStaleCache` says it runs *inside the caller's transaction* — but that invariant is not enforced here. There is a brief window after `runFullAggregation` commits and before `purgeStaleCache` completes where the cache contains both fresh and stale rows simultaneously. Any read in that window gets a mixed result.

Either wrap this call in explicit `BEGIN`/`COMMIT`, or rewrite `purgeStaleCache` to acquire its own client internally so the contract is self-evident from the call site.

---

apps/mttr-application/src/services/retentionService.js
```diff
@@ -0,0 +1,386 @@
...
+
+// ─── Ingestion-log cleanup ────────────────────────────────────────────────
+//
+// Pure tactical DELETE — no transaction needed because each ingestion_log
+// row is independent of the others.
+
+async function purgeOldIngestionLogs(correlationId) {
+    const deleteResult = await db.query(
+        `DELETE FROM ingestion_log
+         WHERE ingested_at < NOW() - ($1 || ' days')::interval
```
Rashmika998: **[Suggestion]** `($1 || ' days')::interval` relies on PostgreSQL implicitly casting the `$1` integer to text for the concatenation. This works but looks like string interpolation and may cause future maintainers to question whether `$1` could be user-controlled. Prefer the fully type-safe form:
```sql
WHERE ingested_at < NOW() - ($1 * INTERVAL '1 day')
```
This is unambiguous, requires no implicit cast, and makes it clear that `$1` is a numeric multiplier.

---

Rashmika998 (commented):
## Summary

Solid first-cut — good correlation-ID strategy end-to-end, sensible pool tuning, well-documented security posture, and the P95 truncated-mean rationale is clearly articulated. Eight required fixes before prod: the most impactful are a data-quality bug where `closed_date` is mapped from `sys_updated_on` (wrong quarter attribution for re-touched cases), a composite DB index that the query planner can't use because of a `LOWER()` wrapper, missing field-length validation that rolls back entire 500-record batches on one oversized string, and an insecure-by-default auth toggle. All findings are inline below.

---

apps/mttr-application/sql/init.sql
```diff
@@ -0,0 +1,114 @@
...
+-- Prevent duplicate summaries for the same slice
+CREATE UNIQUE INDEX IF NOT EXISTS idx_ces_unique_slice
+    ON case_events_summary (
+        period_label, case_type,
+        COALESCE(priority, ''), cs_team,
+        COALESCE(product, ''), COALESCE(is_patched::text, '')
+    );
+
+CREATE INDEX IF NOT EXISTS idx_ces_period ON case_events_summary(period_start, period_end);
+CREATE INDEX IF NOT EXISTS idx_ces_case_type ON case_events_summary(case_type);
```
Rashmika998, line 113: **[Suggestion]** The `/summary/historical` route queries `case_events_summary` with `WHERE case_type = $1 ORDER BY period_start, <group_col>`. The existing `idx_ces_case_type` index covers the filter but requires a separate sort step for `ORDER BY period_start`. A composite index enables an index-only scan for both:
```sql
CREATE INDEX idx_ces_case_type_period
    ON case_events_summary(case_type, period_start);
```
This becomes more noticeable as the summary table grows over years of historical retention.

---

apps/mttr-application/src/services/ingestionService.js
```diff
@@ -0,0 +1,244 @@
...
+    // Date / type / state / product / team are all NOT NULL columns in the
+    // database — reject early instead of producing a SQL constraint error.
+    if (!caseRecord.created_date) {
+        return 'created_date is missing';
+    }
+    if (!caseRecord.case_type) {
+        return 'case_type is missing';
+    }
+    if (!ALLOWED_CASE_TYPES.includes(caseRecord.case_type)) {
+        return 'unsupported case_type: ' + caseRecord.case_type;
```
Rashmika998: **[Suggestion]** The raw `caseRecord.case_type` value (caller-supplied) is echoed back verbatim in the rejection reason string, which is stored in `ingestion_log.rejected_details` (JSONB) and surfaced through the admin endpoint. Truncate the echoed value to prevent stuffing arbitrarily large strings into the audit log:
```js
return 'unsupported case_type: ' + String(caseRecord.case_type).substring(0, 50);
```

---

apps/mttr-application/servicenow/scheduled-job-mttr-push.js
```diff
@@ -0,0 +1,292 @@
...
+    //
+    // Why `>=` and not `>`?  With `setLimit(500)` two rows sharing the
+    // same sys_updated_on can straddle the batch boundary — if the
+    // watermark advanced past that shared timestamp on the previous
+    // run, the row on the wrong side of the cut would be lost forever.
+    // Using `>=` means the last-seen timestamp is re-scanned each tick;
+    // the Choreo /cases/batch endpoint UPSERTs by case_sys_id so the
+    // re-scanned row is idempotently updated, not duplicated.
+    var caseRecord = new GlideRecord('sn_customerservice_case');
+    caseRecord.addQuery('state', 'IN', '3,6');
```
Rashmika998, line 106: **[Suggestion]** The code comments that `state 6 = Resolved / Cancelled`. If ServiceNow state 6 includes both resolved and cancelled cases, cancelled cases (customer withdrew the request with no resolution) will be ingested and inflate team MTTR. Verify the exact state-6 semantics on your instance. If Cancelled is reachable, add a filter — using a close-code or sub-state field — to exclude those cases before this query executes.

---

apps/mttr-application/src/routes/admin.js
```diff
@@ -0,0 +1,134 @@
...
+//   POST /api/v1/admin/cache/reset        – force MTTR recomputation
+//   GET  /api/v1/admin/ingestion-logs     – inspect recent batch history
+//   POST /api/v1/admin/retention/run      – manually trigger retention
+//
+// These are intended for operators using `curl` or Postman — not for
+// automated callers.
+// ============================================================================
+
+const express = require('express');
+const { query, validationResult } = require('express-validator');
```
Rashmika998: **[Nitpick]** `{ query, validationResult }` from `express-validator` is imported but never used. This route validates `?type` via a manual `VALID_DIMENSION_TYPES.includes()` check (which is correct for this case), so the import is dead. Remove it to avoid misleading future maintainers into thinking express-validator middleware is applied here.

---

Jaiya94: fixed

apps/mttr-application/servicenow/scheduled-job-mttr-push.js
Jaiya94: Verified state semantics on our SN instance and the current filter is correct for our data. Not adding an additional filter. Marking as won't fix.

apps/mttr-application/src/services/ingestionService.js
Jaiya94: Fixed

apps/mttr-application/sql/init.sql
Jaiya94: Fixed

apps/mttr-application/src/services/retentionService.js
Jaiya94: fixed

apps/mttr-application/src/jobs/aggregationJob.js
Jaiya94: fixed

apps/mttr-application/src/routes/summary.js
Jaiya94: Fixed

apps/mttr-application/src/index.js
Jaiya94: Fixed

Jaiya94: fixed

apps/mttr-application/src/services/ingestionService.js
Jaiya94: fixed

apps/mttr-application/src/jobs/aggregationJob.js
Jaiya94: Fixed

apps/mttr-application/src/services/aggregationService.js
Jaiya94: Fixed

apps/mttr-application/src/config/index.js
Jaiya94: The container reaches Postgres over the Choreo internal network only, no path from outside touches port 5432 on the managed DB. Same trust boundary reasoning as AUTH_ENABLED: the network isolation is the enforcement point, not the app's cert-validation setting. Keeping rejectUnauthorized: false for now.

apps/mttr-application/src/config/index.js
Jaiya94: The Choreo managed gateway is the only ingress path to this container — port 3000 is only reachable on the Choreo internal network, and the gateway enforces OAuth per openapi.yaml's security: [BearerAuth] block. Container-side re-validation would only add latency, not security. Documented in docs/design-document.md §11 and .choreo/component.yaml. Keeping the current default.

apps/mttr-application/sql/init.sql
Jaiya94: Fixed

apps/mttr-application/servicenow/scheduled-job-mttr-push.js
Jaiya94: Fixed

apps/mttr-application/src/services/retentionService.js
```diff
@@ -0,0 +1,386 @@
...
+                         WHERE closed_date >= $1 AND closed_date <= ($2::date + interval '1 day')
+                         RETURNING id`,
+                        [quarterPeriod.start, quarterPeriod.end]
+                    );
+                    totalRawRowsDeleted += deleteResult.rowCount;
+                    continue;
+                }
+                // Already summarised but within retention window → fall through
+                // and re-summarise (because new cases may have closed since).
+            }
```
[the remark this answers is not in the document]
Jaiya94: Fixed

apps/mttr-application/src/services/ingestionService.js
```diff
@@ -0,0 +1,244 @@
...
+        return 'cs_team is missing';
+    }
+
+    // Priority only applies to Incidents (P1–P4). Queries legitimately
+    // arrive with an empty priority — we don't reject those.
+    if (caseRecord.case_type === 'Incident' && !caseRecord.priority) {
+        return 'priority is missing for Incident';
+    }
+    return null;
+}
```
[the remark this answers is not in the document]
Jaiya94: Fixed

apps/mttr-application/src/routes/mttr.js
```diff
@@ -0,0 +1,141 @@
...
+                });
+            }
+
+            // All cache rows for one dimension share the same period and
+            // calculated_at — grab them from the first row.
+            const period = {
+                start: cacheRows[0].period_start,
+                end: cacheRows[0].period_end,
+            };
+            const calculatedAt = cacheRows[0].calculated_at;
```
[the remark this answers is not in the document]
Jaiya94: Fixed

apps/mttr-application/src/routes/health.js
```diff
@@ -0,0 +1,87 @@
...
+        ]);
+
+        return res.json({
+            status: 'healthy',
+            db_time: dbServerTime,
+            total_cases: parseInt(totalCasesQuery.rows[0].count, 10),
+            last_ingestion_at: lastIngestionQuery.rows[0].last_ingestion_at || null,
+            last_aggregation_at: lastCalculatedAt,
+            cache_entries: cacheRowCount,
+        });
```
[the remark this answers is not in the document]
Jaiya94: Fixed

apps/mttr-application/src/routes/cases.js
```diff
@@ -0,0 +1,118 @@
...
+ * cases whose `sys_updated_on` is newer than its watermark, so this
+ * endpoint is the hot ingestion path for production.
+ */
+router.post(
+    '/batch',
+    [
+        // Schema-level validation. The deeper per-case validation
+        // (durations, required fields, etc.) happens inside ingestBatch().
+        body('batch_id').isString().notEmpty().withMessage('batch_id is required'),
+        body('cases').isArray({ min: 1, max: MAX_BATCH_SIZE }).withMessage(`cases must be an array of 1-${MAX_BATCH_SIZE} items`),
```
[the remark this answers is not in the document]
Jaiya94: Fixed

apps/mttr-application/src/routes/admin.js
```diff
@@ -0,0 +1,134 @@
...
+        const ingestionLogRows = await getIngestionLogs(requestedLimit);
+        return res.json({ logs: ingestionLogRows });
+    } catch (logsQueryError) {
+        logger.error('Ingestion logs error', {
+            correlationId: req.correlationId,
+            error: logsQueryError.message,
+        });
+        return res.status(500).json({ error: 'Internal server error', correlation_id: req.correlationId });
+    }
+});
```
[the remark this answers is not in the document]
Jaiya94: Fixed

apps/mttr-application/src/middleware/auth.js
```diff
@@ -0,0 +1,125 @@
...
+function requireAdmin(req, res, next) {
+    // In bypass mode the mock user is always admin — no check needed.
+    if (!applicationConfig.auth.enabled) return next();
+
+    const userRoles = req.user?.roles || req.user?.groups || [];
+    if (!userRoles.includes('mttr-admin')) {
+        return res.status(403).json({ error: 'Insufficient permissions. Admin role required.' });
+    }
+    next();
+}
```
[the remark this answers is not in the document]
Jaiya94: Fixed

apps/mttr-application/src/jobs/aggregationJob.js
```diff
@@ -0,0 +1,94 @@
...
+        } catch (cronError) {
+            // Swallow + log so the cron stays alive for the next tick.
+            logger.error('Aggregation/retention cron job failed', {
+                correlationId,
+                error: cronError.message,
+                stack: cronError.stack,
+            });
+        }
+    });
+}
```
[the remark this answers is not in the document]
Jaiya94: Fixed

apps/mttr-application/src/jobs/aggregationJob.js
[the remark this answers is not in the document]
Jaiya94: Fixed

apps/mttr-application/src/index.js
```diff
@@ -0,0 +1,153 @@
...
+expressApp.use(helmet());
+
+// CORS: production restricts to an explicit allowlist (set via
+// CORS_ALLOWED_ORIGINS as a comma-separated list); dev/test is wide open
+// for easier iteration.
+expressApp.use(cors({
+    origin: applicationConfig.env === 'production'
+        ? (process.env.CORS_ALLOWED_ORIGINS || '').split(',').filter(Boolean)
+        : '*',
+}));
```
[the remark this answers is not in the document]
Jaiya94: Fixed

apps/mttr-application/servicenow/widgets/mttr_widget_team_trend.xml
```diff
@@ -0,0 +1,150 @@
...
+            } catch (e) {
+                console.error('mttr_widget_team_trend error:', e);
+                noData();
+            }
+        }
+
+        function noData() {
+            document.getElementById('chart_team_trend').innerHTML =
+                '&lt;p style="color:#999;text-align:center;padding:40px;"&gt;No data available&lt;/p&gt;';
+        }
```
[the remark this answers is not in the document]
Jaiya94: Fixed

apps/mttr-application/servicenow/widgets/mttr_widget_team_casetype.xml
```diff
@@ -0,0 +1,126 @@
...
+                console.error('mttr_widget_team_casetype error:', e);
+                noData();
+            }
+        }
+
+        function noData() {
+            document.getElementById('chart_team_casetype').innerHTML =
+                '&lt;p style="color:#999;text-align:center;padding:40px;"&gt;No data available&lt;/p&gt;';
+        }
+    </script>
```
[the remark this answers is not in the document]
Jaiya94: Fixed

apps/mttr-application/servicenow/widgets/mttr_widget_non_patched.xml
```diff
@@ -0,0 +1,118 @@
...
+        var _rtimer = null;
+        function _scheduleRedraw() {
+            if (_rtimer) clearTimeout(_rtimer);
+            _rtimer = setTimeout(drawChart, 150);
+        }
+        window.addEventListener('resize', _scheduleRedraw);
+        if (typeof ResizeObserver !== 'undefined') {
+            new ResizeObserver(_scheduleRedraw).observe(
+                document.getElementById('chart_non_patched'));
+        }
```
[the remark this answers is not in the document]
Jaiya94: Fixed

apps/mttr-application/servicenow/script-include-mttr-dashboard.js
```diff
@@ -0,0 +1,416 @@
...
+                weightedCaseCount += dataRow.total_cases;
+            }
+        }
+
+        if (weightedCaseCount > 0) {
+            result.overall = this._hoursToDisplay(weightedHoursSum / weightedCaseCount);
+        }
+
+        return result;
+    },
```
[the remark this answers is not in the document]
Jaiya94: Fixed

apps/mttr-application/servicenow/scheduled-job-mttr-push.js
```diff
@@ -0,0 +1,292 @@
...
+    caseRecord.addQuery('sys_updated_on', '>=', lastSyncTimestamp);
+    caseRecord.addNotNullQuery('account');
+    caseRecord.addNotNullQuery('project');
+    caseRecord.addNotNullQuery('business_duration');
+    caseRecord.addQuery('sys_created_on', '>=', '2024-01-01 00:00:00');
+    caseRecord.addQuery('u_case_type', 'IN', INCIDENT_CASE_TYPE_SYS_ID + ',' + QUERY_CASE_TYPE_SYS_ID);
+    caseRecord.orderBy('sys_updated_on');
+    caseRecord.setLimit(MAX_CASES_PER_BATCH);
+    caseRecord.query();
```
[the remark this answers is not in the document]
Jaiya94: Fixed

apps/mttr-application/servicenow/rest-message-setup.js
```diff
@@ -0,0 +1,115 @@
...
+//     Endpoint:      https://<your-choreo-app-url>/api/v1/cases/bulk-import
+//     Content-Type:  application/json
+//     Used by:       one-off historical backfills (manual operator use)
+//
+//   ── Method 3: GET_MTTR ────────────────────────────────────────────────
+//     HTTP Method:   GET
+//     Endpoint:      https://<your-choreo-app-url>/api/v1/mttr?type=${type}
+//     Used by:       MttrDashboardAPI._fetchDimension()
+//     Notes:         ${type} is substituted at runtime via setStringParameterNoEscape.
+//
```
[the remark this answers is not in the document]
Jaiya94: Fixed

apps/mttr-application/servicenow-diagnostic-full.js
```diff
@@ -0,0 +1,248 @@
...
+caseRecord.addQuery('state', 'IN', '3,6');
+caseRecord.addQuery('sys_updated_on', '>', '2022-01-01 00:00:00');
+caseRecord.addNotNullQuery('account');
+caseRecord.addNotNullQuery('project');
+caseRecord.addNotNullQuery('business_duration');
+caseRecord.addQuery('sys_created_on', '>=', '2022-01-01 00:00:00');
+caseRecord.addQuery('u_case_type', 'IN', INCIDENT_CASE_TYPE_SYS_ID + ',' + QUERY_CASE_TYPE_SYS_ID);
+caseRecord.orderBy('sys_updated_on');
+caseRecord.setLimit(500);
+caseRecord.query();
```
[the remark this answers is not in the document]
Jaiya94: Fixed

apps/mttr-application/openapi.yaml
```diff
@@ -0,0 +1,668 @@
...
+              total_cases:
+                type: integer
+                example: 120
+              excluded_cases:
+                type: integer
+                example: 6
+              p95_cutoff_hours:
+                type: number
+                format: float
+                example: 480.2
```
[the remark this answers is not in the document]
Jaiya94: Fixed

apps/mttr-application/openapi.yaml
```diff
@@ -0,0 +1,668 @@
...
+                format: float
+                example: 210.5
+              simple_avg_hours:
+                type: number
+                format: float
+                example: 230.1
+              min_sample_met:
+                type: boolean
+                example: true
+
```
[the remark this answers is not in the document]
Jaiya94: Fixed

apps/mttr-application/docs/design-document.md
```diff
@@ -0,0 +1,452 @@
...
+
+### 4.7 Patched vs Non-Patched Derivation
+
+**Decision:** Derive `is_patched` from whether the SN `fix_eta_shared` date field is populated.
+
+**Rationale:**
+- The source field is a date, not a boolean.
+- If the date is non-empty → the case had a fix ETA shared → considered "patched."
+- Conversion happens at extraction time in the SN Scheduled Job, not in Choreo.
+
```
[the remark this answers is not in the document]
Jaiya94: Fixed

apps/mttr-application/src/routes/health.js
```diff
@@ -0,0 +1,124 @@
...
+                    db_time: dbServerTime,
+                    total_cases: parseInt(totalCasesQuery.rows[0]?.estimate ?? 0, 10),
+                    last_ingestion_at: lastIngestionQuery.rows[0].last_ingestion_at || null,
+                    last_aggregation_at: lastCalculatedAt,
+                    cache_entries: cacheRowCount,
+                };
+            })(),
+            timeoutAfter(HEALTH_QUERY_TIMEOUT_MS, 'Health probe'),
+        ]);
+        return res.json(responseBody);
```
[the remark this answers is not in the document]
Jaiya94: Fixed

apps/mttr-application/.env.example
```diff
@@ -0,0 +1,65 @@
...
+DB_POOL_IDLE_MS=30000
+# How long to wait for a free client from the pool before erroring out.
+# Prevents requests from hanging indefinitely when the DB is unreachable.
+DB_POOL_CONNECT_MS=10000
+# Postgres-side per-query deadline (SET statement_timeout, ms). Any query
+# running longer than this is cancelled server-side and its pool client
+# returns to the pool. Must sit above every legitimate query duration —
+# aggregation ~10s, retention ~30s per full run; 2 min gives comfortable
+# headroom. Raise if the workload ever grows past that.
+DB_STATEMENT_TIMEOUT_MS=120000
```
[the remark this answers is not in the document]
Jaiya94: Fixed

Rashmika998 (approved):
LGTM

PR #992, opened 2026-07-01
[CSM Portal][FE] feat(csm-timecards): foundation types, constants and config keys (1/6)

Hesara2003:
## Purpose
Introduces the shared TypeScript types, API query-key constants, runtime config declarations and `config.js.example` documentation that the time-cards feature (ISSU-009) depends on. **This PR must be merged before PRs 2–6 in this series.**

## Goals
- Define `CsmTimeCard`, `CsmTimeSheet` and all supporting types in `types/timeCards.ts`
- Add `TIME_SHEETS_SEARCH`, `TIMECARD_APPROVAL_QUEUE`, `TIMECARD_REPORTS`, `TIMECARD_DELEGATION` to `ApiQueryKeys`
- Declare `CSM_PORTAL_TEAM_LEAD_GROUP` / `CSM_PORTAL_TIMECARD_ADMIN_GROUP` in `authConfig.ts`
- Document both runtime config keys in `config.js.example`

## Approach
Pure type and constant additions — no UI or API logic. All 5 subsequent PRs depend on these definitions.

## User stories
**ISSU-009 — Time Management** (Epic: Issue Management, Priority: Must have)
> As a CS Engineer, I want to log time spent on an issue so that effort is tracked for operational reporting and billing purposes.

Acceptance criteria covered across the full 6-PR series:
- Log entries with date, hours, activity type, and description ✓ (types defined here)
- Activity categories: analysis & debugging, reproduce, setting up, providing solution, answering ✓ (`ACTIVITY_KEYS`)
- Billable / unbillable flag per entry ✓ (`billable` on `CsmTimeCard`)
- Multiple entries per issue ✓
- Team Lead aggregated view ✓ (role config keys)
- Lifecycle notifications on approval/rejection ✓ (notification types)

## Release note
Adds internal TypeScript types and constants for the time-cards feature. No user-visible change in this PR.

## Documentation
N/A

## Training
N/A

## Certification
N/A

## Marketing
N/A

## Automation tests
- Unit tests: N/A — pure types and constants
- Integration tests: N/A

## Security checks
- Followed secure coding standards in http://wso2.com/technical-reports/wso2-secure-engineering-guidelines? yes
- Ran FindSecurityBugs plugin and verified report? N/A (frontend only)
- Confirmed that this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets? yes

## Samples
N/A

## Related PRs
**6-PR series for ISSU-009 — must be merged into `v2` in this order:**
[...]

PR #1028, opened 2026-07-03
[CSM Portal][FE] fix(csm-timecards): post-merge hardening (search truncation, self-approval, validation)

Hesara2003:
## Summary
Frontend fixes for the items tracked in [digiops-cs#2229](https://github.com/wso2-enterprise/digiops-cs/issues/2229), filed after reviewing the shipped ISSU-009 code (merged via #992–#998, #1017).

- **Search truncation (HIGH)**: `searchTimeCards` now paginates through every page up to a 1,000-card safety cap instead of silently stopping after 50 — "My time sheets" and the Approvals queue could miss real data with no indication anything was cut off. Shows a notice if the cap itself is ever hit.
- **Self-approval (HIGH)**: the approver picker now excludes the signed-in user — nothing stopped naming yourself as approver and then approving your own card.
- **Identity gate (MED)**: `useCurrentEngineer` no longer falls back to email — that fallback made the "wait for a real id" gate never actually wait, since the ID-token email resolves before `GET /users/me` does.
- **0h submit (MED)**: form validation now checks the *rounded* hour total (matching what's actually sent), not the raw one — a form could show a valid nonzero total where every individual bucket still rounds to 0.
- **Minor**: removed dead `from`/`to` filters (nothing in the UI set them); `playwright.config.ts`'s `reuseExistingServer` is now `!CI` instead of unconditional.
- **Bonus, found while verifying the above live**: the backend fails an *entire* search page (50 records) if even one record in it is malformed — confirmed live against 9 real corrupt records, reported separately with correlation IDs. Search now degrades to partial results instead of losing an otherwise-successful fetch to one bad record several pages in.

**Correction to item #5 in the issue** (work date ignored for weekly grouping): verified live and this isn't actually a bug — `createdOn` on read is set to whatever date was submitted on create (confirmed by backdating a test card to 2020-01-15 and reading it back), not a separate system-generated timestamp. Weekly grouping already uses the correct date. No fix needed here; the rest of the issue's backend items are still valid and tracked/handled separately with the team.

## Test plan
- [x] `pnpm run lint` / `tsc -b` / `pnpm run test` (124 unit tests) / `pnpm run build` all pass
- [x] Verified live against real staging data, including the partial-result degradation path

## Summary by CodeRabbit

* **New Features**
  * Added tab-specific pagination for time-card listings (**My time sheets**, **All**, **Approvals**) with backend-provided totals.
  * Added a “results are partial” notice when listings are truncated, and improved weekly sheet grouping/status.
  * Added optional hiding of the State filter in the FilterBar.
* **Bug Fixes**
  * Improved identity and scoping for “My” and “Approvals”, removed fallback behavior, and excluded the signed-in user from approver suggestions.
  * Enhanced empty-state messaging and tightened hours validation using backend-style rounding.
* **Tests**
  * Expanded unit tests for truncation, totals rounding, and sheet grouping; updated E2E tests/fixtures for approver search behavior.
* **Chores**
  * Updated Playwright dev-server reuse behavior for local vs CI runs; refreshed E2E README and comments.

### DIFF
--- a/apps/csm-portal/webapp/playwright.config.ts
+++ b/apps/csm-portal/webapp/playwright.config.ts
@@ -46,7 +46,10 @@ export default defineConfig({
     : {
         command: "pnpm run dev",
         url: BASE_URL,
-        reuseExistingServer: true,
+        // Locally, reuse an already-running dev server. In CI, always boot a
+        // fresh one — reusing a stray/stale process there could mask a real
+        // failure to start, or run against unexpected leftover state.
+        reuseExistingServer: !process.env.CI,
         timeout: 120_000,
       },
   projects: [
--- a/apps/csm-portal/webapp/src/api/backend/types.ts
+++ b/apps/csm-portal/webapp/src/api/backend/types.ts
@@ -1221,8 +1221,22 @@ export interface BeTimeCardView {
   case?: BeTimeCardCaseRef;
 }
 
+/**
+ * `entity-service`'s `openapi.yaml` also documents a `caseId` filter here
+ * (and it's genuinely implemented end-to-end, forwarded through to
+ * ServiceNow) — deliberately omitted: confirmed live to be non-functional,
+ * always returning `total: 0` even for a case with cards provably matching
+ * that exact id. See the note on `useCaseTimeCards` in `useTimeCards.ts`
+ * before re-adding it.
+ */
 export interface BeSearchTimeCardsFilters {
   projectIds?: string[];
+  /** Only time cards submitted by this user. */
+  userId?: string;
+  /** Only time cards this user is eligible to approve (SN `approver_list`). */
+  approverId?: string;
+  /** Only time cards actually approved by this user (SN `approved_by`). */
+  approvedById?: string;
   /** ISO 8601 date (YYYY-MM-DD). */
   startDate?: string;
   /** ISO 8601 date (YYYY-MM-DD). */
--- a/apps/csm-portal/webapp/src/constants/apiConstants.ts
+++ b/apps/csm-portal/webapp/src/constants/apiConstants.ts
@@ -62,6 +62,7 @@ export const ApiQueryKeys = {
   CASE_TIME_CARDS_SEARCH: "case-time-cards-search",
   TIME_SHEETS_SEARCH: "time-sheets-search",
   TIME_CARD_APPROVAL_QUEUE: "time-card-approval-queue",
+  TIME_CARD_ALL: "time-card-all",
   PROJECT_DEPLOYMENT_DETAILS: "project-deployment-details",
   UPDATE_LEVELS_SEARCH: "update-levels-search",
   CHANGE_REQUESTS: "change-requests",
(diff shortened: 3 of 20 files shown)


Hesara2003:
Re: the self-approval finding — confirmed still valid against current code. `entity-service`'s `CreateTimeCard` (`sn_time_card_service.go`) validates `approverIds` for non-empty/UUID shape but never checks the submitter isn't in that list, so a crafted request bypassing the dialog's client-side filtering could still self-approve.

This is backend-only and maps to item D ("reject self-approval server-side") already tracked in digiops-cs#2229, alongside the other backend follow-ups from that issue. Not something this FE-only PR can fix — leaving it as-is here; being handled separately with the team.

Hesara2003:
Also addressed the 2 findings from the same review that don't show as regular inline comments (outside-diff-range + nitpick, embedded in the review body):

- \`useCaseTimeCards\` now propagates \`truncated\` — the case detail Time tracking panel previously had no way to know its own 1,000-card cap was hit.
- Extracted the truncation notice into a shared \`TimeCardTruncatedNotice\` component (reused by the page and the case panel now) with \`role="status"\` for the nitpick's accessibility suggestion.

That's everything from this review now — the one remaining item (self-approval enforcement) is backend-only, tracked separately as noted above.

Hesara2003:
Fixed the new one too — `CaseTimeCardsPanel`'s truncation notice could render alongside "Could not load time cards." since React Query keeps stale `data` around across a failed background refetch. Gated on `!isError` now.

rksk:
Heads-up before this merges — the core premise here (client-side pagination because "there's no `caseId`/`engineerId` filter on the backend") is now out of date. The search endpoint supports these **server-side**, verified live against the ServiceNow `time_cards/search` path (the csm-portal-backend passes the filter body through unmodified, and the entity-service OpenAPI already documents them):

- `caseId` — cards on a case (SN `task` equality)
- `userId` — submitter / reporter (SN `user`)
- `approverId` — cards a user is *eligible* to approve (SN `approver_list CONTAINS`)
- `approvedById` — cards a user *actually* approved (SN `approved_by`)
- `projectIds`, `startDate`/`endDate` — as before

Implications for this PR:

1. **Truncation (F1):** rather than paginating through every user's cards up to a 1,000-card cap and filtering client-side, scope at the source — `userId = me` for *My time sheets*, `caseId` for the case panel. The page comes back already narrowed, so truncation stops being a practical problem. Bounded pagination can stay as a residual safety net for a >50 *scoped* result, but it needn't be the primary mechanism.

2. **Approvals 403-on-click (not addressed here):** scope the queue by `approverId = me`. Today the queue lists every submitted card and the decision `PATCH` 403s when you click approve on one you're not the assigned approver for. Filtering `approverId = me` server-side means the queue only ever shows cards you can actually action.

3. **`from`/`to` (removed here as dead):** worth keeping — they map to server-side `startDate`/`endDate` and are what a date-range filter would use (one of the requested filter dimensions: reporter/approver/project/case/time-range).

4. **The batch-500 "bonus":** root cause is a **dangling `task` reference** (a time card pointing at a deleted case), not the page size. It's being addressed server-side (the search row mapper now guards the deleted-task ref before dot-walking it), so the FE doesn't need pagination to work around that specific 500.

The self-approval (F2), identity-gate (F3), and rounded-total (F4) fixes here are all good and orthogonal — worth keeping as-is. It's mainly the F1 strategy and the queue scoping that change once the server-side filters are used. Happy to pair on reworking `searchTimeCards` + the hooks to the server-side approach.

rksk:
@coderabbitai resume

rksk (approved):

PR #1039, opened 2026-07-03
[CSM Portal] gate case reassign while work is in-progress + ongoing

rksk:
## What

Disable the **Assign / reassign engineer** overflow action on the case detail page when the case is **Work in progress** with an **ongoing** work-state, and add a tooltip explaining why.

## Why

The backend rejects an assignee change on a WIP + ongoing case: it silently reverts the change and still returns `200` success. The portal therefore showed a "Case reassigned" success toast while the assignee never actually changed — a confusing false success.

Rather than fire a request we know will no-op, we gate the action in the UI. The tooltip tells the user to pause the work first (via the existing Pause action, available to the current assignee) or have a lead handle the reassignment.

## Notes

- This is a **case-state** gate, mirroring the existing state-based enablement of the pause/resume action. It is **not** a data-source gate.
- Interim guard: a backend change to surface the rejection as a real error (instead of a false success) is tracked separately. Once that lands, this UI gate can be revisited.

## Testing

- `pnpm test` — added 3 cases to `CaseActionBar.test.tsx` (disabled when WIP+ongoing and no dispatch on click; enabled when paused; enabled for non-WIP states). Full file passes (11/11).
- `pnpm build` — passes.
- `pnpm lint` — the changed file is clean (the 2 remaining repo lint errors are pre-existing on `v2` in unrelated files).

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented reassignment of an engineer when a case is actively in progress, avoiding actions that would have no effect.
  * Added clearer hover guidance for disabled actions in the overflow menu.
  * Kept reassignment available in paused or non-in-progress states, with the action behaving as expected.

### DIFF
--- a/apps/csm-portal/webapp/src/features/csm-cases/components/CaseActionBar.test.tsx
+++ b/apps/csm-portal/webapp/src/features/csm-cases/components/CaseActionBar.test.tsx
@@ -219,3 +219,64 @@ describe("CaseActionBar — nextStates-driven buttons", () => {
     expect(screen.getByRole("button", { name: /more/i })).toBeInTheDocument();
   });
 });
+
+describe("CaseActionBar — reassign gating for WIP-Ongoing", () => {
+  /** Open the "More" overflow and return the reassign engineer menu item. */
+  function openReassignItem(): HTMLElement {
+    fireEvent.click(screen.getByRole("button", { name: /more/i }));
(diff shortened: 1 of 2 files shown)


rksk:
@coderabbitai resume

Rashmika998 (approved):

PR #1042, opened 2026-07-05
[CSM][Web] UX polish: case list, chips, internal notes, state dropdown, service request routing

Rashmika998:
## Summary

- **Case list**: split State / Work state into two separate columns; outlined `StateChip`; native hover tooltip on truncated text; removed em-dash fallbacks
- **SeverityChip**: format changed to `S2 (Critical)` instead of `S2 — Critical`
- **State dropdown**: consolidated per-state lifecycle buttons in `CaseActionBar` into a single "Change state ▼" `Menu`; existing confirm-dialog path unchanged
- **Overview header**: entire row in `CaseMetaBand` is now a clickable `component="button"`; Tier chip uses `color="primary" variant="outlined"` so it follows the active theme
- **Internal notes**: theme-neutral styling across `CsmCaseCommentBubble` and `CsmCaseCommentInput` — `primary.main` left-border accent + `action.hover` background instead of hardcoded `info`/`warning` palette; "Internal note" chip changed to `SemanticChip role="default" variant="outlined"`; removed WSO2 role chip from comment bubbles
- **`SemanticChip` / `StateChip`**: added `variant` prop (`filled | outlined`)
- **Loading states**: hide case count subtitle while loading/fetching; replace `CircularProgress` in `ProductVulnerabilitiesTab` with skeleton rows; `variant="rounded"` skeletons in case detail
- **Service request routing**: `OperationsPage` passes `detailBasePath="/operations/service-requests"`; new `App.tsx` route `operations/service-requests/:caseId` renders `CsmCaseDetailPage`; page detects route and sets "Back to service requests" link + correct canonical URL
- **Minor copy**: simplified `CallRequestsWidget` empty state; friendlier `AppErrorBoundary` error message; removed tooltip from `ThemeSelect`

## Test plan

- [x] Case list shows State and Work state in separate columns; outlined chips; hover tooltip on long text
- [x] Severity chip renders `S2 (Critical)` format
- [x] "Change state ▼" dropdown opens and triggers state transitions (including confirm dialog for gated ones)
- [x] Overview header row click toggles collapse; Tier chip colour matches active theme
- [x] Internal note comments show left-border accent and outlined chip across both light and non-default (e.g. Acrylic Orange) themes
- [x] WSO2 role chip no longer shown on engineer comments
- [x] Case count hidden while list is loading / fetching
- [x] Vulnerabilities table shows skeleton rows on load
- [x] Clicking a service request in the Operations tab opens it under `/operations/service-requests/:id` (not `/cases/:id`); "Back to service requests" navigates to `/operations?tab=service_requests`

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated service-request detail route with matching back navigation from the operations view.
  * Introduced a separate “Work state” column in the cases list for clearer status visibility.
  * Updated action controls to use a single “Change state” menu for lifecycle transitions.

* **Bug Fixes**
  * Improved loading and empty-state behavior across cases, comments, and vulnerabilities views.
  * Refined several labels, fallbacks, and chip styles for cleaner, more consistent display.
  * Updated the page error message to be clearer and more user-friendly.

### DIFF
--- a/apps/csm-portal/webapp/src/features/csm-cases/api/useGetCsmCases.ts
+++ b/apps/csm-portal/webapp/src/features/csm-cases/api/useGetCsmCases.ts
@@ -202,7 +202,7 @@ export function useGetCsmCases(
       // customer column (cases embed the project, but not its account). The
       // lookups go through `fetchQuery` with their own stable keys, so they
       // hit the network only when their cache is stale — not on every filter
-      // change. Lookup failures degrade to "—" names, not a failed list.
+      // change. Lookup failures degrade to blank names, not a failed list.
       const [casesResponse, projects, accounts] = await Promise.all([
         api.post<BeCaseSearchPayload, BeCaseSearchResponse>("/cases/search", {
           pagination: { offset, limit: pageSize },
--- a/apps/csm-portal/webapp/src/features/csm-cases/api/useGetCsmCases.ts
+++ b/apps/csm-portal/webapp/src/features/csm-cases/api/useGetCsmCases.ts
@@ -277,13 +277,13 @@ export function useGetCsmCases(
           caseNumber: c.number,
           wso2CaseId: c.internalId,
           subject: c.subject ?? "(no subject)",
-          customer: accountName.get(accountId) ?? "—",
+          customer: accountName.get(accountId) ?? "",
           accountId,
           projectId,
-          projectName: c.project?.name ?? "—",
+          projectName: c.project?.name ?? "",
           // Search embeds deployedProduct as { id, name } (name includes the
           // version); the GET view uses a displayName-shaped ref instead.
-          product: c.deployedProduct?.name ?? "—",
+          product: c.deployedProduct?.name ?? "",
           severity: severityFromPriority(c.severity),
           state: uiStateFromBe(c.state),
           caseType: c.type,
--- a/apps/csm-portal/webapp/src/features/csm-cases/api/useGetCsmCases.ts
+++ b/apps/csm-portal/webapp/src/features/csm-cases/api/useGetCsmCases.ts
@@ -277,13 +277,13 @@ export function useGetCsmCases(
           caseNumber: c.number,
           wso2CaseId: c.internalId,
           subject: c.subject ?? "(no subject)",
-          customer: accountName.get(accountId) ?? "",
+          customer: accountName.get(accountId) ?? "-",
           accountId,
           projectId,
-          projectName: c.project?.name ?? "",
+          projectName: c.project?.name ?? "-",
           // Search embeds deployedProduct as { id, name } (name includes the
           // version); the GET view uses a displayName-shaped ref instead.
-          product: c.deployedProduct?.name ?? "",
+          product: c.deployedProduct?.name ?? "-",
           severity: severityFromPriority(c.severity),
           state: uiStateFromBe(c.state),
           caseType: c.type,
--- a/apps/csm-portal/webapp/src/features/csm-cases/components/CaseActionBar.tsx
+++ b/apps/csm-portal/webapp/src/features/csm-cases/components/CaseActionBar.tsx
@@ -275,6 +275,7 @@ export default function CaseActionBar({
   onAction,
 }: CaseActionBarProps): JSX.Element {
   const [menuAnchor, setMenuAnchor] = useState<HTMLElement | null>(null);
+  const [stateMenuAnchor, setStateMenuAnchor] = useState<HTMLElement | null>(null);
   const [pendingConfirm, setPendingConfirm] = useState<PrimaryButton | null>(
     null,
   );
@@ -308,35 +309,39 @@ export default function CaseActionBar({
         justifyContent: { xs: "flex-start", md: "flex-end" },
       }}
     >
-      {primary.map((p, idx) => {
-        const button = (
+      {primary.length > 0 && (
+        <>
           <Button
             size="small"
-            variant={idx === 0 ? "contained" : "outlined"}
-            color={p.color}
-            startIcon={p.icon}
-            onClick={() => runPrimary(p)}
+            variant="outlined"
+            endIcon={<ChevronDown size={16} />}
+            onClick={(e) => setStateMenuAnchor(e.currentTarget)}
           >
-            {p.label}
+            Change state
           </Button>
-        );
-        // `targetState` is unique per button (each moves to a distinct state),
-        // so it is the correct React key — the lifecycle action is not (e.g.
-        // `resume_work` can map from more than one source state).
-        return p.tooltip ? (
-          <Tooltip key={p.targetState} title={p.tooltip}>
-            {button}
-          </Tooltip>
-        ) : (
-          <Box
-            key={p.targetState}
-            component="span"
-            sx={{ display: "inline-flex" }}
+          <Menu
+            anchorEl={stateMenuAnchor}
+            open={!!stateMenuAnchor}
+            onClose={() => setStateMenuAnchor(null)}
           >
-            {button}
-          </Box>
-        );
-      })}
+            {primary.map((p) => (
+              <MenuItem
+                key={p.targetState}
+                onClick={() => {
+                  setStateMenuAnchor(null);
+                  runPrimary(p);
+                }}
+                sx={{ gap: 1.25, minHeight: 36 }}
+              >
+                <Box sx={{ color: `${p.color}.main`, display: "flex" }}>
+                  {p.icon}
+                </Box>
+                {p.label}
+              </MenuItem>
+            ))}
+          </Menu>
+        </>
+      )}
 
       <Button
         size="small"
--- a/apps/csm-portal/webapp/src/features/csm-cases/components/CaseMetaBand.tsx
+++ b/apps/csm-portal/webapp/src/features/csm-cases/components/CaseMetaBand.tsx
@@ -14,11 +14,11 @@
 // specific language governing permissions and limitations
 // under the License.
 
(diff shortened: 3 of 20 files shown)


apps/csm-portal/webapp/src/features/csm-cases/components/CsmIssuesView.tsx
```diff
@@ -123,7 +123,7 @@ export default function CsmIssuesView({
     [filters, debouncedSearch, lockedFilters],
   );
 
-  const { data, isLoading, isError, error } = useGetCsmCases(
+  const { data, isLoading, isFetching, isError, error } = useGetCsmCases(
```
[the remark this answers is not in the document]
Rashmika998: Fixed in 6bd90ba — reverted to -only for the subtitle and removed the  destructure. The subtitle now stays visible (showing the previous page count) during background refetches triggered by filter/search/page changes, keeping it in sync with the table rows that also stay rendered.

Rashmika998:
Addressed all CodeRabbit review feedback in 6bd90ba:

- **CsmIssuesView subtitle flicker (actionable)**: reverted to `isLoading`-only; removed `isFetching` from destructure. Subtitle stays visible during background refetches so it stays in sync with the table rows.
- **CsmCaseDetailPage empty heading (nitpick)**: restored `—` em-dash fallback when both `wso2CaseId` and `caseNumber` are absent.
- **useGetCsmCases stale comment (nitpick)**: updated to say "blank names" instead of "'—' names".

apps/csm-portal/webapp/src/features/csm-cases/components/CaseMetaBand.tsx
```diff
@@ -14,11 +14,11 @@
...
 
-import { Box, Card, IconButton, Tooltip, Typography } from "@wso2/oxygen-ui";
+import { Box, Card, Chip, Typography } from "@wso2/oxygen-ui";
 import { ChevronDown, ChevronUp } from "@wso2/oxygen-ui-icons-react";
 import { useState, type JSX, type ReactNode } from "react";
 import { Link as RouterLink } from "react-router";
-import { tierColor, tierLabel } from "@features/csm-cases/utils/caseTier";
+import { tierLabel } from "@features/csm-cases/utils/caseTier";
 import type { CsmCaseDetail } from "@features/csm-cases/types/csmCases";
 import SemanticChip from "@components/SemanticChip";
```
rksk, line 23: **Unused import — will fail `pnpm lint`.** The only use of `SemanticChip` in this file was the Tier chip, which this PR replaced with a plain `Chip` (line 240). With no remaining references, `@typescript-eslint`/`unused-imports` will error and break CI. Remove this import.

apps/csm-portal/webapp/src/features/csm-cases/components/CaseMetaBand.tsx
```diff
@@ -235,7 +237,7 @@ export default function CaseMetaBand({
         >
           <Cell label="Tier">
             <Box sx={{ minWidth: 0 }}>
-              <SemanticChip role={tierColor(tier)} label={tierLabel(tier)} />
+              <Chip size="small" variant="outlined" color="primary" label={tierLabel(tier)} />
```
rksk: Two side effects of `SemanticChip role={tierColor(tier)}` → `Chip color="primary"`:

1. **All tiers now render identically.** `tierColor(tier)` mapped each tier to a distinct colour; hardcoding `primary` erases that, so tiers are no longer distinguishable by colour.
2. **Brand orange on a passive metadata chip.** `color="primary"` is the WSO2 brand accent, which we reserve for primary actions — an outlined-orange metadata chip competes with real action buttons.

If the goal is just to follow the active theme (not hardcode a palette colour), `color="default"` gives an outlined neutral chip without spending brand orange. If per-tier differentiation still matters, keep `SemanticChip role={tierColor(tier)}`.

apps/csm-portal/webapp/src/features/csm-cases/components/CaseMetaBand.tsx
```diff
@@ -180,12 +180,22 @@ export default function CaseMetaBand({
       }}
     >
       <Box
+        component="button"
```
rksk, line 183: `Box component="button"` renders a native `<button>`, which defaults to `type="submit"`. Harmless today (not inside a form), but it's a latent footgun if this ever ends up in one. Add `type="button"`. (The pre-existing `LinkButton` at line 121 has the same gap — not introduced here, but easy to fix while nearby.)

apps/csm-portal/webapp/src/features/csm-cases/components/CaseActionBar.tsx
```diff
@@ -308,35 +309,39 @@ export default function CaseActionBar({
...
-            variant={idx === 0 ? "contained" : "outlined"}
-            color={p.color}
-            startIcon={p.icon}
-            onClick={() => runPrimary(p)}
+            variant="outlined"
+            endIcon={<ChevronDown size={16} />}
+            onClick={(e) => setStateMenuAnchor(e.currentTarget)}
           >
-            {p.label}
+            Change state
```
rksk, line 320: Consolidating the lifecycle buttons into a single "Change state" menu drops the per-transition `p.tooltip` that the old buttons carried (the `Tooltip` import now serves only the secondary menu below). It also removes the "recommended next action" emphasis (first button was `contained`+coloured); the menu keeps the icon colour via `p.color`, but every transition now reads as equal-weight. If those tooltips explained non-obvious transitions, that context is gone. Intentional? If so, fine — flagging so it's a conscious drop, not an accident.

apps/csm-portal/webapp/src/features/csm-cases/api/useGetCsmCases.ts
```diff
@@ -277,13 +277,13 @@ export function useGetCsmCases(
           caseNumber: c.number,
           wso2CaseId: c.internalId,
           subject: c.subject ?? "(no subject)",
-          customer: accountName.get(accountId) ?? "—",
+          customer: accountName.get(accountId) ?? "",
```
rksk: Switching the fallback from `"—"` to `""` (here and for `projectName`/`product` on lines 283/286) means a missing value now renders a fully blank cell instead of a "no data" marker, which can read as a rendering glitch. Combined with `CasesList` dropping the Case-ID `"—"` fallback, a row missing an id shows an empty first column. If blank is deliberate, consider keeping a lightweight placeholder for the Case-ID column at least — an empty col-1 is more jarring than a mid-row blank.

rksk (commented):
UX polish looks good overall. A few points inline: one blocking lint issue (unused import) and a couple of deliberate-but-notable UX changes worth a second look.

apps/csm-portal/webapp/src/features/csm-cases/components/CaseMetaBand.tsx
Rashmika998: Fixed in 0a14d6f — removed the SemanticChip import (Chip replaced it for the tier cell, but we've now switched back to SemanticChip, so the Chip import is the one being dropped instead).

apps/csm-portal/webapp/src/features/csm-cases/components/CaseMetaBand.tsx
Rashmika998: Fixed in 0a14d6f — reverted to SemanticChip role={tierColor(tier)} variant="outlined". This preserves per-tier colour differentiation and uses the MUI semantic palette (success/warning/error/info) rather than hardcoding primary, so it's theme-neutral and doesn't compete with action buttons.

apps/csm-portal/webapp/src/features/csm-cases/components/CaseMetaBand.tsx
Rashmika998: Fixed in 0a14d6f — added type="button" to the collapse toggle Box. The LinkButton helper at line 121 already had type="button"; updated it there too.

apps/csm-portal/webapp/src/features/csm-cases/components/CaseActionBar.tsx
Rashmika998: Intentional drop. The per-transition tooltips and the contained/coloured "recommended" button were removed as a deliberate design simplification — the previous N buttons were visually noisy and the priority emphasis wasn't adding much value in practice. The icon colour on each menu item (via p.color) is enough to signal the nature of the action. If engineers find the loss of tooltips confusing for specific transitions we can add MenuItem secondaryAction text later.

apps/csm-portal/webapp/src/features/csm-cases/api/useGetCsmCases.ts
Rashmika998: Updated in ef72e9e — changed the fallback to `"-"` for customer, projectName, and product. Keeps blank cells from reading as a rendering glitch without bringing back the em-dash.

rksk (approved):
Approving. All review findings addressed and verified against the current head (`cafae619`):

- Unused `SemanticChip` import removed; tier chip restored to `SemanticChip role={tierColor(tier)}` (per-tier colour, no hardcoded brand accent); `type="button"` added to the collapse toggle.
- Missing customer/project/product now render a `-` placeholder consistently across both the cases list and the dashboard hook.
- Loading progress bars moved off `warning` to `primary`.

Ran eslint on the changed files locally: 0 errors (the one `react-hooks/exhaustive-deps` warning on `detailPath` pre-dates this PR on v2). The consolidation of lifecycle buttons into the "Change state" menu drops the per-transition tooltips — confirmed intentional by the author; fine to revisit later if engineers find specific transitions unclear.

PR #1075, opened 2026-07-07
[CSM][Web] Fix loading overlay overlap and fill Change state button

Rashmika998:
## Summary
- **Loading overlap fix**: Gate the top-bar `LinearProgress` behind `hasInitialized` so it only appears after auth has settled. Previously both the top-bar loader and the centered auth loader rendered simultaneously during the initial page load, causing two overlapping orange bars.
- **Change state button**: Changed from `variant="outlined"` to `variant="contained" color="primary"` so it reads as the primary action on the case detail page.

## Test plan
- [ ] On initial load, only the centered progress bar + message is shown (no top bar)
- [ ] After auth settles, the top-bar loader appears correctly during route transitions and API calls
- [ ] "Change state" button renders as a filled primary-colored button on the case detail page
- [ ] "More" button remains outlined for visual hierarchy contrast

## Summary by CodeRabbit

* **UI Improvements**
  * Updated the “Change state” action button to use a more prominent primary style.
  * Refreshed activity feed change entries to display updates in a clearer `old → new` format.
  * Improved the appearance of activity history when values are changed or cleared.

### DIFF
--- a/apps/csm-portal/webapp/src/features/csm-cases/components/CaseActionBar.tsx
+++ b/apps/csm-portal/webapp/src/features/csm-cases/components/CaseActionBar.tsx
@@ -313,7 +313,8 @@ export default function CaseActionBar({
         <>
           <Button
             size="small"
-            variant="outlined"
+            variant="contained"
+            color="primary"
             endIcon={<ChevronDown size={16} />}
             onClick={(e) => setStateMenuAnchor(e.currentTarget)}
           >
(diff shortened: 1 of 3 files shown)


Rashmika998:
Added two more fixes in the latest commit:

**1. Loading overlay — full-screen cover**
Changed the pre-auth `Box` from `height: 100dvh` to `position: fixed; inset: 0; z-index: 9999; bgcolor: background.default`. This makes it a true viewport overlay, so nothing underneath (body defaults, Asgardeo ThemeProvider CSS variables on `<html>`, etc.) bleeds through regardless of theme.

**2. State-change display — arrow format**
Replaced the `<new> was ~~old~~` strikethrough pattern in the activities feed with `<old> → <new>`, e.g. `State: Open → Work In Progress`. The old value renders in muted (`text.secondary`) colour; cleared fields show `<old> → cleared`.

---

cloby99 (approved):

PR #1094, opened 2026-07-08
[CSM Portal] collect Post Resolution Activity when closing or proposing a solution

rksk:
`resolutionCode`, `cause`, and `closeNotes` were added to the `PATCH /cases/{id}` contract at the API layer only — the entity-service and this repo's own `openapi.yaml` accept them, and the BFF handler passes them through as-is. But no UI was ever built to collect these from the user: closing a case or proposing a solution just fired a plain state-transition PATCH with none of the Post Resolution Activity (PRA) fields the product requirement (case resolution management) calls for.

Collect resolution code, root cause, and optional close notes from the CS engineer before a case is closed or a solution is proposed, and send them in the same `PATCH /cases/{id}` call as the state transition.

- New `ResolutionDialog` component: a resolution-code select and a cause select (both required, populated from the full backend enums), plus a free-text close-notes field (optional). Opens for both the "Close" and "Propose solution" actions.
- `CsmCaseDetailPage`'s action handler now intercepts `close`/`propose_solution` before the generic state-PATCH branch and opens this dialog instead of PATCHing immediately; submitting sends `{ state, resolutionCode, cause, closeNotes }` in one call.
- The dialog's own confirmation copy (customer-notification text) replaces `CaseActionBar`'s old plain yes/no confirm for these two targets, so that now-unused confirm mechanism (`ActionConfirm` type, `pendingConfirm` state, its `Dialog`) is removed from `CaseActionBar` rather than left dead.
- New `caseResolution.ts` util holds the two enum value lists plus a `humanizeResolutionEnum` helper (mirrors the existing `humanizeState` convention) so dropdown labels stay readable without a hand-maintained label map.
- Extended `BeCaseUpdatePayload`'s `state` variant with the three optional fields; no other request shapes change.

No API contract change — this only starts sending fields the backend has accepted since PR #1077.

As a CS engineer, when I close a case or propose a solution, I record the resolution code, root cause, and any closing notes as part of that action, instead of the case closing with no resolution record at all.

Added: closing a case or proposing a solution now collects the Post Resolution Activity (resolution code, cause, close notes) before submitting.

N/A — internal case-detail UI behavior, not separately documented.

## Automation tests
 - Unit tests
   > Code coverage information

New `ResolutionDialog.test.tsx` covers: submit disabled until both required fields are chosen, correct payload shape on submit, and cancel behavior. Updated `CaseActionBar.test.tsx` for the removed confirm-dialog behavior on Close (confirmation now happens via the resolution dialog upstream, not inside `CaseActionBar`). `pnpm build`, `pnpm test`, and `pnpm lint` all pass (two pre-existing, unrelated test failures on `origin/main` persist unchanged).
 - Integration tests

N/A — no integration test suite for this component.

## Security checks
 - Ran FindSecurityBugs plugin and verified report? N/A (TypeScript codebase; `eslint` runs clean)

## Samples
N/A

## Related PRs
None

## Migrations (if applicable)

## Test environment
Local: `pnpm build` + `pnpm test` + `pnpm lint`, on macOS.

## Learning
N/A

## Summary by CodeRabbit

* **New Features**
  * Added a new case resolution dialog for closing cases or proposing solutions.
  * Users can now select a resolution code and root cause, and add optional closing notes.
  * Case actions now open the dialog flow and save the selected details with the case update.
  * Improved case action behavior so close/propose actions respond immediately and use the shared resolution flow.

### DIFF
--- a/apps/csm-portal/webapp/src/api/backend/types.ts
+++ b/apps/csm-portal/webapp/src/api/backend/types.ts
@@ -87,6 +87,50 @@ export type BeCaseState =
  */
 export type BeCaseWorkState = "ongoing" | "paused";
 
+/**
+ * Resolution code for a closed or solution-proposed case (the Post
+ * Resolution Activity — see UseCases.md ISSU-026). Only accepted by
+ * `PATCH /cases/{id}` alongside `state: "closed"` or `"solution_proposed"`.
+ */
+export type BeCaseResolutionCode =
+  | "SOLVED_FIXED_BY_SUPPORT_GUIDANCE_PROVIDED"
+  | "SOLVED_FIXED_BY_CLOSING_RELATED_INCIDENT"
+  | "SOLVED_FIXED_BY_CLOSING_RELATED_RD_TICKET"
+  | "SOLVED_WORKAROUND_PROVIDED"
+  | "SOLVED_BY_CUSTOMER"
+  | "CONSIDERED_FOR_ROADMAP"
+  | "INCONCLUSIVE_OUT_OF_SCOPE"
+  | "INCONCLUSIVE_CANNOT_REPRODUCE"
+  | "INCONCLUSIVE_NO_WORKAROUND"
+  | "DUPLICATE_ISSUE"
+  | "VOIDED_CANCELED"
+  | "ON_HOLD"
+  | "CONSIDERED_FOR_ROADMAP_ALT"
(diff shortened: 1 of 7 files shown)


rksk:
@coderabbitai resume

Rashmika998 (approved):

PR #1095, opened 2026-07-08
[CSM Portal] fix: break the infinite refresh-grant loop on expired sessions

rksk:
## Purpose
When a user's session sits idle long enough that the refresh token itself expires (not just the access token), `useAuthApiClient` had no handling for the resulting auth-provider "unauthenticated" error. Every backend call kept failing with that error while the auth provider's periodic background token refresh kept retrying the dead refresh token, producing an infinite loop of failing refresh-grant requests instead of ending the session cleanly.

## Goals
- Detect the specific "unauthenticated" error code distinctly from other failures (network errors, real backend 5xx, SDK-not-ready races) so only genuine session-death triggers recovery.
- Give a concurrent or in-flight background refresh a chance to have already fixed the token before giving up.
- End dead sessions with a single, deliberate sign-in redirect instead of a silent infinite retry loop.

## Approach
`useAuthApiClient` now wraps its existing token-fetch-and-request logic (`attemptFetch`) with a retry-then-redirect ladder:
- On the "unauthenticated" error code, retry the request once — a concurrent caller or the provider's periodic background refresh may have already re-minted the token by the time the retry runs.
- If the retry still returns the unauthenticated code, the session (refresh token) is gone, so redirect to a full sign-in.
- If the retry fails for any other reason (e.g. a transient network blip), that error is rethrown untouched so existing error handling still works.
- The sign-in redirect is single-flighted at module scope so concurrent auth failures trigger exactly one redirect, and it returns a never-resolving promise so callers don't render an error page mid-navigation.

This mirrors the same fix already shipped for the customer portal's equivalent hook, adapted to this app's dual access-token/ID-token fetch and its origin-trust and correlation-ID logging wrapper, which are unchanged.

## User stories
As a CS engineer who leaves the portal open past session expiry, returning to the tab should either silently recover or cleanly prompt me to sign in again, not spin forever on failing background requests.

## Release note
Fixed an issue where an expired session could cause the portal to loop indefinitely on failing token-refresh requests instead of prompting the user to sign in again.

## Documentation
N/A — internal auth-recovery behavior change, no user-facing or API documentation impact.

## Automation tests
- Unit tests
  - No new unit tests added in this PR (existing hook had no test file). Verified via `tsc -b` (clean) and manual code-path review against the customer portal's equivalent, already-tested implementation.
- Integration tests
  - N/A

## Security checks
- Ran FindSecurityBugs plugin and verified report? N/A (TypeScript/React project; `eslint` run clean on the changed file)

## Samples
N/A

## Related PRs
None

## Migrations (if applicable)
N/A

## Test environment
Verified with `tsc -b` and `pnpm lint` on the changed file, and the existing `pnpm test` suite (no regressions versus main).

## Learning
Ported the recovery pattern from a prior fix to the customer portal's equivalent auth hook, which classifies the auth provider's expired-token error code and applies a retry-then-redirect ladder instead of an unconditional single retry.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sign-in recovery when authentication tokens are missing or expired.
  * Added a safer fallback that retries requests once before sending users to sign-in.
  * Prevented multiple simultaneous sign-in redirects when several requests fail at the same time.
  * Left non-authentication errors unchanged so normal error handling still applies.

### DIFF
--- a/apps/csm-portal/webapp/src/hooks/useAuthApiClient.ts
+++ b/apps/csm-portal/webapp/src/hooks/useAuthApiClient.ts
@@ -17,9 +17,35 @@
 import { useCallback } from "react";
 import { useAsgardeo } from "@asgardeo/react";
 import { apiConfig } from "@config/apiConfig";
-import { AUTH_NOT_READY_ERROR_MESSAGE } from "@constants/apiConstants";
+import {
+  ASGARDEO_UNAUTHENTICATED_CODE,
+  AUTH_NOT_READY_ERROR_MESSAGE,
+} from "@constants/apiConstants";
 import { useLogger } from "@hooks/useLogger";
 import { CORRELATION_ID_HEADER, newCorrelationId } from "@utils/correlationId";
+
+// Shared across every caller's hook instance. Each useAuthApiClient() call
+// creates its own authFetch closure, so this lives at module scope to ensure
+// only ONE full sign-in redirect is triggered even when many concurrent calls
+// fail authentication at once.
+let signInInFlight = false;
+
+// Only the Asgardeo "unauthenticated" code means the token was expired/missing
+// when the call ran (e.g. the refresh token itself has expired, so the SDK's
+// periodic background refresh can no longer mint a new access token). Anything
+// else (network failures, real backend 5xx) must propagate untouched so
+// existing error handling and error pages still work. Without this
+// classification, a dead refresh token sends the SDK's periodic background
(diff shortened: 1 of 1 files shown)


rksk:
@coderabbitai resume

rksk:
@coderabbitai resume

rksk:
@coderabbitai resume

Rashmika998 (approved):

PR #1116, opened 2026-07-10
[CSM Portal Microapp]-create case flow to CSM Portal /Support page

2003dinijay:
## Purpose
Case creation currently only exists in the CSM Portal webapp. CSMs working from mobile had no way to open a new case without switching to a desktop browser.

## Goals
- Let a CSM create a standard "case" from the microapp, matching what the webapp's `CsmCaseCreatePage` already supports for that case type.
- Give the case list a dedicated "View All" surface (search, filters, state tabs) instead of only the compact recent-cases preview on the Support page.

## Approach
- Ported the webapp's case-creation form fields and validation (cascading project → deployment → deployed-product selects, severity, issue type, subject/description) rather than reinventing the flow, so behavior stays consistent with the desktop experience.
- Attachments upload *after* the case is created, in parallel; a partial attachment failure still leaves the case created (same tradeoff the webapp makes) rather than rolling back the case or blocking on every attachment succeeding.
- Only the standard "case" type is in scope — `service_request` / `security_report_analysis` creation aren't ported, since those flows weren't part of this ask.
- Reused `AllCasesPage`'s infinite-scroll/search/filter pattern for the new "View All" route instead of introducing a second pagination approach.

## Summary
- Ports the webapp's `CsmCaseCreatePage` (the standard "case" type only — service_request/security_report_analysis creation aren't in scope for this mobile app): cascading project → deployment → deployed-product selects, severity, issue type, subject/description, and optional attachments uploaded after the case is created (a partial attachment failure still lands the case, same as the webapp).
- Reachable from a "+" button on the Support page header.

## User stories
- As a CSM on mobile, I can tap "+" on the Support page to open a new case without needing a desktop browser.
- As a CSM, I can search and filter the full case list (not just the recent preview) from a dedicated "View All" page.

## Release note
Added case creation and a full searchable/filterable case list to the CSM Portal microapp.

## Documentation
No user-facing docs outside this PR; the flow mirrors the existing webapp case-creation form CSMs already know.

## Tests
- [x] `tsc --noEmit` clean for the files in this PR
- [x] `eslint` clean
- [x] Manually verified in a browser: "+" button opens New Case, form renders, project/deployment/product cascading selects work, attachments UI works

## Security checks
- Attachment uploads are size-validated client-side before upload (see `utils/attachments.ts`); no new auth/permission surface is introduced — case creation uses the same authenticated API client as the rest of the microapp.

## Test plan
- [x] `tsc --noEmit` clean for the files in this PR
- [x] `eslint` clean
- [x] Manually verified in a browser: "+" button opens New Case, form renders, project/deployment/product cascading selects work, attachments UI works

## Summary by CodeRabbit

* **New Features**
  * Added a full “All cases” experience with URL-driven search, filtering, state tabs, and infinite scrolling.
  * Added a new case creation flow including project/deployment/product selection, severity and issue type, and validated attachment uploads.
* **UI Updates**
  * Updated the support page to show a compact recent-cases preview per tab with “View All” navigation.
  * Simplified case filtering by removing the general state chip controls.
  * Simplified the profile header by removing role badges.
* **Other**
  * Adjusted the active-filter count to avoid double-counting case states.


cloby99 (approved):

2003dinijay:
@coderabbitai resume

Rashmika998:
@coderabbitai pause

PR #1138, opened 2026-07-13
[CSM][BE] Return upstream error message for registry token creation failures

shayanmalinda:
## Problem

When registry token creation fails, customers see the generic message **"Invalid request parameters for creating registry token."** regardless of the actual cause. The registry-subscription-service returns meaningful 400 reasons (e.g. the 5-token limit message, or missing subscriptions), but the portal backend swallows them.

## Changes

- `service.bal`: the token-creation endpoint now returns the upstream error message in the 400 response body instead of a hardcoded generic string. Logging is unchanged.
- `utils.bal`: adds a new `extractClientErrorMessage` helper that extracts the `message` field from a client error's JSON body (map or JSON-encoded string), falling back to the generic message. The existing `extractErrorMessage` and its callers are untouched.
- `constants.bal`: adds the `ERR_MESSAGE` constant.

## Notes

- Companion change (improves the upstream 400 message texts): wso2-enterprise/digiops-sales#1608 — the two changes are independent and can be deployed in any order.

## Verification

- `bal build` (2201.12.10) passes.
- Verified end-to-end against a mock upstream returning a 400 with a `{"message": ...}` body: the exact upstream message is returned by the endpoint.

Fixes wso2-enterprise/wso2-digital-team-project-management#620

### DIFF
--- a/apps/customer-portal/backend/service.bal
+++ b/apps/customer-portal/backend/service.bal
@@ -5009,7 +5009,7 @@ service http:InterceptableService / on new http:Listener(9090, listenerConf) {
                         userInfo.userId}`, response);
                 return <http:BadRequest>{
                     body: {
-                        message: extractErrorMessage(response)
+                        message: extractClientErrorMessage(response)
                     }
                 };
             }
--- a/apps/customer-portal/backend/service.bal
+++ b/apps/customer-portal/backend/service.bal
@@ -5009,7 +5009,7 @@ service http:InterceptableService / on new http:Listener(9090, listenerConf) {
                         userInfo.userId}`, response);
                 return <http:BadRequest>{
                     body: {
-                        message: "Invalid request parameters for creating registry token."
+                        message: extractErrorMessage(response)
                     }
                 };
             }
--- a/apps/customer-portal/backend/utils.bal
+++ b/apps/customer-portal/backend/utils.bal
@@ -224,14 +224,26 @@ public isolated function getStatusCode(error err) returns int {
     return statusCodeValue is int ? statusCodeValue : http:STATUS_INTERNAL_SERVER_ERROR;
 }
 
-# Get HTTP status code from the given error.
+# Get the error message from the given error.
 #
 # + err - Error to handle
 # + return - Error message
 public isolated function extractErrorMessage(error err) returns string {
     map<anydata|readonly> & readonly errorDetails = err.detail();
-    anydata|readonly errorMessage = errorDetails[ERR_BODY] ?: ();
-    return errorMessage is string ? errorMessage : UNEXPECTED_ERROR_MSG;
+    anydata|readonly errorBody = errorDetails[ERR_BODY] ?: ();
+    if errorBody is map<anydata> {
+        anydata message = errorBody[ERR_MESSAGE];
+        return message is string ? message : UNEXPECTED_ERROR_MSG;
+    }
+    if errorBody is string {
+        json|error parsedBody = errorBody.fromJsonString();
+        if parsedBody is map<json> {
+            json message = parsedBody[ERR_MESSAGE] ?: ();
+            return message is string ? message : errorBody;
+        }
+        return errorBody;
+    }
+    return UNEXPECTED_ERROR_MSG;
 }
 
 # Log forbidden project access attempt.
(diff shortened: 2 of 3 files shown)


apps/customer-portal/backend/utils.bal
```diff
@@ -224,14 +224,26 @@ public isolated function getStatusCode(error err) returns int {
...
+    anydata|readonly errorBody = errorDetails[ERR_BODY] ?: ();
+    if errorBody is map<anydata> {
+        anydata message = errorBody[ERR_MESSAGE];
+        return message is string ? message : UNEXPECTED_ERROR_MSG;
+    }
+    if errorBody is string {
+        json|error parsedBody = errorBody.fromJsonString();
+        if parsedBody is map<json> {
+            json message = parsedBody[ERR_MESSAGE] ?: ();
+            return message is string ? message : errorBody;
```
shayanmalinda: **🟡 Minor: Parsed-JSON branch falls back to the raw body instead of `UNEXPECTED_ERROR_MSG`**

When a string body parses to a JSON map without a string `message` field, this returns the raw JSON blob to the caller (which the registry-token endpoint now shows to end users), while the `map<anydata>` branch above falls back to `UNEXPECTED_ERROR_MSG` — the two branches should degrade the same way. Also `?: ()` on the line above is a no-op since `json` already includes `()`; use `return message is string ? message : UNEXPECTED_ERROR_MSG;` and drop the `?: ()`.

---

apps/customer-portal/backend/service.bal
```diff
@@ -5009,7 +5009,7 @@ service http:InterceptableService / on new http:Listener(9090, listenerConf) {
                         userInfo.userId}`, response);
                 return <http:BadRequest>{
                     body: {
-                        message: "Invalid request parameters for creating registry token."
+                        message: extractErrorMessage(response)
```
shayanmalinda: **🔵 Question: Upstream 400 bodies are now shown to customers verbatim — is every current and future 400 path user-safe?**

This forwards whatever the registry-subscription-service puts in a 400 body, which includes Ballerina's auto-generated payload-binding 400s (raised before the resource code runs) — if the two services' payload types ever drift, customers would see internal binding details like field names and types. Worth confirming the upstream messages are treated as a user-facing contract (the companion digiops-sales PR does this for the hand-written paths), or filtering here to known-safe shapes.

---

shayanmalinda (commented):
Reviewed error message pass-through for registry token creation: 2 comments (1 minor fallback inconsistency, 1 design question).

apps/customer-portal/backend/utils.bal
shayanmalinda: Aligned the fallback: the parsed-JSON branch now returns `UNEXPECTED_ERROR_MSG` when the map has no string `message`, and the redundant `?: ()` is removed (d4b82320e).

---
<sub>Reply drafted by Claude on behalf of @shayanmalinda</sub>

apps/customer-portal/backend/service.bal
shayanmalinda: Yes — the upstream 400 messages are treated as a user-facing contract: the companion PR (wso2-enterprise/digiops-sales#1608) rewrites all hand-written 400 bodies in `POST /robot-accounts` to customer-appropriate text, and the max-limit message was already written for end users. Payload-binding drift is mitigated because the portal builds the upstream payload from its own validated types with the same `robotName` constraint (modules/registry/types.bal), and with the fallback fix (d4b82320e) any JSON body without a string `message` degrades to the generic message rather than leaking raw content. Keeping this open-by-design for pass-through of future upstream 400 paths is intentional so new upstream messages reach users without a portal release.

---
<sub>Reply drafted by Claude on behalf of @shayanmalinda</sub>

apps/customer-portal/backend/utils.bal
shayanmalinda: Reworked in bd7cebb32: `extractErrorMessage` is restored to its original behavior (its existing callers are untouched), and the JSON-body handling now lives in a dedicated `extractClientErrorMessage` used only by this endpoint. The consistent `UNEXPECTED_ERROR_MSG` fallback from this thread is preserved in the new function.

---
<sub>Reply drafted by Claude on behalf of @shayanmalinda</sub>

cloby99 (approved):

PR #1143, opened 2026-07-14
[CSM][Web] Rich text editors for the change-request planning fields and work log comment

Hesara2003:
## Purpose
Adds the existing Lexical rich text editor to form fields that ServiceNow itself treats as rich text, but this portal only offered as plain textareas — matching ServiceNow's own form behavior for these fields.

> **Scope note:** Scoped to fields where there's real evidence the backend expects/renders HTML (or, for one write-only field, no risk either way) — not a blanket sweep. See Approach for exactly what was checked and excluded.

## Goals
- Let users actually produce formatted (bold/italic/lists/links) content for change-request planning fields that the detail page already renders as HTML
- Match ServiceNow's own work-notes convention for the time card work log comment
- Avoid corrupting rich text content via naive character-limit truncation
- Avoid silently accepting a "blank" rich text field as filled in

## Approach
- `CreateChangeRequestPage.tsx`: Description, Justification, Implementation plan, Risk and impact analysis, Backout plan, and Test plan now use `Editor` (`@components/rich-text-editor`, Lexical-based, already used elsewhere for case descriptions/comments and security reports) instead of plain `TextField`s. `CsmChangeRequestDetailPage.tsx` already sanitizes and renders these via `dangerouslySetInnerHTML` — the create side just had no way to produce real formatting for it to render. "Additional comments"/"Internal work note" on the same form are deliberately left plain — they're write-only fields, absent from the detail schema entirely, so there's no evidence the backend treats them as HTML.
- `LogTimeCardDialog.tsx`: the "Work log comment" field now uses the same editor. This field is write-only (accepted on create, never read back or displayed anywhere in the portal), so there's no direct evidence either way, but also no risk of a broken display. By contrast, `TimeCardReviewDialog`'s "Lead's comment" is deliberately **not** touched — it comes back as `rejectionReason` and is rendered via plain `Typography` elsewhere, so making it rich text would send HTML that then displays as literal tags.
- Dropped the old live character-limit truncation (`.slice(0, MAX)` on every keystroke) on all of these fields: safe for a plain string, but truncating HTML at a fixed offset risks cutting a tag in half and corrupting the markup. The backend still returns a real validation error on submit if a field is too long.
- Switched the "is this field empty" check from `.trim()` truthiness to `isBlankHtml` (existing utility, already used by the detail page for the same purpose) on every affected field. An untouched Lexical editor's HTML output (e.g. `<p><br></p>`) is non-empty as a *string*, so `.trim()` would have treated a genuinely blank field as filled in.
- Also audited every other `multiline` field in the app (call-request dialogs, case resolution notes, GitHub issue body, deployment/deployed-product descriptions, catalog variables) against how each is actually rendered downstream. All confirmed genuinely plain text (or, for the GitHub issue body, Markdown — a different format entirely) — no other gaps found.

## User stories
As a CS engineer filling out a change request, I can format the planning fields (bold, lists, links) the same way I would in ServiceNow directly.
As a CS engineer logging time, I can format my work log comment the same way.

## Release note
Change request creation and the time card "Log time" dialog now support rich text formatting (bold, italic, lists, links) in their long-form fields, matching ServiceNow's own form behavior.

## Documentation

## Automation tests
- Unit tests: added 2 new tests for `timeCardDraftErrors`' HTML-blank detection (`timeCardTotals.test.ts`, 8/8 passing); full webapp suite 223 passed / 2 failed, and those 2 (`CaseActionBar.test.tsx`) are pre-existing failures unrelated to this change.
- Verified against staging: yes, live. Opened `/operations/change-requests/new`, confirmed all 6 editors render with their toolbars, typed into two independent fields to confirm no cross-talk, confirmed the Create button's gating is unaffected. Opened a real case's Time tracking tab → Log time dialog, typed into the work log comment editor, confirmed clean text (no corruption), the character counter updating correctly, and no console errors.

## Security checks

## Test environment
- `./node_modules/.bin/tsc -b`, `./node_modules/.bin/eslint src`, `./node_modules/.bin/vitest run`, and `./node_modules/.bin/vite build` all passing locally.

## Summary by CodeRabbit

- **New Features**
  - Added rich-text editing for change request planning fields (description, justification, implementation plan, risk/impact, backout, and test plan).
  - Added rich-text editing for the timecard “Work log comment” field.
- **Bug Fixes**
  - Improved validation to treat rich-text placeholders/empty HTML as blank (so they aren’t submitted).
  - Prevented rich-text markup from being truncated while typing; updated Enter-to-submit to work correctly with editor contenteditable elements.
- **Tests**
  - Added test coverage for rich-text blank vs. populated work log comments.

### DIFF
--- a/apps/csm-portal/webapp/src/features/csm-operations/pages/CreateChangeRequestPage.tsx
+++ b/apps/csm-portal/webapp/src/features/csm-operations/pages/CreateChangeRequestPage.tsx
@@ -37,6 +37,8 @@ import { useRef, useState, type JSX } from "react";
 import { useNavigate } from "react-router";
 import { BackendApiError } from "@api/backend/client";
 import { useErrorBanner } from "@context/error-banner/ErrorBannerContext";
+import Editor from "@components/rich-text-editor/Editor";
+import { isBlankHtml } from "@utils/sanitizeHtml";
 import { usePostChangeRequest } from "@features/csm-operations/api/usePostChangeRequest";
 import { useGetUsersMe } from "@features/settings/api/useGetUsersMe";
 import { useSearchGroups } from "@api/useSearchGroups";
@@ -65,11 +67,13 @@ const UNSET = "";
 const SELECT_PLACEHOLDER = "-- Select --";
 
 // Field limits and defaults below mirror the legacy ServiceNow "Create New
-// Change Request" form (Short description: 500 chars; the five Planning
-// journal fields: 4000 chars each; Type/Category/Impact/Risk pre-selected
-// rather than left blank).
+// Change Request" form (Short description: 500 chars; Type/Category/Impact/
+// Risk pre-selected rather than left blank). The Planning journal fields
+// (Description onward) are ServiceNow rich-text fields — no character cap
+// here, since naively truncating HTML at a fixed offset risks cutting a tag
+// in half and corrupting the markup; the backend rejects an overlong
+// submission with a real error instead (see handleSubmit's onError).
 const SUBJECT_MAX = 500;
-const PLAN_FIELD_MAX = 4000;
 
 const TYPE_OPTIONS: Array<{ value: BeChangeRequestType; label: string }> = [
   { value: "standard", label: "Standard" },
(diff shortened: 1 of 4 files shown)


Hesara2003:
Both fixed: Enter-key-submits bug (isContentEditable check, dd12420) and dangling component="label" dropped in both files (dd12420, e831cfd).

s-sandali:
@coderabbitai review

cloby99 (approved):

PR #1162, opened 2026-07-16
feat(customer-portal): close Novera chats + auto-abandon on early case creation

SasmithaDilshan:
Implements wso2-enterprise/digiops-cs#2353 — let users close a Novera chat from history, plus a stats fix for chats where a case is created before the assistant responds.

## What & why
- **Close chat:** a "Close" action on resumable chats in chat history moves them to a terminal state so they can't be accidentally resumed, while keeping the chat visible and any created case intact.
- **Auto-abandon:** if a user asks a question and clicks *Create Case* **before Novera responds**, the conversation was being left as **Open**, skewing stats. Those are now moved to a terminal state.

### Design update (supersedes the original "reuse Abandoned, no new state" plan)
The original plan was to reuse the existing `Abandoned` state for both flows with no ServiceNow change. We've since split them into **two distinct terminal states**, so a customer-initiated close is recorded differently from a system auto-abandon:
- **User-initiated close → new `Close` state (id 6).**
- **Auto-abandon (case created before any AI response) → `Abandoned` (id 5).**

The `Close` choice has been added on the ServiceNow side (state key 6), and the entity-service now accepts state keys 5/6 (wso2-enterprise/digiops-cs#2547, merged; constant-naming follow-up wso2-enterprise/digiops-cs#2548).

## Changes
**Backend**
- Single `PATCH /conversations/:id` taking `{ "status": "closed" | "abandoned" }`, mapped to the configured state id and forwarded to the internal `updateConversation`. Replaces the earlier separate `/close` + `/abandon` routes — one handler, invalid status → `400`.
- `conversationStateIds.close` added with a **default (6)** so environment configs that override the map without this newer field still resolve (closed record, avoids a config-resolution break).

**Frontend**
- `useUpdateConversationState(projectId, status)` — one hook for both close and abandon (`PATCH` + `{ status }`).
- `AllConversationsList`: a **Close** action, shown only for resumable chats; confirmation dialog + mutation owned by `AllConversationsPage`.
- `conversationsList.ts`: terminal states (`Abandoned`, `Resolved`, `Converted`, `Close`) resolve to **View** (not Resume) + `isConversationResumable()` helper. State matching is **exact** (won't mis-treat `Reopened`/`Inactive` as resumable). Disables chat input in the resume UI for non-resumable chats.
- `NoveraChatPage`: auto-abandon on create-case-before-AI-response; the seeded greeting is excluded from the "assistant responded" check via `NOVERA_WELCOME_MESSAGE_ID`.
- Distinct `Close` status chip colour + icon.

## Dependencies
- ServiceNow `Close = 6` choice on the conversation state field — **done**.
- entity-service PATCH allow-list accepts 5/6 — wso2-enterprise/digiops-cs#2547 (**merged**); naming follow-up #2548.

## AC coverage
- ✅ Close option in history (resumable chats only)
- ✅ Closed chats can't be resumed / no input
- ✅ Case association intact (one-way `conversationId` link — unaffected)
- ✅ Closed chats demarcated in history (status chip colour-codes state, including Close)

## Validation
- `tsc --noEmit`: 0 errors
- `eslint`: clean on changed files (2 *pre-existing* `preserve-manual-memoization` errors remain on `AllConversationsPage`'s search-request memo — present on `main`, not introduced here)
- Unit tests: `conversationsList` 7/7 (terminal = View, `isConversationResumable`, exact-match regressions for `Reopened`/`Inactive`)
- ⚠️ Backend endpoint is **code-reviewed, not runtime-tested** (no local Ballerina runtime). A unit test for the auto-abandon predicate is a tracked follow-up.

## Summary by CodeRabbit

* **New Features**
  * Added the ability to close resumable conversations from the conversation list.
  * Added a confirmation dialog before closing a conversation, with loading and error states.
  * Conversations started without a bot response are automatically marked terminal when creating a case.
  * Added distinct closed-status styling and iconography.

* **Bug Fixes**
  * Improved conversation-state classification to avoid incorrectly treating terminal or similarly named states as resumable.


chanukaranaba (commented):
## Code Review — close Novera chats + auto-abandon

Reviewed the two new backend routes, the frontend hooks/pages, and the resumability utils. Verified: both new routes extract user info via the auth interceptor and delegate authorization to the entity service (401/403 mapped); Apache headers are correct for this repo; the change is SQL-free (delegates to the entity PATCH); and `isConversationResumable` is well tested, including the substring edge cases ("Reopened"/"Inactive"). Findings below.

### Should fix
- **The code contradicts the PR's stated design, and the new state may not exist downstream.** The description says the agreed approach was to *reuse `Abandoned`, add no new "Closed" state, and require no ServiceNow change* — "only new route: `/abandon`." But the diff adds a new `close: 6` state (`types.bal`/`utils.bal`), a `/close` route, and wires the user-facing Close button to it (`useCloseConversation → /close → stateKey 6`). Auto-abandon uses `Abandoned` (5). So user-initiated close writes **state id 6**, whose existence in the entity-service/ServiceNow isn't confirmed and which the PR itself says was never meant to be created. Since the backend is *"code-reviewed, not runtime-tested,"* a `stateKey: 6` the entity service doesn't recognize will fail only in production. Resolve before merge: either (a) drop the `/close` + state-6 path and point the Close button at `/abandon` (matching the described design), or (b) update the description and confirm state 6 "Close" is configured in every environment.
- **Adding a required field to the `conversationStateIds` closed record is a config-compatibility risk.** `ConversationStateIds` is `{| ... |}` (closed); every environment's `Config.toml` that sets `conversationStateIds` must now also supply `close`, or config resolution breaks / silently falls back to the default `6`. Confirm all env configs are updated in lockstep.
- **The riskiest new logic — auto-abandon — has no test.** The `userAsked && !aiResponded` decision (with the welcome-message exclusion) and the deferred-abandon-on-`conversation_created` path are the trickiest additions and are untested. Extract the predicate into a pure function and unit-test it (answered/unanswered, welcome-only, id-not-yet-available). The `conversationsList` tests are good; this is the gap.

### Consider
- **Backend handler duplication.** `/abandon` and `/close` are ~95% identical — same auth extraction, same 401/403/500 mapping, differing only in `stateKey` and one string. Extract a private `updateConversationState(ctx, id, stateKey, verb)` helper; keeps `service.bal` lean.
- **Wrong verb in the `/abandon` Forbidden message.** It reads *"You're not authorized to **close** the requested conversation."* in the abandon handler (copy-paste) — and abandon is machine-triggered, not a user "close."
- **Frontend hook duplication.** `useAbandonConversation` and `useCloseConversation` are near-identical (85 lines each); a shared factory parameterized by action removes ~80 duplicated lines and keeps cache-invalidation logic in one place.
- **Magic string `m.id !== "1"`** for the welcome message — reference a named welcome-message-id constant so the `aiResponded` check can't silently break if that id changes (a broken check would count the welcome as a response and never abandon).
- **Deferred abandon doesn't re-check `aiResponded`** — it fires unconditionally when `conversation_created` arrives. Fine per the AC, just noting the small window.

**Verdict:** The frontend resumability model and tests are solid, but the Close/state-6 path materially diverges from the described (and apparently agreed) design and writes a possibly-unconfigured state through an untested endpoint — that needs resolving before merge.

---

*Reviewed with Claude Code per WSO2 Digital code-review process.*

apps/customer-portal/backend/modules/entity/utils.bal
```diff
@@ -25,7 +25,7 @@ public configurable CaseStateIds & readonly caseStateIds = {
     workInProgress: 10
 };
 public configurable ConversationStateIds & readonly conversationStateIds =
-    {open: 1, active: 2, resolved: 3, converted: 4, abandonded: 5};
+    {open: 1, active: 2, resolved: 3, converted: 4, abandonded: 5, close: 6};
```
chanukaranaba, line 28: **Should fix — this new `close: 6` state contradicts the PR's stated design and may not exist downstream.**

The PR description says the agreed approach was to *reuse `Abandoned` and add no new "Closed" state / no ServiceNow change* — yet this adds state id `6` plus a `/close` route, and the UI Close button writes it (`useCloseConversation → /close → stateKey 6`). Two concrete risks:

1. **Unconfigured state.** If the entity-service/ServiceNow has no state id 6, `updateConversation` with `stateKey: 6` fails — and the backend is *code-reviewed, not runtime-tested*, so this surfaces only in production.
2. **Closed-record config break.** `ConversationStateIds` is `{| ... |}` (closed), so every environment's `Config.toml` that overrides `conversationStateIds` must now also supply `close`, or config resolution errors / silently uses this default `6`.

Either drop `/close` + state 6 and point the Close button at `/abandon` (matching the described design), or update the description and confirm state 6 "Close" is provisioned in every environment before merge.

apps/customer-portal/backend/service.bal
```diff
@@ -2066,6 +2066,105 @@ service http:InterceptableService / on new http:Listener(9090, listenerConf) {
         return mapConversationResponse(conversationResponse);
     }
 
+    # Abandon a conversation, moving it to a terminal state so it can no longer
+    # be resumed. Used when a case is created from a chat before the assistant
+    # has responded.
+    #
+    # + id - ID of the conversation to abandon
+    # + return - Ok on success or error
+    resource function post conversations/[entity:IdString id]/abandon(http:RequestContext ctx)
```
Rashmika998: Instead of having [2 different endpoints](https://github.com/wso2-open-operations/cs-tools/pull/1162/changes#diff-eca07fe60f4c779acb6655f1aa59203532751ff527a16d51de371ef62ea3259eR2124), let's make one by adding this action to the payload.

we could have an endpoint named(if already exists, let's use it),
```
PATCH /conversations/{id}
```
and in the body, 
```
{
  "status": "closed"
}
```

apps/customer-portal/backend/modules/entity/types.bal
```diff
@@ -2152,6 +2152,8 @@ public type ConversationStateIds record {|
     int resolved;
     # Abandoned state ID
     int abandonded;
+    # Close state ID
+    int close;
```
chanukaranaba: **Should fix — adding a required field to this closed record is a config-compatibility risk.** `ConversationStateIds` is `{| ... |}` (closed), so every environment's `Config.toml` that overrides `conversationStateIds` must now also supply `close`, or config resolution errors / silently falls back to the default `6`. Confirm all env configs are updated in lockstep — and see the note on the value `6` in `utils.bal` (the state may not exist in the entity-service/ServiceNow).

apps/customer-portal/webapp/src/features/support/pages/NoveraChatPage.tsx
```diff
@@ -333,12 +338,35 @@ export default function NoveraChatPage(): JSX.Element {
   const handleCreateCase = useCallback(() => {
     setIsCreateCaseLoading(true);
 
+    // If the user creates a case before Novera has produced any response, mark
+    // the conversation Abandoned so it isn't left counted as an open chat.
+    // ("1" is the static welcome message id; real bot replies have non-empty text.)
+    const userAsked = messages.some((m) => m.sender === ChatSender.USER);
+    const aiResponded = messages.some(
+      (m) => m.sender === ChatSender.BOT && m.id !== "1" && m.text.trim() !== "",
```
chanukaranaba: **Should fix — this is the riskiest new logic and it's untested.** The `userAsked && !aiResponded` decision (plus the deferred-abandon path when the id isn't ready yet) is the trickiest addition in the PR. Extract this predicate into a pure function and unit-test it: answered, unanswered, welcome-message-only, and id-not-yet-available.

**Consider — the `m.id !== "1"` magic string.** No named constant for the welcome-message id turned up in the support feature. If that id ever changes, this check silently counts the welcome as a real response and the conversation is never abandoned. Reference a named constant instead.

apps/customer-portal/webapp/src/features/support/pages/NoveraChatPage.tsx
```diff
@@ -469,6 +497,11 @@ export default function NoveraChatPage(): JSX.Element {
           const nextConversationId = String(event.conversationId ?? "");
           if (nextConversationId) {
             setConversationId(nextConversationId);
+            // The user created a case before the id was available — abandon now.
+            if (pendingAbandonRef.current) {
```
chanukaranaba, line 508: **Consider —** the deferred abandon fires unconditionally when `conversation_created` arrives, without re-checking `aiResponded`. Fine per the AC (the user did click Create Case before any response), just flagging the small window between setting `pendingAbandonRef` and this handler.

apps/customer-portal/backend/service.bal
```diff
@@ -2066,6 +2066,105 @@ service http:InterceptableService / on new http:Listener(9090, listenerConf) {
...
+        }
+        return http:OK;
+    }
+
+    # Close a conversation, moving it to a terminal state so it can no longer
+    # be resumed. Used when a user explicitly closes a chat from the chat list.
+    #
+    # + id - ID of the conversation to close
+    # + return - Ok on success or error
+    resource function post conversations/[entity:IdString id]/close(http:RequestContext ctx)
```
chanukaranaba: **Consider — handler duplication.** This `/close` resource is ~95% identical to `/abandon` above: same user-info extraction, same 401/403/500 mapping, differing only in `stateKey` and one message string. Extract a private `updateConversationState(ctx, id, stateKey, verb)` helper to keep `service.bal` lean (and note the broader question of whether `/close` should exist at all — see the `utils.bal` comment).

apps/customer-portal/backend/service.bal
```diff
@@ -2066,6 +2066,105 @@ service http:InterceptableService / on new http:Listener(9090, listenerConf) {
...
+                        message: ERR_MSG_UNAUTHORIZED_ACCESS
+                    }
+                };
+            }
+
+            if getStatusCode(response) == http:STATUS_FORBIDDEN {
+                log:printWarn(string `User: ${userInfo.userId} is forbidden to update conversation with ID: ${id}!`);
+                return <http:Forbidden>{
+                    body: {
+                        message: "You're not authorized to close the requested conversation."
```
chanukaranaba: **Consider — wrong verb.** This is the `/abandon` handler, but the Forbidden message says "...not authorized to **close** the requested conversation." (copy-paste from `/close`). Abandon is machine-triggered, so this message is doubly off — say "abandon" or a neutral "update".

apps/customer-portal/webapp/src/features/support/api/useCloseConversation.ts
```diff
@@ -0,0 +1,85 @@
...
+/**
+ * Closes a Novera conversation via POST /conversations/:id/close, moving it to
+ * the terminal Close state so it can no longer be resumed. Used when a user
+ * explicitly closes a chat from the chat list. On success, refreshes the
+ * conversation list and stats for the project.
+ *
+ * @param {string} projectId - Project ID for cache invalidation.
+ * @returns {UseMutationResult<void, Error, string>} Mutation keyed by conversationId.
+ */
+export function useCloseConversation(
```
chanukaranaba: **Consider — hook duplication.** `useCloseConversation` and `useAbandonConversation` are near-identical (85 lines each), differing only in the URL path (`/close` vs `/abandon`) and the log tag. A shared factory parameterized by action removes ~80 duplicated lines and keeps the cache-invalidation logic in one place.

cloby99 (approved):

PR #1171, opened 2026-07-19
[Customer Portal] Source Usage & Metrics products from per-product metrics/usage-counts APIs

cloby99:
## Summary

- Replace the bulk-derived usage/metrics data in the Usage & Metrics panel with per-product calls to the new `POST /deployments/{deploymentId}/products/{productId}/metrics/search` and `.../metrics/usage-counts/search` endpoints, sourcing the product list itself from `products/search`.
- Each product accordion row now fetches its own core-usage and usage-count metrics independently, replacing the old bulk-derivation helper and its tests.
- Combine product name and version into a single accordion title, and drop the now-unused `ProductAccordionRowProps`/`ProductExpandedViewProps` types and `UsageEnvironmentProduct` import.
- Update component/hook mocks and tests to match the new per-product hook shape.
- Fix a `countTypes` payload-binding failure (500) between the customer-portal backend and the entity service: `countTypes` was typed as `map<int>` but ServiceNow returns per-type aggregate objects (`{aggregation, min, max, avg}`) — added a `CountTypeAggregation` record and updated both type layers to match.

## Test plan

- [ ] Verify each product row in Usage & Metrics loads core-usage and usage-count charts independently
- [ ] Verify accordion titles show combined product name + version
- [ ] Confirm `usage-counts/search` no longer 500s and charts render real aggregate values
- [ ] `npm run test` / `npx tsc --noEmit` pass

## Summary by CodeRabbit

* **New Features**
  * Added product-level metrics and usage-count searches for deployment views.
  * Product usage panels now display per-product summaries, trends, charts, and latest instances.
  * Added detailed aggregation values, including minimum, maximum, and average results.
  * Added support for searching metrics across a selected date range.

* **Improvements**
  * Product data now loads independently, with loading states shown per product.
  * Updated product display and interaction behavior to include version information.


apps/customer-portal/backend/modules/types/types.bal
```diff
@@ -2205,7 +2217,7 @@ public type DeployedProductMetricsUsageCountsSummary record {|
         string 'end;
     |} dateRange;
     # Count types aggregated over the range
-    map<int> countTypes;
+    map<CountTypeAggregation> countTypes;
```
Rashmika998, line 2220: Let's update the `openapi.yaml` file

Rashmika998 (approved):

PR #1174, opened 2026-07-19
[CSM][Web] Align QuickNav search palette with customer-portal's style

Rashmika998:
## Summary
- Restyles the header QuickNav/search palette (Cmd+K) to visually match customer-portal's search dropdown, while keeping csm-portal's extra command-palette behavior (Pinned/Recent/Pages sections, keyboard navigation, Cmd+K shortcut).
- Case results (live search, and Pinned/Recent entries snapshotted from the case detail page) now render as rich cards with severity/status/case-type chips and updated/assignee metadata, reusing the existing `SeverityChip`/`StateChip` components.
- Swaps the `Dialog` for `Modal` + a plain `Paper` (both from `@wso2/oxygen-ui`) so the theme's translucent "acrylic" background applies automatically — matching the glassy look of customer-portal's dropdown instead of `Dialog`'s more opaque default styling.
- Adds a skeleton loading state for the "Cases" section and an illustrated empty state; hides the "Pages" section on the empty-query default view (still matches when typed); fixes an autofocus race that required a second click before typing worked; widened the search trigger.

## Test plan
- [x] `tsc -b --noEmit` passes
- [x] `eslint` passes on all touched/new files
- [x] `vitest run` passes (`useRecentViews.test.ts`, updated for the per-user storage key)
- [x] `vite build` succeeds
- [x] Manually verified in the running app: Cmd+K palette opens, case search shows rich cards, Pinned/Recent cases show the same rich card, glassy background renders correctly, autofocus works on first click, Pages hidden until typing

## Summary by CodeRabbit

* **New Features**
  * Enhanced quick navigation with rich case cards (severity, state/work state, type, updated time, and assignee).
  * Added “no results” empty-state illustration and quick-nav result skeletons.
  * Added reusable work-state chip and persisted richer case metadata for recents/pins.

* **Bug Fixes**
  * Recent views are now stored separately per signed-in user.
  * Improved quick-navigation focus behavior when opening.


apps/csm-portal/webapp/src/features/csm-cases/api/useQuickCaseSearch.ts
```diff
@@ -70,6 +85,13 @@ export function useQuickCaseSearch(
         caseNumber: c.number,
         wso2CaseId: c.internalId,
         subject: c.subject ?? "(no subject)",
+        severity: severityFromPriority(c.severity),
+        state: (c.state ?? "open") as CaseState,
+        workState: c.workState,
+        caseType: c.type,
+        updatedOn: c.updatedOn,
+        createdOn: c.createdOn,
+        assigneeName: c.assignedEngineer?.name,
```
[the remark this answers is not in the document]
Rashmika998: Fixed — now mapping through `uiStateFromBe(c.state)` instead of a raw cast, matching the normalization the rest of the app (`useGetCsmCaseDetail`, `useGetCsmCases`) already applies. Pushed in 0d8f4901a.

apps/csm-portal/webapp/src/features/csm-recent/components/QuickNavCaseCard.tsx
```diff
@@ -0,0 +1,123 @@
...
+          />
+        )}
+        {hit.caseType && (
+          <Chip
+            size="small"
+            variant="outlined"
+            label={CASE_TYPE_LABEL[hit.caseType]}
+            sx={{ height: 20, fontSize: "0.75rem" }}
+          />
+        )}
```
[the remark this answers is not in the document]
Rashmika998: Fixed — now guarding on `hit.caseType && CASE_TYPE_LABEL[hit.caseType]`, mirroring the existing guard in `CsmCaseDetailPage`. Pushed in 0d8f4901a.

shayanmalinda (commented):
Reviewed the QuickNav palette restyle. The Modal+Paper swap, per-user recent-views scoping, and rich case cards all look solid; leaving a few findings — a skeleton/results overlap during background refetches, the lifecycle of the `.pending` recent-views bucket, a duplicated work-state chip, and two smaller nits.

apps/csm-portal/webapp/src/features/csm-recent/components/QuickNav.tsx
```diff
@@ -74,6 +85,14 @@ export default function QuickNav(): JSX.Element | null {
   // API-backed case lookup: a CS/WSO2 id (or any subject text) resolves to real
   // cases. Disabled until the query is long enough (see the hook).
   const caseSearch = useQuickCaseSearch(open ? debouncedQuery : "");
+  // True while a case search is in flight (or its result is for a stale
+  // query) — drives the "Cases" section's skeleton independently of whether
+  // Pinned/Recent/Pages already have matches to show.
+  const casesLoading =
```
shayanmalinda, line 91: **🟡 Minor: Skeleton renders above real results during a background refetch**

`casesLoading` is true whenever `isFetching` is — including a background refetch of an already-rendered result set (e.g. retyping a previous query after the 15s `staleTime`, with `refetchOnMount: true`), so the skeleton block and the real "Cases" section render together with two "Cases" headers. Gate the skeleton on having nothing to show yet, e.g. `casesLoading && !caseSearch.data`.

---

apps/csm-portal/webapp/src/features/csm-recent/hooks/useRecentViews.ts
```diff
@@ -52,13 +54,69 @@ export interface RecentView {
...
+/**
+ * The signed-in user's stable id (ID token `sub` claim), or `null` before
+ * it's resolved. Every read/write is scoped under this so a different
+ * engineer signing in on the same browser never sees a previous user's
+ * recent/pinned cases — see {@link useSyncRecentViewsIdentity}.
+ */
+let activeUserKey: string | null = null;
+
+function currentStorageKey(): string {
+  return `${STORAGE_KEY_BASE}.${activeUserKey ?? "pending"}`;
```
shayanmalinda, line 88: **🟠 Medium: Views recorded before identity resolves are stranded in a shared `.pending` bucket**

`useIdTokenClaims` decodes the token asynchronously, so any view recorded in that window lands in `csm.recentViews.v1.pending` — it disappears from Recents once `sub` resolves, is never cleaned up, and is readable by the next user on the same browser during their own pre-resolution window (or for the whole session if the token decode fails and `activeUserKey` stays `null`), which is the leak this scoping set out to prevent. Consider migrating or clearing the pending bucket in `useSyncRecentViewsIdentity` once `sub` resolves.

---

apps/csm-portal/webapp/src/features/csm-recent/components/QuickNavCaseCard.tsx
```diff
@@ -0,0 +1,123 @@
...
+        sx={{ flexWrap: "wrap" }}
+      >
+        {idLabel && (
+          <Typography variant="body2" fontWeight={500} color="text.secondary">
+            {idLabel}
+          </Typography>
+        )}
+        <SeverityChip severity={hit.severity} />
+        <StateChip state={hit.state} variant="outlined" />
+        {hit.state === "work_in_progress" && hit.workState && (
```
shayanmalinda, line 76: **🟡 Minor: Work-state chip duplicates the one in `CasesList`**

This inline `Chip` (outlined, `paused → warning`, `fontWeight: 600`, `WORK_STATE_LABEL`) is the same construction as `CasesList.tsx:266-272`, so the two will drift when one changes. Extract a `WorkStateChip` beside `SeverityChip`/`StateChip` — both of which exist precisely as the "single source of truth" for their badges — and use it in both places.

> 📖 **Reference ([duplicate-code](https://refactoring.guru/smells/duplicate-code)):** "Duplicate Code / Copy-Paste Programming — the same logic in more than one place; fixes/bugs must be applied N times and drift. Extract a shared function/module/factory."

---

apps/csm-portal/webapp/src/features/csm-recent/components/QuickNav.tsx
```diff
@@ -231,18 +274,41 @@ export default function QuickNav(): JSX.Element | null {
...
+            width: { xs: "calc(100% - 32px)", sm: "calc(100% - 64px)" },
+            maxWidth: 760,
+            maxHeight: "65vh",
+            display: "flex",
+            flexDirection: "column",
+            overflow: "hidden",
+            outline: "none",
+          }}
         >
+          <Box
```
shayanmalinda, line 314: **🟡 Minor: Wrapped block kept its old indentation**

The children of this `Box` (down to the stacked closing tags at the bottom of the results pane) sit at the same indent level as the `Box` itself, so the nesting reads as siblings. Re-indent the wrapped block one level.

---

apps/csm-portal/webapp/src/features/csm-cases/pages/CsmCaseDetailPage.tsx
```diff
@@ -519,6 +519,20 @@ export default function CsmCaseDetailPage(): JSX.Element {
...
+      caseHit: {
+        caseNumber: data.caseNumber,
+        wso2CaseId: data.wso2CaseId,
+        subject: data.subject,
+        severity: data.severity,
+        state: data.state,
+        workState: data.workState,
+        caseType: data.caseType,
+        updatedOn: data.updatedAt,
+        assigneeName: data.assignee !== "Unassigned" ? data.assignee : undefined,
```
shayanmalinda: **🟡 Minor: Infers "no assignee" by comparing against a display fallback string**

`data.assignee !== "Unassigned"` round-trips the display label produced in `useGetCsmCaseDetail.ts:41` (`c.assignedEngineer?.name?.trim() || "Unassigned"`) — if that fallback ever changes wording or casing, this silently snapshots the literal as a person's name. Prefer exposing the raw `assignedEngineer?.name` on the detail model (or sharing the sentinel as an exported constant) instead of comparing the label.

> 📖 **Reference ([magic-string](https://en.wikipedia.org/wiki/Magic_string)):** "Magic Strings / Magic Numbers — unexplained literals, or the *same* key/value hardcoded in several files instead of one shared, exported constant → silent drift on rename."

---

apps/csm-portal/webapp/src/features/csm-recent/components/QuickNav.tsx
Rashmika998: Confirmed and fixed — gated the skeleton on `casesLoading && !caseSearch.data` instead of `casesLoading` alone, so a background refetch of an already-visible result set no longer renders the skeleton block alongside the real "Cases" section. Pushed in c2c1d1631.

apps/csm-portal/webapp/src/features/csm-recent/hooks/useRecentViews.ts
Rashmika998: Confirmed and fixed — added `migratePendingBucket`, called once `sub` resolves, which folds any views recorded during the pre-resolution window into the resolved user's bucket (existing entries in the target bucket win on conflict) and clears "pending". Also found and fixed a related bug while verifying this: `activeUserKey` is scoped per-tab, and I'd only synced it from one hook mounted in `AppLayout` — a component whose `useRecentViews()` ran before that sync (or the only one mounted in a fresh tab) could get stuck reading "pending" indefinitely even after real resolution elsewhere. Moved the sync into every `useRecentViews`/`useRecordRecentView` call site itself (`useSyncActiveUserKey`) so each resolves its own identity independently. Pushed in c2c1d1631.

apps/csm-portal/webapp/src/features/csm-recent/components/QuickNavCaseCard.tsx
Rashmika998: Confirmed and fixed — extracted `WorkStateChip` (mirroring `SeverityChip`/`StateChip`) and switched both `CasesList` and `QuickNavCaseCard` to it. Pushed in c2c1d1631.

apps/csm-portal/webapp/src/features/csm-recent/components/QuickNav.tsx
Rashmika998: Confirmed and fixed — re-indented the wrapped block one level. Pushed in c2c1d1631.

apps/csm-portal/webapp/src/features/csm-cases/pages/CsmCaseDetailPage.tsx
Rashmika998: Confirmed and fixed — added `CsmCaseDetail.assigneeName` (the raw `assignedEngineer?.name`, no "Unassigned" fallback baked in) and use that directly instead of string-comparing `data.assignee` against the display label. Pushed in c2c1d1631.

cloby99 (approved):

PR #1178, opened 2026-07-20
[Customer Portal MicroApp] Add case feedback UI and smooth touch scrolling

Hesara2003:
## Purpose
The Customer Portal microapp had no way to submit case feedback, even though the backend has fully supported it (`GET`/`POST /cases/{id}/feedback`) for a while — the webapp already has this. Separately, a few internal scroll areas (project switcher list, rendered code blocks, Novera reply tables) felt rough on mobile because they were missing iOS momentum scrolling.

## Goals
- Let users rate a closed case (emoji scale + optional tags + comment) and submit it to the existing feedback API.
- Show a persistent prompt on the Case Details page for closed cases without feedback yet, and a read-only summary once given.
- Enable native momentum scrolling on the app's few internal `overflow: auto` containers.

## Approach
- Ported the case feedback flow from the webapp's existing `CaseFeedbackModal` pattern into the microapp's `services/`, `types/`, and `components/features/support/` layout: new `CaseFeedbackDialog`, feedback types/DTOs, and a `cases.feedback` / `cases.submitFeedback` query pair (404 on GET is treated as "no feedback yet", not an error).
- Wired the dialog into `CaseDetailPage`: opens automatically on resolve / accept-solution, plus a standing "Give Feedback" prompt / read-only summary section.
- Along the way, fixed two dialog bugs surfaced during testing: the comment field losing focus every keystroke (an inline `slots.paper` component was being redefined on every render, forcing MUI to remount it — fixed by hoisting it) and the dialog reading as translucent in dark mode (this theme's default `Dialog` styling is an intentional "frosted glass" look — opacity 0.7, backdrop blur, and a translucent `background.paper` color — which doesn't work for a text-heavy form, so this dialog opts out with a solid, fully opaque background).
- Added `-webkit-overflow-scrolling: touch` to the three `overflow: auto` containers in the app (`ProjectSelector`, `RichText` code blocks, `TypewriterText` tables) — the standard fix for momentum scrolling inside a WebView. No blocking/non-passive scroll or touch listeners were found elsewhere, and lists paginate in small batches, so no further scroll-performance changes were needed.

## User stories
As a customer, after my case is resolved or a solution is accepted, I can rate my support experience and leave an optional comment. As a customer, scrolling through code blocks, tables, and the project switcher feels smooth on my phone instead of janky.

## Release note
Added case feedback submission to the Customer Portal mobile app, and smoothed touch scrolling in a few content areas.

## Documentation
N/A — internal customer portal UI change, no external doc surface affected.

## Automation tests
- No automated test runner is wired up for this app yet.
- Verified locally: `tsc -b` and `eslint` clean on all changed files; manually exercised the feedback flow (auto-open on resolve, standing prompt, validation, submit, read-only display after submit) and the scroll fix against a running dev server.

## Security checks

## Test environment
- Local dev server (Vite), manual browser testing.

## Summary by CodeRabbit

* **New Features**
  * Added feedback collection for closed support cases.
  * Users can select an emoji, optional feedback tags, and add comments up to 1,000 characters.
  * Existing feedback is displayed on closed case details.
  * Feedback submission includes loading, error, and success states.
* **Usability Improvements**
  * Improved touch scrolling for project lists, formatted text, and Markdown tables on WebKit browsers.

### DIFF
--- a/apps/customer-portal/microapp/src/services/cases.ts
+++ b/apps/customer-portal/microapp/src/services/cases.ts
@@ -154,13 +154,12 @@ const getFeedback = async (id: string): Promise<CaseFeedback | null> => {
 };
 
 const submitFeedback = async (id: string, body: CaseFeedbackInput): Promise<CaseFeedbackResponseDto> => {
-  return (
-    await apiClient.post<CaseFeedbackResponseDto>(CASE_FEEDBACK_ENDPOINT(id), {
-      emojiId: body.emojiId,
-      chipIds: body.chipIds,
-      additionalComment: body.additionalComment,
-    })
-  ).data;
+  const response = await apiClient.post<CaseFeedbackResponseDto>(CASE_FEEDBACK_ENDPOINT(id), {
+    emojiId: body.emojiId,
+    chipIds: body.chipIds,
+    additionalComment: body.additionalComment,
+  });
+  return response.data;
 };
 
 /* Mappers */
--- a/apps/customer-portal/microapp/src/services/cases.ts
+++ b/apps/customer-portal/microapp/src/services/cases.ts
@@ -14,6 +14,7 @@
 // specific language governing permissions and limitations
 // under the License.
 
+import axios from "axios";
 import apiClient from "@src/services/apiClient";
 import { infiniteQueryOptions, mutationOptions, queryOptions, skipToken } from "@tanstack/react-query";
 import type {
@@ -41,6 +42,10 @@ import type {
   AttachmentDto,
   Pagination,
   CreateAttachmentRequestDto,
+  CaseFeedback,
+  CaseFeedbackInput,
+  CaseFeedbackDto,
+  CaseFeedbackResponseDto,
 } from "@src/types";
 
 import {
@@ -49,6 +54,7 @@ import {
   CASE_CLASSIFICATION_ENDPOINT,
   CASE_COMMENTS_ENDPOINT,
   CASE_DETAILS_ENDPOINT,
+  CASE_FEEDBACK_ENDPOINT,
   CASE_STATS_ENDPOINT,
   CREATE_CASE_ENDPOINT,
   PROJECT_CASES_ENDPOINT,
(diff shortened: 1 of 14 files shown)


apps/customer-portal/microapp/src/services/cases.ts
```diff
@@ -137,6 +143,26 @@ const createAttachment = async (id: string, body: CreateAttachmentRequestDto): P
   await apiClient.post<CommentDto>(CASE_ATTACHMENTS_ENDPOINT(id), body);
 };
 
+const getFeedback = async (id: string): Promise<CaseFeedback | null> => {
+  try {
```
v15a1, line 147: Other service functions don't have a try/catch. Perhaps any errors are caught in the UI, check that and remove the try/catch here.

apps/customer-portal/microapp/src/services/cases.ts
```diff
@@ -137,6 +143,26 @@ const createAttachment = async (id: string, body: CreateAttachmentRequestDto): P
...
+};
+
+const submitFeedback = async (id: string, body: CaseFeedbackInput): Promise<CaseFeedbackResponseDto> => {
+  return (
+    await apiClient.post<CaseFeedbackResponseDto>(CASE_FEEDBACK_ENDPOINT(id), {
+      emojiId: body.emojiId,
+      chipIds: body.chipIds,
+      additionalComment: body.additionalComment,
+    })
+  ).data;
```
v15a1:
```suggestion
    const response = await apiClient.post<CaseFeedbackResponseDto>(CASE_FEEDBACK_ENDPOINT(id), {
      emojiId: body.emojiId,
      chipIds: body.chipIds,
      additionalComment: body.additionalComment,
    })
    return response.data;
```

v15a1 (changes requested):

apps/customer-portal/microapp/src/services/cases.ts
Hesara2003: Applied this. It now matches editCase's pattern in this file. Thanks!

apps/customer-portal/microapp/src/services/cases.ts
Hesara2003: Keeping this on purpose. A 404 here just means the case has no feedback yet, not an error. Most cases are in that state until someone submits feedback, so it needs to return null instead of throwing. The other service functions do not need this because none of them have a normal not found case. The webapp's useGetCaseFeedback does the same thing for the same reason. Without the try/catch, every case without feedback would show as an error instead of the normal empty state.

v15a1 (approved):

PR #1186, opened 2026-07-21
[CSM Portal Microapp] Add the announcement detail view

Hesara2003:
## Purpose
Announcement cards in the CSM microapp weren't tappable — there was no way to open one to read its full content, even though the webapp added this a while back.

## Goals
- Let an agent tap an announcement to see its full detail (description/activity, attachments) on mobile.
- Don't expose actions that don't apply to announcements: state transitions, severity changes, commenting, SLAs, time tracking, call requests.

## Approach
- The webapp doesn't have a dedicated announcement detail page — `announcements/:caseId` just routes to the same `CsmCaseDetailPage` used for regular cases and engagements, gated down by `caseType === "announcement"`. Mirrored that here rather than building a new page.
- `AnnouncementCard` is now a `Link` to the existing `/cases/:id` route (the same `CaseCard` pattern already used for the main case list) — no new route needed, since this microapp doesn't do the webapp's per-type route aliasing.
- `CaseDetailPage` now computes `isAnnouncement` from the loaded case's own `type` and hides what doesn't apply: status/severity chips, the action bar, the severity-change control, the comment composer, and the SLA/Time tracking/Call requests tabs. Activities (read-only), Details, and Attachments stay visible. Simpler than the webapp's version, which also needs a route-based check (`/announcements/` vs `/cases/`) since it exposes both URL shapes — this page is always reached via `/cases/:id` regardless of case type, so the loaded case's own type is the only signal needed.

## User stories
As an agent, I can tap an announcement from the list and read its full content and attachments on my phone, without seeing case actions that don't apply to it.

## Release note
Announcements are now tappable in the CSM Portal mobile app, opening a read-only detail view.

## Documentation
N/A — internal CSM portal UI change, no external doc surface affected.

## Automation tests
- No automated test runner is wired up for this app yet.
- Verified locally: `tsc -b`, `eslint`, and `npm run build` all clean.

## Security checks

## Test environment
- Local dev server (Vite), manual browser testing.

## Summary by CodeRabbit

* **New Features**
  * Announcement cards are now clickable and navigate to the related case.
  * Announcement case pages open in read-only mode, focusing on the activities feed.

* **Bug Fixes**
  * Announcement pages now hide SLA/time tracking and call-request tabs and fall back to “activities” when needed.
  * Comment composing and case actions (status/severity changes) are disabled for announcements.
  * Improved announcement search pagination by deriving “has more” when it’s not provided by the backend.

### DIFF
--- a/apps/csm-portal/microapp/src/components/announcements/AnnouncementCard.tsx
+++ b/apps/csm-portal/microapp/src/components/announcements/AnnouncementCard.tsx
@@ -15,16 +15,23 @@
 // under the License.
 
 import { Card, Skeleton, Stack, Typography } from "@wso2/oxygen-ui";
+import { Link } from "react-router-dom";
 import type { CaseSummary } from "@src/types";
 import { fromNow } from "@utils/dateTime";
 import { AnnouncementStateChip } from "./AnnouncementStateChip";
 
-// Read-only announcement card (not tappable — mirrors the webapp's read-only
-// list). Announcements are cases of type "announcement", so `item` is a
-// `CaseSummary`; only the fields the list shows are rendered.
+// Announcements are cases of type "announcement", so `item` is a
+// `CaseSummary`; only the fields the list shows are rendered. Taps through to
+// the shared case detail page (/cases/:id), same as the webapp's
+// CsmCaseDetailPage reuse — it renders a read-only view for this case type.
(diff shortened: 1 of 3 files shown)


s-sandali:
@coderabbitai review

SSSD-2001:
@coderabbitai review

cloby99 (approved):

PR #1188, opened 2026-07-21
[CSM Portal] Prefill Post Resolution Activity dialog; align resolution-code/cause labels and cause encoding with the backing data source

rksk:
## Purpose
The CSM portal's "Close case" / "Propose a solution" dialog (`ResolutionDialog`) always opened blank, even when a case already carried a prior resolution code, cause, and notes (e.g. a reopened case). Separately, its two dropdowns' displayed text was mechanically generated from the backend enum tokens, which doesn't match the backing data source's real picklist wording, and the `cause` field's backend enum/label set didn't correspond to any real choice in that picklist at all — sending it would not have round-tripped correctly in production.

## Goals
- Prefill the dialog from the case's existing resolution data instead of opening blank.
- Show the exact picklist wording for both dropdowns instead of a mechanically title-cased approximation.
- Make the `cause` field's backend enum and wire encoding actually match the backing data source's real "cause" choice list (verified live against the field's picklist definition), so the value written back is meaningful.

## Approach
- **FE (CSM webapp):** `useGetCsmCaseDetail` now maps the case's existing `resolutionCode`/`cause`/`resolutionNotes` (already returned by `GET /cases/{id}`, previously dropped by the FE type) onto a new `resolution` field; `ResolutionDialog` accepts an `initial` prop to seed its fields when opening. Added explicit label maps (`RESOLUTION_CODE_LABELS`, `CASE_CAUSE_LABELS`) with the verbatim source wording for both dropdowns, replacing mechanical title-casing (which can't reproduce the source's punctuation, capitalization, or acronyms like `IAAS`/`JDK`/`LDAP`).
- **Entity-service (Go):** `domain.CaseCause` is replaced with a 25-value enum matching the real "cause" picklist (previously 16 fabricated values). The backing data source encodes this field as a plain sequential integer (1-25) in production rather than a label string — verified live against the field's picklist definition — so the mapping follows the same pattern already used for `resolutionCode` (domain enum <-> integer, sent as the string form since the downstream integration payload's cause field stays a plain string). Resolution codes were already correct end-to-end and are untouched.
- **csm-portal-backend:** `openapi.yaml` enum updated to match (docs only — this service passes the request body through unvalidated).

## User stories
As a CS engineer reopening a previously-resolved case, I see its prior resolution code, cause, and notes prefilled when I close it or propose a solution again, instead of having to re-enter them from scratch. As a CS engineer picking a resolution code or cause, I see the backing source's actual wording in the dropdown.

## Release note
Fixed: the case resolution dialog (Close/Propose solution) now prefills from the case's existing resolution data, shows the exact picklist wording for resolution codes and causes, and writes cause values the backing data source actually recognizes.

## Documentation
N/A — internal contract fix; no customer- or partner-facing documentation covers this field's value set.

## Automation tests
- Unit tests: `ResolutionDialog.test.tsx` covers prefill-and-resubmit, and dropdown selection now targets the picklist-wording labels; `go test ./...` passes for entity-service.
- Integration tests: none added; manually verified the mapping values against the backing data source's live picklist definition for the "cause" field.

## Security checks
- Ran FindSecurityBugs plugin and verified report? N/A — Go/TS project; `go vet` and `eslint` ran clean instead

## Samples
N/A

## Related PRs
None

## Migrations (if applicable)
N/A — no schema/table changes. Note for reviewers: this changes the domain `CaseCause` enum's wire values; any already-closed/solution-proposed case whose `cause` was previously written using the old fabricated enum will not decode against the new mapping (returns `nil`/omitted rather than a wrong value) — it does not corrupt existing data.

## Test environment
Go 1.26, Node/pnpm (per `apps/csm-portal/webapp/package.json`), macOS (host, no devcontainer). Verified the cause mapping locally against the backing data source's live picklist definition.

## Learning
Traced the actual wire path end-to-end (FE -> csm-portal-backend passthrough -> entity-service -> downstream integration) before changing anything, since the backend openapi enum turned out not to reflect any real choice in the source picklist. Read the source's own picklist definition directly (read-only, via an authenticated session) rather than guessing values from a UI screenshot, and caught a real discrepancy between two environments of the backing data source in how the same field is configured before finalizing the mapping.


rksk:
@coderabbitai resume

Rashmika998 (approved):

PR #1217, opened 2026-07-23
[CSM Portal] add case tasks and tags endpoints, document new case fields in backend API

rksk:
## Purpose
Extends the CSM portal backend's case API contract to match the companion entity-service PR in this repo, so the CSM webapp can consume: case watchers, auto-closure hold, editable case subject/description/deployment/deployed-product, child/related case linking, case tags, and case task create/update.

## Goals
- Document all new case read fields and PATCH variants in this service's OpenAPI contract.
- Add real new routes for case task create/update and case tag add/remove, following this backend's existing thin-passthrough handler pattern.
- Leave every other new field/PATCH variant as pure passthrough, matching how this backend already forwards the bulk of the case-update contract without modeling it server-side.

## Approach
This backend forwards case reads and most case-update fields to the entity-service verbatim, so most of this PR is OpenAPI documentation plus forwarding tests proving the new fields pass through unchanged. Two items needed real new code because they're genuinely new sub-resources rather than fields on the existing case object: case tasks (new `POST /cases/{id}/tasks` and `PATCH /tasks/{id}`) and case tags (new `POST /cases/{id}/tags` and `DELETE /cases/{id}/tags/{tagId}`), each following this backend's existing auth/validation/error-mapping conventions.

## User stories
As the CSM webapp, I need a documented, testable backend contract for the new case capabilities so the frontend can build against a stable API surface.

## Release note
Adds case watchers, auto-closure hold, editable case details, case linking, case tags, and case task endpoints to the CSM portal backend API.

## Documentation
N/A — documented in this PR's own `openapi.yaml` changes; no external product docs affected.

## Automation tests
- Unit tests: added for every new route (auth/UUID/body-size/JSON guards, verbatim forwarding, error mapping) and for every passthrough field/variant.
- Integration tests: N/A, none exist in this package beyond the unit level.

## Security checks
- Ran FindSecurityBugs plugin and verified report: N/A (Go codebase — `go vet` run clean as part of the build)
- Confirmed no keys/passwords/tokens/usernames/secrets committed: yes

## Samples
N/A

## Related PRs
Stacked alongside companion PRs for the entity-service and CSM webapp in this repo.

## Migrations (if applicable)
N/A

## Test environment
Go 1.x (as pinned by this repo's go.mod), macOS, run via `make build` (vet + test -race + build).

## Learning
N/A


rksk:
@coderabbitai resume

Rashmika998 (approved):

PR #1226, opened 2026-07-25
feat(csm-microapp): namespace cached auth tokens by environment

Hesara2003:
## Purpose
The microapp caches its access/ID tokens under fixed localStorage keys inside the superapp's webview. That webview storage partition is shared across the dev/staging/prod builds of the microapp, so a device that previously ran one environment's build can end up reusing a still-unexpired token cached by a different environment. This only ever surfaced as "the app doesn't work," and was only fixed by uninstalling and reinstalling the superapp to clear the stale value.

## Goals
- Prevent a token cached by one environment's build from being reused by another.
- Fail loudly (build-time env var required) rather than silently falling back to a shared key.

## Approach
- Added a new required build-time var, `VITE_APP_ENV`, read once in `constants.ts` and appended as a suffix to both `LocalStorageKeys` entries (`csm_portal_accessToken_<env>` / `csm_portal_idToken_<env>`), mirroring the existing per-microapp key prefix that was added to prevent CSM Portal/Customer Portal token collisions.
- Deliberately did not derive `<env>` by parsing `VITE_BACKEND_URL`'s hostname — the current staging URL contains the literal substring `prod.` (Choreo's own proxy domain), so string-matching the URL would misclassify staging as prod.
- Typed the new var in `vite-env.d.ts` and documented it in `.env.example`.

> **Scope note:** the dev/staging/prod build pipeline needs to actually set `VITE_APP_ENV` per target for this to take effect. That's outside this diff.

## User stories
As a CSM Portal microapp user, my cached session from one environment's build never gets reused by a different environment's build.

## Release note
Fix stale cross-environment token reuse in the CSM Portal microapp by namespacing cached tokens with a per-build `VITE_APP_ENV`.

## Documentation
N/A — internal CSM portal microapp change, no external doc surface affected.

## Automation tests
- No test runner wired up yet for the microapp.
- Verified `npm run build` and `npm run lint` pass locally.

## Security checks

## Test environment
- `npm run lint`, `npm run build` passing locally.

## Summary by CodeRabbit

* **Bug Fixes**
  * Scoped authentication token storage by build environment to prevent accidental cross-environment reuse.
* **Chores**
  * Added a `VITE_APP_ENV` configuration option and updated the app to derive environment-scoped token keys from it (with a `prod` fallback).
  * Extended typed `import.meta.env` support to include `VITE_APP_ENV`.

### DIFF
--- a/apps/csm-portal/microapp/src/utils/constants.ts
+++ b/apps/csm-portal/microapp/src/utils/constants.ts
@@ -18,6 +18,20 @@ export const ErrorMessages = {
   NATIVE_BRIDGE_NOT_AVAILABLE: "Native bridge is not available",
 };
 
+// Same superapp webview also shares its storage partition across dev/staging/prod
+// builds of this microapp (e.g. a device that had the staging build installed, then
+// gets the prod build). VITE_APP_ENV must be set distinctly per build so a token
+// cached under one environment is never reused by another — parsing the env out of
+// VITE_BACKEND_URL instead isn't safe, since Choreo's own proxy domain contains
+// "prod." even for staging URLs. Without this, refreshToken()'s isTokenExpiringSoon()
+// short-circuit could reuse a still-unexpired token issued by a different
+// environment's IdP, which only surfaced as "the app doesn't work" until the
+// superapp was uninstalled and reinstalled to clear the stale cached value.
+const APP_ENV = import.meta.env.VITE_APP_ENV;
+if (!APP_ENV) {
+  throw new Error("VITE_APP_ENV is not defined");
+}
+
 // Prefixed per-app so a token cached here can never be picked up by a sibling
 // microapp (e.g. customer-portal) sharing the same storage partition. Each
 // microapp is registered as its own OAuth client in the superapp, so the two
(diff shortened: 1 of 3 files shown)


apps/csm-portal/microapp/src/utils/constants.ts
```diff
@@ -18,6 +18,20 @@ export const ErrorMessages = {
...
+// builds of this microapp (e.g. a device that had the staging build installed, then
+// gets the prod build). VITE_APP_ENV must be set distinctly per build so a token
+// cached under one environment is never reused by another — parsing the env out of
+// VITE_BACKEND_URL instead isn't safe, since Choreo's own proxy domain contains
+// "prod." even for staging URLs. Without this, refreshToken()'s isTokenExpiringSoon()
+// short-circuit could reuse a still-unexpired token issued by a different
+// environment's IdP, which only surfaced as "the app doesn't work" until the
+// superapp was uninstalled and reinstalled to clear the stale cached value.
+const APP_ENV = import.meta.env.VITE_APP_ENV;
+if (!APP_ENV) {
```
Rashmika998: If there's no env config provided, let's use prod one by default

rksk (approved):

PR #1239, opened 2026-07-27
[Entity-Service] send the auto-closure hold as a date, not a datetime

rksk:
## Purpose
The case auto-closure hold (`autocloseHoldUntil`) carries a date with no time component, but the case-update path formatted it with the shared datetime layout. The integration service rejected the payload before it reached the backing data source, so the field could never be set (staging report: `Hold auto closure` returns 400).

## Goals
- Let `autocloseHoldUntil` writes succeed against the backing data source.

## Approach
Adds `snDateOnlyLayout`/`formatSNDateOnly` and uses it for `autocloseHoldUntil` only; every other date field on this path stays on the existing datetime layout. Updates the one test that asserted the old datetime layout for this field.

## Release note
Fixes case auto-closure hold, which previously failed with a 400 on every attempt.

## Documentation
N/A — internal date-formatting fix, no contract/shape change.

## Security checks
- FindSecurityBugs / static analysis: N/A for Go — `go vet ./...` clean
- No secrets committed: yes

## Related PRs
N/A

### DIFF
--- a/entity-service/internal/service/sn_case_service.go
+++ b/entity-service/internal/service/sn_case_service.go
@@ -370,6 +370,20 @@ func formatSNDate(t *time.Time) string {
 	return t.UTC().Format(snCreatedOnLayout)
 }
 
+// snDateOnlyLayout is the date-only wire format the integration service expects for
+// fields whose contract carries a date with no time component.
+const snDateOnlyLayout = "2006-01-02"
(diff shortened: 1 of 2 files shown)


rksk:
@coderabbitai resume

rksk:

PR #1243, opened 2026-07-27
[Entity-Service] convert incident watchlist UUIDs to ServiceNow sysids

rksk:
Incident watchlist PATCH/POST requests fail on staging with a 404 ("User not found"). `CreateIncident` and `UpdateIncident` in `entity-service/internal/service/sn_incident_service.go` convert every other reference-ID field (`parentId`, `assignmentGroupId`, `assignedEngineerId`, `serviceId`, etc.) from the platform's UUID to the backing data source's identifier format before building the outgoing payload, but `watchList` was passed through unconverted. The backing data source expects each watchlist entry to resolve to a user record identifier, so a raw platform UUID fails to resolve and the write is rejected.

- Incident create/update requests with a `watchList` succeed against the backing data source instead of 404ing.
- A malformed UUID in `watchList` is rejected with a clean 400 validation error before any outbound call, matching the behavior of every other UUID field on this resource.

- In `CreateIncident`, run `req.WatchList` through the existing `uuidsToSysids()` helper (already used for other multi-value UUID fields such as filter ID lists) before assigning it to the outgoing payload, and validate each entry with the existing `validateUUIDs()` helper alongside the other required/optional UUID checks.
- In `UpdateIncident`, do the same for the optional `*[]string` `WatchList` field: validate with `validateUUIDs()` when non-nil, and convert with `uuidsToSysids()` before assigning to the payload.
- No API contract change: the field shape and JSON key are unchanged, only the outgoing value is now correctly translated.

## User stories
N/A — internal bug fix, no new user-facing behavior beyond the fix.

## Release note
Fixed a bug where adding or updating watchers on an incident could fail because watcher identifiers were not translated to the format the backing data source expects.

## Documentation
N/A — internal service-layer bug fix, no user-facing or API contract change to document.

## Training

## Certification
N/A — no certification content impact.

## Marketing
N/A — internal bug fix, no marketing impact.

## Automation tests
- Unit tests
  Added `entity-service/internal/service/sn_incident_service_test.go` covering: (1) `CreateIncident` and `UpdateIncident` convert every `watchList` UUID to the backing identifier format in the outgoing payload, and (2) an invalid UUID in `watchList` is rejected by validation before any outbound call, for both create and update.
- Integration tests
  N/A — covered by the above unit tests against a mocked backing-service client; no separate integration suite for this service.

## Security checks
- Ran FindSecurityBugs plugin and verified report? N/A (Go codebase; ran `go vet ./...`, clean)

## Samples
N/A — no sample content impacted.

## Related PRs
N/A

## Migrations (if applicable)

## Test environment
Go 1.x (per `entity-service/go.mod`), macOS, local `go build`/`go vet`/`go test` runs.

## Learning
Root cause was confirmed via staging logs prior to this change: other reference-ID fields on this resource already went through a UUID-to-backing-identifier conversion helper, but `watchList` was missed when it was added, so it was passed through raw.

### DIFF
--- a/entity-service/internal/service/sn_incident_service.go
+++ b/entity-service/internal/service/sn_incident_service.go
@@ -527,6 +527,9 @@ func (s *snIncidentService) CreateIncident(ctx context.Context, req domain.Creat
 			}
 		}
 	}
+	if err := validateUUIDs("watchList", req.WatchList); err != nil {
+		return domain.CreateIncidentResponse{}, err
+	}
 
 	token := middleware.UserIDTokenFromContext(ctx)
 
@@ -537,7 +540,7 @@ func (s *snIncidentService) CreateIncident(ctx context.Context, req domain.Creat
 		ImpactKey:          snIncidentImpactKeyMap[req.Impact],
 		UrgencyKey:         snIncidentUrgencyKeyMap[req.Urgency],
 		Subject:            req.Subject,
-		WatchList:          req.WatchList,
+		WatchList:          uuidsToSysids(req.WatchList),
 		AdditionalComments: req.AdditionalComments,
 		WorkNotes:          req.WorkNotes,
 	}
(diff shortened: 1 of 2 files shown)


rksk:

PR #1249, opened 2026-07-27
[CSM Portal] fix: forward real upstream message on change-request 409/422 errors

rksk:
## Purpose
The csm-portal backend's upstream-error mapper (`mapUpstreamError`) already had a pre-existing branch for HTTP 409/Conflict and 422/Unprocessable Entity responses from the backing data source (general infrastructure, not specific to any one endpoint), intended to relay the real rejection reason to the caller. However, it passed the raw upstream response body straight through as the outer envelope's `message` field. The backing data source's error responses are themselves JSON objects (`{"code":409,"message":"..."}`), so this double-encoded the whole JSON blob into an escaped string instead of the plain-text message callers expect, e.g.:

```
{"message":"{\"code\":409,\"message\":\"State transition rejected\"}"}
```

instead of:

```
{"message":"State transition rejected"}
```

This surfaced while verifying a change-request `requestApproval` rejection (an illegal state transition), which returns 409 — this PR fixes the shared message-extraction for both codes already handled by this branch, it does not add 422 handling that didn't exist before.

## Goals
Forward the actual human-readable rejection message on 409/422 upstream errors, without breaking the case where the upstream body is already plain text.

## Approach
Added `upstreamErrorMessage`, which parses the upstream body as `{"message": "..."}` and returns the extracted message when present, falls back to the raw body when the body isn't a JSON object with a message field (preserves prior plain-text behavior), and falls back to the handler's existing fallback message when the body is empty. `mapUpstreamError`'s existing 409/422 case now calls this helper instead of forwarding the raw body directly. No status-code branches were added or removed — only the message extraction inside the existing branch changed.

## User stories
As a CS engineer, when I try to request approval on a change request that isn't in a state that allows it, I see the real rejection reason instead of a generic/garbled error.

## Release note
Fixed: change-request approval requests (and any other 409/422 response from the backing data source) now surface the real rejection message instead of an escaped JSON blob.

## Documentation
N/A — internal error-mapping bug fix, no user-facing API contract change (response shape is unchanged: `{"message": "<text>"}`).

## Automation tests
- Unit tests: extended the shared `upstreamErrors` table (`internal/handler/cases_test.go`, exercised by every handler that calls `mapUpstreamError`) with cases for a JSON-envelope upstream body, a plain-text upstream body (existing behavior, unchanged), and an empty body. `make test` (vet + `go test -race ./...`) passes.
- Integration tests: N/A — no upstream integration harness in this component.

## Security checks
- Ran FindSecurityBugs plugin and verified report? N/A (Go component; `go vet ./...` run clean as part of `make build`)

## Samples
N/A

## Related PRs
N/A

## Migrations (if applicable)
N/A

## Test environment
Go (per `go.mod`), macOS, `make build` (vet + `go test -race ./...` + `go build`) run locally.

## Learning
N/A


rksk:
@coderabbitai resume

rksk:
@coderabbitai resume

Rashmika998 (approved):

rksk:
@coderabbitai resume

rksk:
@coderabbitai resume

PR #1252, opened 2026-07-27
[Entity-Service] bug fixes: auto-closure date format, incident watchlist sysid conversion

rksk:
## Purpose
Consolidates two independent, unrelated-but-small entity-service bug fixes into one PR (each was originally opened separately as #1239/#1243; those are closed in favor of this one) to reduce review overhead. Each commit is self-contained and independently revertable.

## Goals
- Fix the case auto-closure hold field, which failed with a 400 on every attempt (datetime vs. date-only format mismatch).
- Fix incident watchlist updates, which failed with a 404 because platform UUIDs were sent to the backing data source instead of being converted to its native id format.

## Approach
Two independent commits, each addressing one bug in its own files with its own tests — see individual commit messages for detail. No shared code between them; a review can evaluate each commit in isolation.

(A third fix — mapping HTTP 422 in the SN-integration client — was included and then reverted in this branch's history: the bug it was meant to unblock, a call-request rejection returning a masked 500, is instead being fixed by changing the ServiceNow scripted API to return 409, which this client already handles correctly with no code change needed. Carrying the 422 case forward with no current caller would just be unused code.)

## Release note
Fixes two staging bugs: case auto-closure hold, and incident watchlist updates.

## Documentation
N/A — internal bug fixes, no contract/shape changes beyond what each commit's own tests cover.

## Security checks
- FindSecurityBugs / static analysis: N/A for Go — `go vet ./...` clean
- No secrets committed: yes

## Related PRs
Supersedes #1239, #1243 (closed in favor of this consolidated PR).


rksk:
@coderabbitai resume

Rashmika998 (approved):

rksk:
@coderabbitai resume

rksk:
@coderabbitai resume

PR #1260, opened 2026-07-27
[CSM Portal] don't offer Request approval when the change request has no assigned team

rksk:
## Purpose
On a change request's detail page, clicking "Request approval" could still round-trip all the way to the backing data source and fail there, even though the frontend already has a client-side gate. The existing gate only checked the generic state-machine transition list (`legalNextStates`), which reflects the structural New -> Assess move being legal — but the backing data source separately enforces a precondition not captured by that list: the change request must also have an assignment group/team set, or the transition is rejected server-side.

## Goals
Extend the client-side gate so a CS engineer can't trigger a request that is guaranteed to fail: require an assigned team in addition to the existing state check before treating "Request approval" as available.

## Approach
- `canRequestApproval`-equivalent logic now requires `cr.assignedTeam` to be set in addition to `legalNextStates` including `"assess"`.
- When the transition is otherwise legal but no team is assigned, the button is now shown disabled with a tooltip ("Set an assigned team before requesting approval") instead of being silently hidden, mirroring the existing disabled+tooltip pattern used elsewhere in this app (case detail's action bar) for actions blocked by a nameable, specific reason.
- Updated/added unit tests covering: button enabled (state ok + team set), button disabled (state ok + no team, click is a no-op), and existing hide/show-by-state cases.

## User stories
As a CS engineer, when I open a change request that's eligible for approval but has no assigned team yet, I see why the Request approval action isn't available instead of hitting a failed request after clicking it.

## Release note
Fix: the change request detail page no longer offers "Request approval" when the change request has no assigned team, preventing a request that was guaranteed to fail once it reached the backing data source. The button is now shown disabled with an explanatory tooltip in that case.

## Documentation
N/A — internal client-side validation change, no user-facing product documentation to update.

## Training
N/A — no training content is affected by this change.

## Certification
N/A — no certification exam content is affected by this change.

## Marketing
N/A — internal bug fix, no marketing impact.

## Automation tests
 - Unit tests
   > Updated `CsmChangeRequestDetailPage.test.tsx`: added a case asserting the button is disabled (and clicking it is a no-op) when the state allows the transition but no team is assigned, and a case asserting it stays enabled when both conditions are met. All 11 tests in the file pass (`npx vitest run`).
 - Integration tests
   > N/A — no integration test suite covers this page; covered by the unit tests above plus a full `pnpm build`.

## Security checks
 - Ran FindSecurityBugs plugin and verified report? N/A — TypeScript/React frontend; ran `npx eslint` and `npx tsc -b` clean instead.

## Samples
N/A — no sample code affected.

## Related PRs
None.

## Migrations (if applicable)

## Test environment
Node.js + pnpm, local build via `npx tsc -b`, `npx eslint`, `npx vitest run`, and `pnpm build` (all passing).

## Learning
N/A — root cause was already confirmed prior to this change; followed the existing disabled+tooltip pattern already established in this codebase (case detail action bar) rather than inventing a new one.

### DIFF
--- a/apps/csm-portal/webapp/src/features/csm-operations/pages/CsmChangeRequestDetailPage.test.tsx
+++ b/apps/csm-portal/webapp/src/features/csm-operations/pages/CsmChangeRequestDetailPage.test.tsx
@@ -88,6 +88,7 @@ const BASE_CR: BeChangeRequestDetail = {
   createdOn: "2026-01-01T00:00:00Z",
   state: "new",
   type: "normal",
+  assignedTeam: { id: "team-1", name: "Platform" },
 };
 
 function mockQueryResult(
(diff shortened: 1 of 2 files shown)


rksk:
Consolidated into #1264 to simplify review — no code changes, same commit content.

PR #1261, opened 2026-07-27
[CSM Portal] give case watchers their own tab, separate from Related

rksk:
The case-detail "Related" tab currently combines two unrelated concerns in one screen: the watch list (who gets notified on case updates) and child cases. Watchers aren't actually "related case" content, and cramming them into the same tab as child cases makes both harder to scan and prevents showing an unambiguous watcher count on the tab bar.

Give Watchers its own top-level tab, matching the existing pattern used by the SLAs tab, so it's independently discoverable and can carry its own count in the tab label.

In `CsmCaseDetailPage.tsx`:
- Added a new `"watchers"` tab id, positioned between "Related" and "SLAs" in the tab bar, with an `Eye` icon (already used elsewhere in this webapp for view/visibility affordances) to keep it visually distinct from the other icons on this tab bar.
- Moved the `WatchersWidget` render out of the "Related" tab's content block into its own `{activeTab === "watchers" && ...}` block. The widget itself, its inline add/remove chip behavior, and the `onAddWatcher`/`onRemoveWatcher` handlers are unchanged — only relocated.
- The "Related" tab now renders only `ChildCasesWidget` and switches from the previous 2-column grid to a single-column layout, since it's no longer combining two widgets.
- Added a watcher count to the new tab's label (`c.watchers.length`), consistent with how SLAs/Attachments/Time/Call requests/Tasks already show counts. "Related" still shows no count (its widget runs its own scoped data fetch, so no count is available at this level without an extra query).
- The action bar's "Manage watchers" item now jumps to the new "watchers" tab instead of "related".
- Added "watchers" to the same announcement-only tab-hiding logic that already excludes "related"/"sla"/"time"/"call-requests" (announcements have no watch list), in both places that gate: the tab-bar filter and the render-time fallback-to-Activities guard.

No changes to `WatchersWidget`, `WatcherAddPicker`, or any backend/entity-service code.

As a CS engineer viewing a case, I want Watchers to have its own tab, so I can find and manage the watch list without it being mixed in with child-case information.

Case detail: Watchers now has its own tab, separate from Related.

N/A — internal CSM-portal UI reorganization, no external-facing product documentation covers this screen.

## Automation tests
 - Unit tests
   > Code coverage information

Existing `CaseDetailWidgets.test.tsx` (WatchersWidget unit tests) and `CaseActionBar.test.tsx` (secondary-action dispatch tests, including "manage watchers") pass unchanged — the tab move doesn't affect either component's own test surface. Manually verified `tsc -b`, `eslint`, and `pnpm build` are clean.

## Security checks
 - Ran FindSecurityBugs plugin and verified report? N/A (TypeScript/React change; ran `eslint` clean instead)

### DIFF
--- a/apps/csm-portal/webapp/src/features/csm-cases/pages/CsmCaseDetailPage.tsx
+++ b/apps/csm-portal/webapp/src/features/csm-cases/pages/CsmCaseDetailPage.tsx
@@ -35,6 +35,7 @@ import {
   ArrowLeft,
   CheckSquare,
   Clock,
+  Eye,
   Layers,
   Link as LinkIcon,
   ListChecks,
@@ -252,6 +253,7 @@ type CaseTabId =
   | "activities"
   | "details"
   | "related"
+  | "watchers"
   | "sla"
   | "attachments"
   | "time"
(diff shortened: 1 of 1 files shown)


rksk:
Consolidated into #1264 to simplify review — no code changes, same commit content.

PR #1268, opened 2026-07-27
Give attachment uploads their own size cap instead of the generic 1 MiB default

rksk:
## Purpose
Attachment uploads fail with `400 "request body too large"` for any file over ~1 MiB, including small legitimate files well under every advertised limit (client-side check and downstream service cap both allow far larger uploads). Reported by QA on 2026-07-12 (2 MB attachment repro), retested and confirmed still broken on 2026-07-14.

Root cause: `CreateCaseAttachment` in `entity-service/internal/handler/case_handler.go` decoded its body through the generic `decodeRequest` helper, which applies a blanket 1 MiB cap (`maxRequestBodySize`) intended for small structured JSON payloads to every endpoint, including this one. The attachment payload is a base64-encoded file plus JSON envelope, which is legitimately much larger than 1 MiB for ordinary attachments.

## Goals
Give attachment uploads their own size cap, sized to actually accommodate the file sizes the rest of the stack already allows, without loosening the cap for any other endpoint.

## Approach
- Added `decodeRequestWithLimit`, a generic variant of the existing `decodeRequest` helper that accepts a caller-supplied byte limit and too-large message, instead of always applying the package-wide default.
- `CreateCaseAttachment` now calls `decodeRequestWithLimit` with a new `maxAttachmentBodySize` constant of 15 MiB, matching the size ceiling already used one layer up in the request path for this same payload (a 10 MB file becomes ~13.3 MB once base64-encoded, plus JSON envelope overhead, so 15 MiB is the intended safe ceiling for the encoded body).
- All other endpoints continue to use `decodeRequest` and the unchanged 1 MiB default — this is a scoped override for one handler, not a global change.
- The body-too-large error message for this endpoint now states the actual limit (10 MB, the underlying file-size ceiling) instead of the generic "request body too large" string, so any bypass of the client-side check surfaces something concrete.

## User stories
As a CS engineer, I can upload a legitimate attachment to a case (e.g. a 2 MB screenshot or log file) without it being rejected for being "too large" when it isn't.

## Release note
Fixed a bug where attaching a file over ~1 MiB to a case failed with a generic "request body too large" error, even for small files well under the platform's advertised upload limits.

## Documentation
N/A — internal API behavior fix, no user-facing documentation to update.

## Training
N/A — not training content.

## Certification
N/A — no certification impact.

## Marketing
N/A — bug fix, not a feature.

## Automation tests
- Unit tests
  - Added `entity-service/internal/handler/case_handler_test.go` covering `CreateCaseAttachment`:
    - a body over the old 1 MiB cap but under the new 15 MiB cap now reaches the service layer and returns 201 (previously rejected at decode).
    - a body over the new 15 MiB cap is still rejected, with the attachment-specific error message.
  - Full suite: `go build ./...`, `go vet ./...`, `go test ./...` all pass.
- Integration tests
  - None added; this is a handler-level fix isolated with a stub service implementation.

## Security checks
- Ran FindSecurityBugs plugin and verified report? N/A (Go component) — ran `go vet ./...` clean instead.

## Samples
N/A

## Related PRs
None

## Migrations (if applicable)

## Test environment
Local Go toolchain (go 1.26.3), macOS.

## Learning
N/A

### DIFF
--- a/entity-service/internal/handler/case_handler.go
+++ b/entity-service/internal/handler/case_handler.go
@@ -28,6 +28,13 @@ import (
 	"github.com/wso2-open-operations/cs-tools/entity-service/internal/service"
 )
 
+// maxAttachmentBodySize caps the CreateCaseAttachment JSON body at 15 MiB,
+// matching the csm-portal backend's own maxAttachmentBodyBytes ceiling: a 10
+// MB file becomes ~13.3 MB once base64-encoded, plus JSON envelope overhead.
+// The generic maxRequestBodySize (1 MiB) is far too small for this endpoint
+// and rejects legitimate small attachments (e.g. a 2 MB file).
+const maxAttachmentBodySize = int64(15 << 20)
+
 // safeAttachmentTypes is the allowlist of Content-Type values that may be
 // forwarded as-is. Anything not in this set is coerced to application/octet-stream
 // to prevent a stored-XSS attack via a crafted upstream Content-Type (e.g. text/html).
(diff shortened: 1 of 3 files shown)


rksk:
Consolidated into #1267 — this repo only tracks one open PR for CSM work at a time right now. Same commit content, no changes.

PR #1269, opened 2026-07-27
[Entity-Service] add SF ID to ServiceNow-backed account mapping

rksk:
The CSM portal's Account pages show a blank SF ID even though the backing
data source already provides it and it is well populated (a large majority
of sampled accounts carry a real value). The gap was upstream of this
service: the backing data source's account-summary mapping never forwarded
that field at all, so this service's account types had nothing to carry it
even though the shape supports it.

Add the field the backing data source now provides so it flows through
`GET /accounts/{id}` and `POST /accounts/search` to CSM portal consumers.

- `internal/service/sn_account_service.go`: added `SfID *string \`json:"sfId"\`` to
  the `snAccount` wire struct (matching the existing `pod`/`region` nullable-pointer
  convention), wired into both `snAccountToDomain` and `snAccountToDetail` via the
  existing `nilIfEmpty` helper.
- `internal/domain/entity.go`: added `SfID *string \`json:"sfId"\`` to
  `SNAccountView` and `SNAccountDetail`, next to `Pod`, mirroring the naming
  convention of the existing Postgres-path `Account.SfID string \`json:"sfId"\`` field
  (nullable here since it's sourced from the backing data source rather than
  Postgres).
- `openapi.yaml`: documented `sfId` on both `SNAccountView` and `SNAccountDetail`.

No handler/repository changes were needed — both mapping functions already
pass every other pointer field straight through the same way.

As a CS engineer, I want to see an account's SF ID on its detail/search
views so I can cross-reference it against Salesforce.

Account detail and search responses now include `sfId` when the backing data
source has a linked Salesforce ID for the account.

N/A — internal API field addition documented inline in `openapi.yaml`.

## Training

## Certification

## Marketing
N/A — internal API field addition, not user-facing marketing content.

## Automation tests
 - Unit tests
   > Added `internal/service/sn_account_service_test.go`: asserts `sfId` round-trips
   from the backing data source's JSON response into `SNAccountDetail` via
   `GetAccountByID` (both when present and when absent, confirming `nil` rather
   than an empty string). `go build ./...`, `go vet ./...`, and `go test ./...`
   all pass in `entity-service/`.
 - Integration tests
   > None added; covered by the unit test above plus manual verification against
   the backing data source directly (upstream change).

## Security checks
 - Ran FindSecurityBugs plugin and verified report? N/A — Go codebase; `go vet ./...` ran clean instead

## Samples
N/A — no sample code affected.

## Related PRs
A corresponding Ballerina entity-service change (private repo) adds the
matching field upstream so it reaches this service's account responses.

## Migrations (if applicable)
N/A — no data migration required.

## Test environment
Go 1.x (per `go.mod`), macOS, local `go test` run.

## Learning
N/A.

### DIFF
--- a/entity-service/internal/domain/entity.go
+++ b/entity-service/internal/domain/entity.go
@@ -230,6 +230,7 @@ type SNAccountView struct {
 	Name             string     `json:"name"`
 	Classification   string     `json:"classification"`
 	Pod              *string    `json:"pod"`
+	SfID             *string    `json:"sfId"`
 	Region           *string    `json:"region"`
 	SupportTier      *string    `json:"supportTier"`
 	ArrToday         *string    `json:"arrToday"`
@@ -266,6 +267,7 @@ type SNAccountDetail struct {
 	Name             string            `json:"name"`
 	Classification   string            `json:"classification"`
 	Pod              *string           `json:"pod"`
+	SfID             *string           `json:"sfId"`
 	Region           *string           `json:"region"`
 	SupportTier      *SNSupportTierRef `json:"supportTier"`
 	ArrToday         *string           `json:"arrToday"`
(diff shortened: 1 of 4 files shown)


rksk:
Consolidated into #1267 — this repo only tracks one open PR for CSM work at a time right now. Same commit content, no changes.

PR #1272, opened 2026-07-27
[CSM Portal] add 3 date-only fix-ETA estimates, replace dead fixEta field

rksk:
## Purpose
Adds Build Set Fix ETA support end-to-end: CSM engineers can record three independent, internal-only fix-commitment date estimates on a case — best case, most likely, and worst case — each a plain date (no time-of-day component).

An earlier, incomplete pass at this feature had added a single customer-facing `fixEta` field (backed by the data source's "shared with customer" fix-ETA field) alongside the three internal estimates, using full date-time values. That `fixEta` field's backing write path was never actually live, so it was dead code on every layer, and the three internal estimates were wired with the wrong date-time wire format. This PR replaces the broken `fixEta` field with the confirmed-working three-field contract and corrects the wire format to date-only.

## Goals
- Let a CSM engineer set/update best-case, most-likely, and worst-case fix-ETA dates on a case independently of one another.
- Consolidate on a single, working "fix ETA" concept instead of leaving a dead customer-facing field alongside the real internal-only one.
- Match the date-only wire contract the backing case-update resource actually expects (the same single-field-PATCH pathway already used for other case edits).

## Approach
- **entity-service**: `bestCaseFixEta` / `mostLikelyFixEta` / `worstCaseFixEta` on `UpdateCaseRequest` / `CaseView` / `UpdatedCase` are now `*string` date-only ("YYYY-MM-DD") fields instead of `*time.Time`, validated with a strict date-only parse before being forwarded. Removed the old single `fixEta` field and its now-dead read/write mapping. Fixed a gap where the Postgres-fallback path didn't reject these ServiceNow-only fields.
- **csm-portal-backend**: pure byte-passthrough handler, so only the OpenAPI spec needed updating to match (drop `fixEta`, correct the three estimates to `format: date`).
- **webapp**: `SetFixEtaDialog` now renders three date pickers (matching the existing `SetAutocloseHoldDialog` date-only picker convention) instead of four datetime pickers. Each field still saves independently via its own single-field PATCH. `CsmCaseDetailPage` now correctly passes all three current values into the dialog (previously only the unused `fixEta` value was wired) and maps `onSave(field, date)` to the matching PATCH payload variant. `CaseMetaBand` displays the three dates with a new date-only formatter so the viewer's timezone can't shift the displayed day.

## User stories
As a CSM engineer, I want to record best-case/most-likely/worst-case fix-ETA estimates on a case so the team has a shared, internal view of expected resolution timing, without exposing those estimates to the customer.

## Release note
CSM engineers can now set best-case, most-likely, and worst-case fix-ETA date estimates on a case from the "Set fix ETA…" action.

## Documentation
N/A — no customer-facing or admin-guide documentation exists for this internal CSM-portal action; the in-app dialog is self-explanatory.

## Automation tests
- Unit tests
  - entity-service: extended `sn_case_service_test.go` covering field-count validation, independent single-field PATCH per estimate, malformed-date rejection, and GET/PATCH read-back mapping for the three date-only fields.
  - csm-portal-backend: updated `cases_test.go` pass-through coverage for the three fields (date-only values) and removed the tests for the retired `fixEta` field.
  - webapp: rewrote `SetFixEtaDialog.test.tsx` for the three-field, date-only dialog.
- Integration tests
  - N/A — covered by the unit tests above; no dedicated integration suite exists for this endpoint.

## Security checks
- Ran FindSecurityBugs plugin and verified report? N/A (Go and TypeScript, not Java) — `go vet` and `eslint` both ran clean on all changed files

## Samples
N/A

## Related PRs
N/A — this is the full, self-contained change across the three components in this repo. A corresponding change to the backing entity-service's data-source integration layer, adding write support for these three date fields, is tracked separately outside this repository.

## Migrations (if applicable)
N/A — no schema/data migration involved.

## Test environment
Verified locally: Go 1.x (entity-service and csm-portal-backend — `go build`, `go vet`, `go test`/`make test` all pass), Node/pnpm (webapp — `tsc -b`, `eslint`, `vitest`, `pnpm build` all pass).

## Learning
N/A

### DIFF
--- a/apps/csm-portal/backend/internal/handler/cases_test.go
+++ b/apps/csm-portal/backend/internal/handler/cases_test.go
@@ -1993,56 +1993,13 @@ func TestRemoveCaseTag(t *testing.T) {
 	})
 }
 
-func TestPatchCaseFixEta(t *testing.T) {
-	// Item 3: fixEta is a pure pass-through single-field PATCH variant. It does not
-	// trip the state/workState peek, so patchCaseFn is invoked directly with the raw
-	// body and the upstream response passes through unchanged.
-	const testCaseID = "11111111-1111-1111-1111-111111111111"
-	const reqBody = `{"fixEta":"2026-08-01T00:00:00Z"}`
-	const upstream = `{"message":"Case updated successfully","case":{"id":"` + testCaseID + `","updatedOn":"2026-07-23T10:00:00Z","fixEta":"2026-08-01T00:00:00Z"}}`
-
-	var capturedBody []byte
-	client := &mockEntityCaseClient{
-		patchCaseFn: func(_ context.Context, _ string, body []byte) ([]byte, error) {
-			capturedBody = body
-			return []byte(upstream), nil
-		},
-	}
-	h := NewCaseHandler(client)
-	r := withUser(httptest.NewRequest(http.MethodPatch, "/cases/"+testCaseID, strings.NewReader(reqBody)))
-	r.SetPathValue("id", testCaseID)
-	w := httptest.NewRecorder()
-	h.PatchCase(w, r)
-
-	assertStatus(t, w, http.StatusOK)
-	assertContentType(t, w, "application/json")
-
-	if string(capturedBody) != reqBody {
-		t.Errorf("upstream received body %s, want %s (must forward verbatim)", capturedBody, reqBody)
-	}
-
-	var wrapper struct {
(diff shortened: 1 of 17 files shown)


rksk:
Consolidated into #1271 to keep everything on a single PR/single CodeRabbit review — same commit content, no changes.

PR #1282, opened 2026-07-28
[CSM Portal] expose project's linked account on project search results

rksk:
## Purpose
Project search results (`POST /projects/search`) had no way to identify which account a project belongs to — only the single-project detail endpoint (`GET /projects/{id}`) returned the linked account. This blocked account-scoped project filtering/display from the search path (the account detail page's project list currently has to fall back to a workaround because it can't rely on an account id being present in search results). The backing ServiceNow project-summary response now includes a compact account reference (change tracked separately); this PR carries it through the entity-service search path instead of silently dropping it.

## Goals
`POST /projects/search` results (`domain.ProjectView`) include an `account` field (`{id, name}`) mirroring what the detail endpoint already exposes, nil when the project has no linked account.

## Approach
Added `snProjectSummaryAccount` (compact `id`/`name`) to the `snProject` wire struct in `sn_project_service.go`, and `domain.ProjectView.Account *domain.EntityRef` (reusing the existing generic `EntityRef` id/name shape used elsewhere in this codebase). Mapped in `SearchProjects`: a blank account id from ServiceNow (no linked account) maps to a nil `Account` rather than a zero-valued ref.

## User stories
As a CS engineer viewing an account's page, project rows show which account they belong to without a separate lookup, and the FE can filter search results by account id.

## Release note
`POST /projects/search` results now include the project's linked account (`{id, name}`), matching what `GET /projects/{id}` already returned.

## Documentation
N/A — internal API behavior change, no external doc surface for entity-service.

## Automation tests
- Unit tests: added `TestSNProjectService_SearchProjects_MapsAccountRef` covering both a project with a linked account and one without. `go build ./...`, `go vet ./...`, `go test ./...` all pass.
- Integration tests: N/A, covered by unit tests against a mocked SN client.

## Security checks
- Ran FindSecurityBugs plugin and verified report? N/A (Go service — ran `go vet ./...` clean instead)

## Samples
N/A

## Related PRs
None

## Migrations (if applicable)
N/A

## Test environment
Go 1.x (per go.mod), tested against a mocked ServiceNow client (unit tests only); the corresponding live ServiceNow change was verified against the DEV tenant separately.

## Learning
N/A

### DIFF
--- a/entity-service/internal/domain/entity.go
+++ b/entity-service/internal/domain/entity.go
@@ -450,6 +450,8 @@ type ProjectView struct {
 	// for this project (e.g. ServiceNow leaves it blank).
 	EndDate   *time.Time `json:"endDate"`
 	CreatedOn time.Time  `json:"createdOn"`
+	// Account is nil when the project has no linked account (ServiceNow data source only).
+	Account *EntityRef `json:"account"`
 	ProjectClosureFields
 }
 
--- a/entity-service/internal/service/sn_project_service.go
+++ b/entity-service/internal/service/sn_project_service.go
@@ -51,16 +51,24 @@ type snProjectClosureFields struct {
 }
 
 type snProject struct {
-	ID        string        `json:"id"`
-	Name      string        `json:"name"`
-	Key       string        `json:"key"`
-	Type      snProjectType `json:"type"`
-	EndDate   string        `json:"endDate"`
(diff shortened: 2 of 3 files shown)


rksk:
Consolidated into #1274 (commit 5954e6b46).

PR #1294, opened 2026-07-30
[Entity Service] remove Go close-gate blocking every case close

rksk:
Closing any case is currently broken in production: `PATCH /cases/{id}
{"state":"closed"}` fails with `400 {"message":"Failed to bind request
payload to the expected schema."}` for every case, regardless of state or
task content.

Unblock case closing now, and stop reimplementing this business rule in Go
via a fragile multi-round-trip dependency.

Root cause: before allowing a close, `UpdateCase`'s close-gate calls
`hasOpenVisibleTasks`, which POSTs `/cases/{id}/tasks/search` with a
hardcoded pagination limit of 100. The entity-service's shared `Pagination`
record constrains `limit` to a `maxValue` of 50, so a limit of 100 fails
ServiceNow's own request validation at bind time -- for every case,
independent of that case's actual task count. That bind failure was
propagating up as if the close itself had failed.

Rather than patch the limit value again, this removes the Go-side
close-gate entirely. The underlying rule -- "don't close a case that still
has an open, customer-visible task" -- is a server-side authority concern.
ServiceNow's own `CaseUtils.patchCaseState` already has a directly analogous,
zero-round-trip pattern for the same class of rule (blocks closing while
open *child cases* exist, via a single in-transaction GlideRecord query).
The Go-side version required two extra network calls per close (task search
+ per-task detail fetch) purely to reconstruct visibility info SN already
has locally -- that fragility is exactly what broke production here.

This PR removes the Go-side check so case closing works again immediately.
The equivalent check moving into ServiceNow (mirroring the child-case-block
pattern) is scoped separately:
`tasks/active/2026-07-30-sn-close-gate-migration.md` in the planning repo --
that work still has open blockers (exact SN task-visibility field/table, the
target update set, and a customer-portal impact check, since
`patchCaseState` is confirmed to be on the customer portal's live path).
Until that lands, there is no server-side enforcement of the open-task rule
-- an accepted, explicit tradeoff to unblock closing now rather than leave
every case close broken while that follow-up is scoped and reviewed.

## Automation tests
 - Unit tests
   Replaced `TestSNCaseService_UpdateCase_CloseGate_BlocksOnOpenVisibleTask`
   and `TestSNCaseService_UpdateCase_CloseGate_AllowsWhenNoVisibleOpenTask`
   with `TestSNCaseService_UpdateCase_Close_NoLongerCallsTaskSearch`, which
   asserts closing a case succeeds and no longer calls
   `/cases/{id}/tasks/search` at all. Full `go test ./...` and
   `go vet ./...` are clean.

## Test environment
Go 1.x, `go test ./...` locally.

## Security checks
 - Ran FindSecurityBugs plugin and verified report: N/A (Go, not Java)
 - Confirmed no keys/passwords/tokens/secrets committed: yes

### DIFF
--- a/entity-service/internal/service/sn_case_service.go
+++ b/entity-service/internal/service/sn_case_service.go
@@ -1349,20 +1349,6 @@ func (s *snCaseService) UpdateCase(ctx context.Context, req domain.UpdateCaseReq
 		payload.PublicTicket = req.PublicTicket
 	}
 
-	// Close-gate: reject closing a case that still has an open, customer-visible task.
-	// This is the authoritative server-side check (item 1's close-gating requirement) --
-	// it only needs the existing read path (task search + task detail), so it is fully
-	// wired even though task writes themselves are still Ballerina-blocked.
-	if payload.StateKey != nil && *payload.StateKey == snStateIDMap[domain.CaseStateClosed] {
-		blocked, err := s.hasOpenVisibleTasks(ctx, uuidToSysid(req.ID), token)
-		if err != nil {
-			return domain.UpdateCaseResponse{}, err
-		}
-		if blocked {
-			return domain.UpdateCaseResponse{}, &apierror.ValidationError{Msg: "case cannot be closed while it has an open task visible to the customer"}
-		}
-	}
-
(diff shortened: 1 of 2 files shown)


rksk:
Moving this commit into #1293 instead, per feedback — closing here.

PR #1309, opened 2026-07-31
[CSM Portal] make the cases Product filter lazy/async, matching Project

Hesara2003:
## Purpose
The Cases list's Product filter paged through the *entire* product catalogue up front in an unbounded loop (no page cap at all) just to build a deduplicated set of family names, then filtered that full list client-side as the user typed — the same class of problem as the Project catalogue fetch fixed earlier for Time Cards, but worse (no safety cap on how many pages it would scan).

> **Scope note:** search-as-you-type on this filter is wired but currently non-functional — see Approach below. The dropdown's default (untyped) behavior is a strict improvement regardless.

## Goals
- Opening the Product filter dropdown loads its first page immediately, without a prior full-catalogue scan.
- Further options load lazily as the user scrolls, the same way the Project filter already behaves.
- No change to the filter's external behavior/props — same component, same call site.

## Approach
- New `useInfiniteProductSearch` (`useProductSearch.ts`), modeled directly on the existing `useInfiniteProjectSearch`: `useInfiniteQuery` over `POST /products/search`, first page on open, `fetchNextPage` wired to the listbox scroll.
- One structural difference from Project: a product search row is one product **model/version**, and many rows share the same family name — the filter operates on distinct names, not rows 1:1. The hook paginates the *raw* rows (matching the backend's actual `hasMore`/offset contract) while accumulating and deduplicating names across all fetched pages for what the UI renders — pagination count and rendered-option count are intentionally different numbers here.
- Page size is 50 (`BE_MAX_PAGE_LIMIT`), larger than Project's 10 — verified live that a real page can be dozens of rows all sharing one name, so a small page risks the first load surfacing only one or two options.
- Rewrote `ProductNameMultiSelect.tsx` in place (same name/props) to consume the new hook, copying the scroll-threshold/debounce logic already proven in `AsyncProjectMultiSelect`. Deleted the now-superseded `useProductNameOptions.ts` (no other consumers).
- **Search-as-you-type**: wired to send `searchQuery` to `/products/search`, matching how Project's search already works — but that endpoint currently rejects any request with a non-empty `searchQuery` with a schema-binding 400. The old code never actually sent `searchQuery` (it fetched everything and filtered locally), so this path was never previously exercised. This is a live, pre-existing backend issue, not something this change introduces, and it's tracked separately. Left the wiring in place rather than removing it: once fixed server-side, search starts working with no further frontend change.

## User stories
As a CS engineer, opening the Product filter shows results immediately instead of waiting on a full catalogue scan, and I can scroll to see more.

## Release note
Improved: the Cases list's Product filter no longer loads the entire product catalogue up front — it loads lazily, the same way the Project filter already does.

## Documentation

## Automation tests
- Unit tests: no existing test covers this component/hook's data-fetching directly. Full suite run locally: 569 passed, 9 pre-existing failures unrelated to this change (same set seen throughout recent CSM Portal PRs).
- Verified against staging: live network capture confirmed the first page returns real paginated data (50 rows, `hasMore: true`, `total: 367`) instead of scanning the whole catalogue; confirmed the dedup-to-family-names logic is correct against real data; confirmed (and separately reported) that the `searchQuery` parameter itself currently fails server-side.

## Security checks

## Test environment
- `pnpm run lint`, `pnpm run test`, `pnpm run build` all passing locally (via the `./node_modules/.bin/` workaround for the local `pnpm` CLI issue).

## Summary by CodeRabbit

* **New Features**
  * Product selection now supports debounced server-side search.
  * Additional product options load automatically as users scroll.
  * Selected products remain available in the dropdown, even when not in the current results.
  * Search terms are retained after selecting an option.

* **Bug Fixes**
  * Improved product loading and filtering behavior for large product lists.
  * Added clearer loading states for initial results and additional pages.


madubashinii:
@coderabbitai review

rksk (approved):

PR #1310, opened 2026-07-31
[CSM Portal] Don't route a case's parent chip to a page that 404s

rksk:
## What was wrong

A case's `parentCase` reference carries a `type` discriminator (`case` / `incident` / `change_request` / `problem`) because a parent can be any task-derived record, not just another case. Two problems around that field:

1. **The `type` contract was ambiguous.** It was documented as "nil when the backing data source doesn't resolve a type", which reads as "probably a case" and invited consumers to fall back to `/cases/{id}`. For a parent that is actually an incident or change request, that fallback lands on a 404.
2. **The parent chip acted on that fallback.** The case detail page routed an unresolved-type parent to a case detail page, producing a dead link the user cannot tell apart from a working one.

## Why it happened

The discriminator was added after the parent chip already existed, so the chip's original "every parent is a case" assumption survived as the `default:` branch of the new switch. Nothing in the contract said that was unsafe, so it did not look like a bug.

## What changed

**Contract (both `openapi.yaml` files, and the Go domain comment).** `type: null` is now stated to mean *the data source could not resolve the referenced record's kind*, with an explicit instruction that a client must not then assume it is a case. This is a documentation/semantics change only: the field, its type and its enum are unchanged.

**Entity-service read path (`case_repo.go`).** The relational data source now sets `Type: "case"` explicitly when it resolves a parent. `cases.parent_case_id` is a foreign key into `cases`, so a parent from that source is always another case; saying so removes the only case where a legitimately-known parent kind was being reported as unknown. After this, `null` really does mean "unknown" from every source.

**Frontend.** The inline `parentCasePath` helper is extracted to `features/csm-cases/utils/parentRecordRoute.ts` as `parentRecordPath`, which returns `string | null` and returns `null` for an absent or unrecognised kind instead of defaulting to `/cases/{id}`. The chip renders non-clickable with an explanatory tooltip in that case, so the parent is still visible but never a dead link. New unit tests cover all four kinds, absent `type`, explicit `null`, an unrecognised value, and an absent parent.

## Deliberate decisions a reviewer may query

**The parent reference is retyped on case *detail* only; the case *search* shape is left alone.** The new nullable-kind type cannot bind the search response shape, and changing both would break search. Detail is where the parent chip renders, so that is where the stricter type is needed. Search keeps its existing shape deliberately.

**Merge resolution note.** `main` has since gained its own parent-chip work on this same block, including forwarding the "back to list" target through `navigate(..., { state: { from: resolvedBackPath } })`. Both sides are kept: the null-safe routing from this branch *and* main's back-path forwarding. The only thing dropped is main's `default: /cases/{id}` fallback, which is precisely what this change exists to remove and what the contract text here now forbids.

## Verification

`entity-service`:

```
$ go build ./...
$ go vet ./...
$ go test ./...
ok  github.com/wso2-open-operations/cs-tools/entity-service/internal/handler   0.742s
ok  github.com/wso2-open-operations/cs-tools/entity-service/internal/service   1.052s
ok  github.com/wso2-open-operations/cs-tools/entity-service/internal/servicenow-integration-service  1.483s
```

`apps/csm-portal/webapp`:

```
$ npx tsc -b
(clean)

$ npx eslint src/features/csm-cases/pages/CsmCaseDetailPage.tsx \
    src/features/csm-cases/utils/parentRecordRoute.ts \
    src/features/csm-cases/utils/parentRecordRoute.test.ts
  680:6  warning  React Hook useEffect has a missing dependency: 'detailPath'
✖ 1 problem (0 errors, 1 warning)

$ npx vitest run
 Test Files  3 failed | 70 passed (73)
      Tests  9 failed | 579 passed (588)
```

The one eslint warning is pre-existing and on a line this change does not touch. The 9 failures are pre-existing on `main` and confined to three files unrelated to this change: `CaseActionBar.test.tsx`, `CaseActivitiesFeed.test.tsx`, `CsmAnnouncementsPage.test.tsx`.

## Not verified

**This was not exercised against a running backend.** No live request was made, so the end-to-end behaviour of a real parent reference with an unresolved kind is unverified here; the change is covered by unit tests and type checking only. A corresponding entity-service change that supplies the parent kind from the non-relational data source is released separately, and this frontend handles its absence safely in the meantime.

## Deliberately not changed

- The case search response shape and its type (see above).
- The `type` field's name, JSON shape, or enum values: only its documented meaning.


rksk:
@coderabbitai resume

rksk:
@coderabbitai resume

rksk:
@coderabbitai review

rksk:
Superseded by #1311, which now carries these commits verbatim.

Both PRs edit the same `CsmCaseDetailPage.tsx` block — this one replaces the parent-chip
routing, #1311 changes the permalink scroll/highlight and adds the linked-change-requests
widget in the same region. Reviewing that block across two PRs produced conflicting reads
of what the final file should be, so the change was merged into #1311's branch
(`a11eb31d3`, clean auto-merge, no conflicts) and re-verified there: Go build and tests
clean, `tsc --noEmit` clean, and the `csm-cases` suite showing only the pre-existing
`origin/main` baseline failures.

Nothing is lost by closing this: `csm-expose-readpath-fields` is fully contained in
`csm-portal-enhancements`, and #1311's description now carries this change's rationale and
its deliberate decisions as section 5 — including that the parent reference is retyped on
case detail only, with the case search shape left alone.

Review continues on #1311.

PR #1330, opened 2026-08-03
Add ACP Closure Service (Phase 1: subscription end-date closure)

Ananth-Abi:
## Summary

Implements Phase 1 of the ACP (Account Closure Process) migration from 
ServiceNow — a Go-based scheduled component that replaces the legacy 
ServiceNow subscription end-date closure automation.

Scope: subscription end-date closure only (90/60/30/15/7/0-day notice 
cascade, then suspend). Invoice- and compliance-based closure (Phase 2) 
are explicitly out of scope — see `docs/legacy-servicenow-reference/` 
for the original ServiceNow Script Include source this replaces.

Closes wso2-enterprise/digiops-cs#2524

## Architecture

New top-level Go module (`acp-closure-service/`), deployed as a Choreo 
Scheduled Task (run-to-completion CLI, not a long-running process). 
Calls `csm-integration-service` (not `entity-service` directly) via M2M 
client-credentials auth — see CLAUDE.md for the full architecture 
rationale.

Package structure:
- `internal/closure` — pure decision logic (days-remaining, cascading 
  notice-window thresholds)
- `internal/recipients` — three-tier customer contact fallback 
  (business-contact role → primary contact → notify Account Manager)
- `internal/entity` — API client for csm-integration-service
- `internal/suspensionstate` — translates between the real 
  `suspensionProcessState` JSON blob and internal types, preserving 
  invoice/compliance sections untouched
- `internal/notify` — notification interface (currently logs only — 
  real email-sending is pending WSO2's message-queue design)
- `internal/sweep` — orchestrates the full daily sweep, with a dry-run 
  toggle (`DRY_RUN` env var, defaults to `true`) via dependency injection

## Testing

- 34 unit/integration tests, all passing
- Verified against real staging data via direct Postman testing 
  throughout development
- Broad dry-run tested locally against all 53 real open projects in 
  staging (1 known, pre-existing data issue, unrelated to this code)
- Deployed and manually executed successfully in Choreo (Azure-Staging, 
  POC project) against the same real data, confirming identical behavior 
  to local testing

## Notes for reviewers

- `DRY_RUN` defaults to `true` — no real writes or emails happen unless 
  explicitly set to `false`
- See CLAUDE.md for detailed architecture decisions and confirmed cases 
  where live API behavior diverged from documented/coded values

## Summary by CodeRabbit

* **New Features**
  * Added an ACP Closure Service to evaluate open projects, send scheduled notices, and suspend eligible projects.
  * Supports customer, internal, and Account Manager notification paths with contact fallbacks.
  * Added dry-run mode and optional single-project execution for safer testing.
  * Added authenticated integration, configuration templates, and build/test commands.

* **Documentation**
  * Added setup, configuration, architecture, testing, and operational guidance.
  * Updated repository documentation with the new service location.

* **Tests**
  * Added coverage for closure decisions, notifications, pagination, suspension safeguards, and error handling.


acp-closure-service/internal/sweep/sweep.go
```diff
@@ -0,0 +1,247 @@
...
+
+	_, err = updater.UpdateProject(ctx, proj.ID, body)
+	return err
+}
+
+// suspend writes endDateClosureState=Suspended, unless the project is
+// already suspended (checked against the already-fetched closureState — no
+// extra round-trip), mirroring legacy's checkForOpenProject guard.
+func suspend(ctx context.Context, updater projectUpdater, proj project) error {
+	if proj.ClosureState != nil && *proj.ClosureState == "Suspended" {
```
rksk: **The idempotency guard reads a different field from the one it writes.**

This reads `closureState` and compares it to `"Suspended"`, but the write two lines down sets `endDateClosureState`. Per `entity-service/internal/domain/entity.go` on `main`, those are separate fields: `ClosureState` sits alongside `EndDateClosureState`, `InvoiceDueDateClosureState`, and `ComplianceViolationClosureState`, and its doc comment says `ClosureState` itself is **not settable** because it is derived. So it is a roll-up over all three dimensions, not an echo of what this code wrote.

Casing is a second concern: the closure-status vocabulary in that same file is lowercase (`"open"`, `"suspended"`, `"read_only"`), while this compares against and writes `"Suspended"`.

If the derived roll-up never surfaces exactly `"Suspended"`, this guard never trips, and the job re-PATCHes every past-due project on every scheduled run indefinitely, repeatedly poking a flow that is shared with the live customer-facing path.

`TestProcessProject_SuspendGuardSkipsAlreadySuspendedProject` cannot catch this: the fixture hard-codes `suspended := "Suspended"`, so it proves the branch executes, not that real data ever produces that value. And because all the testing described in the PR ran under `DRY_RUN=true`, this write path has never executed against real data.

Suggested fix: read a genuinely suspended project back from staging, confirm the exact field name and casing, and guard on the field this code actually writes (`endDateClosureState`) rather than the derived roll-up. Worth a comment explaining why the three similar names in play here (`closureStatus` in the search filter, `closureState` read, `endDateClosureState` written) are each correct.

integrations/acp-closure-service/internal/sweep/sweep.go
```diff
@@ -0,0 +1,247 @@
...
+	pcRaw, err := reader.SearchProjectContacts(ctx, proj.ID, []byte(`{}`))
+	if err != nil {
+		return nil, nil, fmt.Errorf("search project contacts: %w", err)
+	}
+	var pcResp projectContactSearchResponse
+	if err := json.Unmarshal(pcRaw, &pcResp); err != nil {
+		return nil, nil, fmt.Errorf("parse project contacts: %w", err)
+	}
+
+	acRaw, err := reader.SearchAccountContacts(ctx, proj.accountID(), []byte(`{}`))
```
rksk, line 202: **Missing the empty-account-ID guard that `resolveAccountManagerEmail` has.**

`resolveAccountManagerEmail` correctly short-circuits on `accountID == ""`, but this path does not. `url.PathEscape("")` is `""`, so a project with no linked account produces `POST /accounts//contacts/search`, which will fail, propagate out of `fetchContacts`, and fail the whole project in `processProject`.

The timing is unfortunate: `fetchContacts` is only reached when `needsCustomerAudience(window)` is true, i.e. the 15/7/0-day windows, so an account-less project fails exactly at the point where it matters most.

`TestProject_AccountIDIsEmptyWhenAccountIsAbsent` proves the accessor returns `""`, but nothing covers what `fetchContacts` does with that value. Suggest guarding on `proj.accountID() == ""` (skip the account-contact lookup and let the three-tier fallback land on the AM nudge) plus a test for the account-less project reaching a customer-audience window.

integrations/acp-closure-service/cmd/acp-closure/main.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+
+	slog.Info("acp-closure-service finished",
+		"runID", runID,
+		"dryRun", dryRun,
+		"projectsEvaluated", result.ProjectsEvaluated,
+		"failureCount", len(result.Failures),
+	)
+	for _, f := range result.Failures {
+		slog.Error("project failed", "runID", runID, "projectID", f.ProjectID, "err", f.Err)
```
rksk, line 85: **The process exits 0 even when every project failed.**

`sweep.Run`'s error return is reserved for a fatal page-fetch failure, which is a reasonable design. But it means a token expiry or an upstream 401 partway through the sweep fails every remaining project, and `main` still falls off the end with exit status 0.

For a scheduled task, the exit code is the alerting signal: the Choreo run reports green while nothing was actually processed. The per-project `slog.Error` lines are there, but they only help someone who is already reading logs.

Suggest exiting non-zero when `len(result.Failures) > 0` (or above some threshold, if partial failure is expected to be routine). Worth deciding deliberately either way and documenting it, since "scheduled job succeeded" is what operations will trust.

integrations/acp-closure-service/internal/sweep/run.go
```diff
@@ -0,0 +1,109 @@
...
+
+		for _, proj := range page.Projects {
+			result.ProjectsEvaluated++
+			if err := processProject(ctx, reader, updater, ntf, now, proj); err != nil {
+				slog.ErrorContext(ctx, "processProject failed", "projectID", proj.ID, "err", err)
+				result.Failures = append(result.Failures, ProjectFailure{ProjectID: proj.ID, Err: err})
+			}
+		}
+
+		if !page.HasMore {
```
rksk, line 106: **Unbounded pagination loop with write side effects.**

The loop terminates purely on `page.HasMore`, and `searchProjectsResponse.Total` is parsed but never used. If the upstream ever returns a stuck or incorrect `hasMore`, this becomes an infinite loop that keeps issuing PATCHes against a shared data source. `hasMore` is also one of the response fields the PR documents as undocumented in the upstream `openapi.yaml`, which is exactly the sort of field worth not trusting as the sole loop condition.

Cheap hardening, either or both:
- break when a page returns zero projects (`len(page.Projects) == 0`), which also covers the offset-past-the-end case;
- bound the iteration count using the already-parsed `Total`, and log a warning if the loop hits that bound while `hasMore` is still true.

Minor related note: this is offset-based paging over a result set the same loop is mutating. `sortBy: endDate` is stable here because the sweep never writes `endDate`, so I do not think projects can be skipped today. But if a future change causes a write to move a project out of the `closureStatus: "Open"` filter, offsets shift underneath the loop and projects get silently skipped. Worth a comment recording why it is safe now.

rksk (commented):
Four inline comments below on the sweep/orchestration layer. Two of them (the suspend guard and the empty-account-ID path) I'd treat as merge blockers; the other two are hardening for a job that writes to a shared, live data source on a cron.

Separately, and not inline because it isn't tied to one line: `notify.LoggingNotifier` is the only `Notifier` implementation, but `recordNoticeSent` writes `actionSendEmailNotification: "SUCCESSFUL"` unconditionally. With `DRY_RUN=false` that means the job records notices as sent, advances `based_on_subscription_end_date.event_type`, and suspends past-due projects, while no email was ever sent. `DRY_RUN` defaulting to `true` is good, but it is one config value away from that outcome. Consider making the absence of a real sender structurally block writes (fail fast in `main`, or write `IGNORED` rather than `SUCCESSFUL` when the notifier does not actually send) rather than relying on the default.

The pure/I-O split, the DI-based dry-run, and the byte-for-byte preservation of the sibling `suspensionProcessState` sections are all well done.

acp-closure-service/go.mod
```diff
@@ -0,0 +1,5 @@
+module github.com/wso2-open-operations/cs-tools/acp-closure-service
```
rksk: **This should not be a new top-level directory — please move it under `integrations/` to match the rest of the repo.**

The PR body calls this a "new top-level Go module", but the repo root only has three code directories today, and each earns its place:

```
cs-tools/
├── apps/                 # user-facing products (csm-portal, customer-portal, mttr-application, security-advisory-patches-portal)
├── entity-service/       # the single shared data layer everything else consumes
└── integrations/         # backend services and integrations (csm-integration-service, customer-service)
```

`entity-service/` sits at root because it is *the* shared data layer, not because root is a general-purpose bucket. This component is a backend service that consumes `csm-integration-service` over M2M and deploys as a Choreo component — that is squarely what `integrations/` is for. Adding a fourth root entry makes the repository structure less legible for a fourth service, then a fifth.

Suggested: `integrations/acp-closure-service/`.

Three things that come with the move:

**1. Keep the module path and the directory in agreement.** After the move this should read:

```
module github.com/wso2-open-operations/cs-tools/integrations/acp-closure-service
```

Worth being deliberate here, because there is an existing counterexample not to copy: `integrations/csm-integration-service/go.mod` declares `.../cs-tools/operations/csm-integration-service` while actually living under `integrations/`. That mismatch is confusing enough already without a second instance. This is also why the move is much cheaper now than post-merge: every internal import across the ~20 Go files in this PR carries the module prefix.

**2. Add a `.choreo/` directory.** `integrations/csm-integration-service/` has one, and this PR ships none despite the component being deployed as a Choreo Scheduled Task. Right now the deployment config for this service exists only in whatever was clicked into the Choreo console, which makes the schedule, the env-var set, and the build config unreviewable and unreproducible. For a job whose behavior hinges entirely on two env vars (`DRY_RUN`, `TEST_PROJECT_ID`), having that config in the repo is worth more than usual.

**3. Register the component in the root `README.md`.** It has a "Repository Structure" tree and a per-component "Components" section; this PR updates neither, so the service is invisible to anyone starting at the repo root. (Heads-up: that tree is already stale on the `operations/` vs `integrations/` point above, so if you fix it while you're in there, it is an unrelated but welcome cleanup.)

No objection to any of the module's internal layout — `cmd/` + `internal/` with a `Makefile`, `.env.example`, `README.md`, and `CLAUDE.md` mirrors `csm-integration-service` closely. It is only the parent directory that is out of step.

rksk (commented):
One more, on placement: this shouldn't be a new top-level directory. Detail inline on `go.mod`, since the module path has to change with the move and that ripples through every import in the PR — much cheaper to do now than after merge.

rksk (changes requested):

integrations/acp-closure-service/.env.example
```diff
@@ -0,0 +1,18 @@
+# csm-integration-service (M2M gateway — see internal/entity for why this,
+# not entity-service directly)
+CSM_INTEGRATION_BASE_URL=
+CSM_INTEGRATION_TOKEN_URL=
+CSM_INTEGRATION_CLIENT_ID=
+CSM_INTEGRATION_CLIENT_SECRET=
```
Rashmika998, line 6: I believe you need scopes as well

acp-closure-service/go.mod
Ananth-Abi: 2. Checked Choreo's own docs for Scheduled Task components specifically — looks like they're configured entirely through the console UI (schedule + env vars via the Configurations pane), not via a checked-in .choreo/component.yaml. That file seems to be specific to Service-type components (REST/GraphQL/etc.) for declaring endpoints, which would explain why csm-integration-service has one. Am I missing something, or does a Task-type component genuinely not use one? If not, would documenting the actual deployed schedule/env-vars as plain text in the module's README address the reviewability concern instead?

integrations/acp-closure-service/.env.example
Ananth-Abi: Good catch on scopes — currently they're intentionally hardcoded in the code (entity.RequiredScopes), not something set via .env, specifically so a deployment can't accidentally misconfigure them with a typo or a missing scope. So there's no new env var to add here, but I take your point that .env.example doesn't currently tell someone what scopes their client credentials need to actually be granted upstream (in whatever system issues the client ID/secret) — I'll add a comment documenting the required scope list next to the credential fields so that's clear during setup. Let me know if you meant something else by this.

acp-closure-service/internal/sweep/sweep.go
Ananth-Abi: Fixed — the guard now checks endDateClosureState != "Open" instead of closureState == "Suspended". Verified against a real suspended project: closureState was "Suspended" (the derived roll-up) but endDateClosureState had already progressed to "Closed" — confirming equality-to-"Suspended" alone would miss that case. Added EndDateClosureState to the project struct, updated the guard, and added regression tests for both the Closed and explicit-Open cases.

integrations/acp-closure-service/internal/sweep/sweep.go
Ananth-Abi: Fixed — fetchContacts now returns early with just the project contacts when proj.accountID() == "", before ever calling SearchAccountContacts. Added TestProcessProject_CustomerAudienceWindowSkipsAccountContactLookupWhenNoAccount to cover it.

integrations/acp-closure-service/cmd/acp-closure/main.go
Ananth-Abi: Fixed — main now calls os.Exit(exitCode(len(result.Failures))). Pulled the decision into a small pure exitCode(failureCount int) int function so it's actually testable (added cmd/acp-closure/main_test.go — this module's first test on the main package).

integrations/acp-closure-service/internal/sweep/run.go
Ananth-Abi: Fixed — the loop now breaks immediately on an empty page, and separately bounds iteration using the already-parsed Total field against cumulative ProjectsEvaluated, logging a warning if that bound is hit while hasMore is still true. Added two tests, both using a call-count safety valve in the mock so a run against the old unbounded code fails fast in the test itself instead of hanging.

acp-closure-service/go.mod
Ananth-Abi: Moved — acp-closure-service/ → integrations/acp-closure-service/, module path updated to match, and it's registered in the root README's structure tree and Components section (also fixed the stale operations/csm-integration-service line while I was in there). Confirmed with you separately that .choreo/ isn't needed here since Scheduled Task components are console-configured — treating that part as resolved, not skipped.

integrations/acp-closure-service/.env.example
Ananth-Abi: Following up on this — after talking it through with Sajith, reversing course from my last reply here: CSM_INTEGRATION_SCOPES is now a required env var (same pattern as the other CSM_INTEGRATION_* vars), not hardcoded. Asgardeo enforces the actual grant regardless, so there's no real safety benefit to hardcoding it in code, and it means the scope list can be adjusted without a redeploy. .env.example documents it (points at entity.RequiredScopes for the current required set) — thanks for catching this.

integrations/acp-closure-service/internal/notify/notify.go
```diff
@@ -0,0 +1,85 @@
...
+func (n *LoggingNotifier) Send(ctx context.Context, notice Notice) error {
+	n.Logger.InfoContext(ctx, "notice",
+		"kind", notice.Kind,
+		"window", notice.Window,
+		"projectID", notice.ProjectID,
+		"recipient", notice.Recipient,
+		"resolvedVia", notice.ResolvedVia,
+	)
+	return nil
+}
```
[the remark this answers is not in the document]
Ananth-Abi: Fixed — LoggingNotifier.Send no longer logs notice.Recipient. Existing tests already only asserted on resolvedVia, so this doesn't lose any test coverage; ResolvedVia/Kind/ProjectID/Window still give full observability without the PII.

rksk:
@coderabbitai resume

integrations/acp-closure-service/internal/notify/notify.go
```diff
@@ -0,0 +1,85 @@
...
+	Logger *slog.Logger
+}
+
+// Send logs the notice and always succeeds.
+func (n *LoggingNotifier) Send(ctx context.Context, notice Notice) error {
+	n.Logger.InfoContext(ctx, "notice",
+		"kind", notice.Kind,
+		"window", notice.Window,
+		"projectID", notice.ProjectID,
+		"recipient", notice.Recipient,
```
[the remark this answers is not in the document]
rksk: @Ananth-Abi If this log is there only for testing purposes until the notification service is plugged, then it is fine. Otherwise let's sanitize the PII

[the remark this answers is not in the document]
Ananth-Abi: @rksk Yeah the log is only for testing purpose

rksk (approved):

Rashmika998 (approved):
LGTM

PR #1371, opened 2026-08-05
fix(customer-portal): split product-consumption tracking base URL from subscription

Rashmika998:
## Summary
- The Ballerina reference (`apps/customer-portal/backend`) configures `product_consumption_subscription` and `product_consumption_tracking` as two independently-configurable base URLs (`productConsumptionBaseUrl` vs `productConsumptionTrackingBaseUrl` in `config.toml`) — they only happen to be set to the same host in that backend's current config, not guaranteed by the module architecture.
- `backend-v2`'s `internal/productconsumption` package had collapsed this into a single `BaseURL`, silently assuming the two would always match.
- Adds `PRODUCT_CONSUMPTION_TRACKING_BASE_URL`, which the `usage-tracking` API (`ImportDeploymentUsage` / `POST /deployment-usages`) now uses instead of the subscription/license API's base URL. Falls back to `PRODUCT_CONSUMPTION_BASE_URL` when left unset, so existing deployments where both happen to coincide keep working unchanged.
- Updated `.env.example`, `README.md`, and `CLAUDE.md` accordingly.

## Test plan
- [x] `go build ./...`, `go vet ./...`, `gofmt -l .` all clean
- [x] `gosec -fmt=text ./...` — 0 issues
- [x] `go test ./...` passes

## Summary by CodeRabbit

* **New Features**
  * Product consumption integrations can now use separate base URLs for subscription/license and usage-tracking services.
  * Usage-tracking requests are routed to the configured tracking service.
  * Tracking service configuration falls back to the primary service URL when not provided.
  * Deployment-product metrics routes continue to support existing URL formats.

* **Documentation**
  * Updated setup guidance and configuration examples to explain the separate service URLs and fallback behavior.

### DIFF
--- a/apps/customer-portal/backend-v2/.env.example
+++ b/apps/customer-portal/backend-v2/.env.example
@@ -31,9 +31,14 @@ AI_CHAT_AGENT_WS_SCOPES=
 # local development only.
 WS_ALLOWED_ORIGINS=
 
-# Product-consumption service (deployment license provisioning + usage import) —
-# a separate service, not entity-service.
+# Product-consumption services — not entity-service. The Ballerina backend
+# configures these as two independently-configurable base URLs
+# (product_consumption_subscription vs product_consumption_tracking), so set
+# both here too. PRODUCT_CONSUMPTION_TRACKING_BASE_URL falls back to
+# PRODUCT_CONSUMPTION_BASE_URL when left unset — only needed if the two
+# services are hosted separately.
 PRODUCT_CONSUMPTION_BASE_URL=
+PRODUCT_CONSUMPTION_TRACKING_BASE_URL=
 PRODUCT_CONSUMPTION_SCOPES=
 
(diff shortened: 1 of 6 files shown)


Rashmika998:
**Update: this PR now also includes an urgent crash fix.**

`POST /deployments/{deploymentId}/products/{productId}/metrics/search` and `POST /deployments/products/{id}/instances/metrics/search` register as ambiguous patterns on Go 1.22+'s `net/http.ServeMux` — this panics at `mux.HandleFunc` time and crashes the server on every startup (confirmed this is what's showing in the Choreo deployment logs). Fixed by merging both under one wildcard pattern with manual dispatch (see the latest commit). Recommend merging this PR promptly since the server currently cannot start on `dev-app-csm-portal`'s tip.

rksk:
@coderabbitai resume

rksk (approved):

PR #1390, opened 2026-08-06
[CSM Portal Microapp] Fix case time cards to scope by caseId and show properly styled entries

2003dinijay:
## Summary

- The case detail **Time Tracking** tab believed `BeSearchTimeCardsFilters.caseId` was non-functional against `/time-cards/search` (a stale assumption — it used to always return `total: 0`), so it worked around that by scoping to the case's project and filtering client-side, capped at a single page with a "some entries may not be shown" notice and a separate "Open Time Cards" button as the escape hatch.
- A live payload capture confirmed `caseId` now scopes correctly and returns an accurate `total` — the webapp's `useCaseTimeCards` already switched to it, with its own comment pointing at the backend fix (PR #1133) this relies on. Dropped the project-scoped workaround for the direct `caseId` filter.
- The tab now pages through the case's own cards the same way the "Time cards" > All tab does — an `IntersectionObserver` scroll sentinel fetches further pages for the rare case with more than one page's worth of entries — instead of stopping at a fixed first page. The truncation notice and "Open Time Cards" navigation button are gone; the tab is now the complete list on its own.
- Each logged entry now gets a proper bordered, tinted card treatment (matching this app's own `TimeSheetCard` row and the webapp's `CaseTimeCardsPanel`) instead of rendering as a bare, borderless line of text.

## Test plan

- [x] `tsc --noEmit` and `eslint` pass clean.
- [x] Not exercised in a local browser session — the microapp authenticates via a native superapp bridge with no standalone fallback; verification relied on a live payload capture and code review against the webapp's already-fixed equivalent.

## Summary by CodeRabbit

* **New Features**
  * Time cards now load directly for the selected case with reliable server-side filtering.
  * Additional time cards load automatically as you scroll.
  * Counts now reflect partially loaded results as more records are retrieved.

* **UI Improvements**
  * Updated time-card rows with bordered, tinted cards and improved metadata layout.
  * Added loading skeletons during pagination.
  * Improved empty-state behavior while pages are still loading.

### DIFF
--- a/apps/csm-portal/microapp/src/components/case-detail/TimeTrackingTab.tsx
+++ b/apps/csm-portal/microapp/src/components/case-detail/TimeTrackingTab.tsx
@@ -14,11 +14,10 @@
 // specific language governing permissions and limitations
 // under the License.
 
-import { useState } from "react";
+import { useEffect, useRef, useState } from "react";
 import { Box, Button, Divider, Skeleton, Stack, Typography } from "@wso2/oxygen-ui";
 import { Plus } from "@wso2/oxygen-ui-icons-react";
-import { useNavigate } from "react-router-dom";
-import { useQuery, useQueryClient } from "@tanstack/react-query";
+import { useInfiniteQuery, useQueryClient } from "@tanstack/react-query";
 import { timecards } from "@src/services/timecards";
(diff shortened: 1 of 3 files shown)


SSSD-2001:
@coderabbitai review

rksk (approved):

PR #1414, opened 2026-08-10
Add E2E coverage for case and service request creation

SParaparan:
### What

Adds Playwright end-to-end coverage for the two main customer-facing creation
flows, on top of the scaffolding merged earlier. **8 tests across 3 suites**,
all verified green against staging.

| Suite | Tests | Covers |
|---|---|---|
| `Create Case` | 6 | One per project type (Subscription, Managed Cloud Subscription, Cloud Support), plus S1/S2/S3 severities on Subscription |
| `Close Case` | 1 | Creates a case, then closes it via the confirmation dialog |
| `Service Request` | 1 | Raises a Generic Request from the Get Help dropdown |

### Structure

Test data is fully separated from test logic:

- `tests/e2e/config/testData.ts` — **master config, values only.** Project
  fixtures per type (id, deployment, product), severity and issue-type
  vocabularies, and per-flow input. Repointing the suite at another tenant means
  editing this file and nothing else.
- `tests/e2e/pages/` — page objects (locators + actions, no behavioural
  assertions)
- `tests/e2e/utils/selectors.ts` — how the UI is addressed
- `tests/e2e/utils/caseFlows.ts` — multi-step flows shared across specs
- `tests/e2e/specs/` — the specs

The Create Case suite is data-driven: it loops over the configured project types,
and a type whose fixture is incomplete **skips with a message naming the missing
fields** rather than failing obscurely.

### Configuration

Run config moved into `.env.e2e` (committed, no secrets), loaded by
`playwright.config.ts` via node's built-in `process.loadEnvFile` — no dotenv
dependency. Precedence: real env / CLI > `.env.e2e.local` (git-ignored) >
`.env.e2e`. So `pnpm run test:e2e` works with no env vars on the command line.

### Behaviour worth knowing about

Several things the UI does that the tests had to be built around, each verified
live rather than assumed:

- **Cloud Support hides the Deployment field entirely** and locks it to primary
  production. The spec asserts it stays hidden — a regression that widened
  deployment choice would now fail.
- **Service requests POST to `/cases`**, not `/service-requests`; an SR is a case
  carrying `requestType: "service_request"`.
- **The "Generic Requests" catalog contains an item named "General Requests"** —
  the wording genuinely differs, recorded in config.
- The case description is a **Lexical contenteditable**, so it is typed rather
  than filled.
- Catalog-item variables have no stable ids or `<label for>`, so the Request
  Details input is located by role ∩ enabled; a `toHaveCount(1)` guard makes an
  ambiguous match fail loudly.

## Summary by CodeRabbit

- **Tests**
  - Added end-to-end coverage for creating support cases and service requests across supported project types.
  - Added coverage for closing cases, confirmation dialogs, status updates, and success navigation.
  - Expanded validation for form fields, project-specific options, and backend responses.
  - Improved reusable test setup with captured sessions and centralized test data.
  - Added a CI check to prevent false-success results when test sessions are unavailable.

- **Documentation**
  - Updated E2E setup guidance, environment configuration, session requirements, and local testing instructions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,10 @@ handoffs/
 
 # Local working docs (PR reply/decision drafts, etc.) — never committed/pushed
 local-docs/
+
+# Local Claude Code config — kept out of version control.
+# Note other CLAUDE.md files in this repo (csm-portal, entity-service,
+# integrations/*) are deliberately tracked, so these are ignored by path rather
+# than by a blanket CLAUDE.md pattern.
+.claude/agents/e2e-test-writer.md
+apps/customer-portal/CLAUDE.md
--- a/apps/customer-portal/webapp/.env.e2e
+++ b/apps/customer-portal/webapp/.env.e2e
@@ -1,0 +1,30 @@
+# Playwright E2E configuration — read by playwright.config.ts.
+#
+# Committed defaults for the whole team. Precedence, highest first:
+#   1. real environment / CLI  (E2E_BASE_URL=… pnpm run test:e2e)
+#   2. .env.e2e.local          (your personal overrides, git-ignored)
+#   3. this file
+#
+# NO SECRETS HERE — this file is committed. Login is by replaying a captured
+# browser session (tests/e2e/storageState/session.json, git-ignored), so no
+# credentials or tokens are ever needed as env vars.
+
+# Environment the specs run against.
+#
+# This MUST match the `origin` of the captured session bundle: the storage
+# replay is scoped to that origin, so a mismatched run restores nothing and the
+# app boots signed-out. withSession() detects this and skips with a message
+# naming both origins.
+#
+# Set to http://localhost:3000 if you recapture session.json against the local
+# dev server (and drop E2E_NO_WEBSERVER so Playwright boots it for you).
+E2E_BASE_URL=http://localhost:3000
+
+# Don't boot the local vite dev server — required whenever E2E_BASE_URL points
+# at an already-running deployment.
+#
+# To have Playwright start `pnpm run dev` itself, override with an explicitly
+# EMPTY value (`E2E_NO_WEBSERVER=`) rather than omitting the key: a key absent
+# from .env.e2e.local still picks up the value below, while an empty value reads
+# as falsy.
+E2E_NO_WEBSERVER=1
--- a/apps/customer-portal/webapp/.gitignore
+++ b/apps/customer-portal/webapp/.gitignore
@@ -39,3 +39,5 @@ config.js
 /playwright/.cache/
 # Captured auth sessions hold real tokens — never commit them.
 tests/e2e/storageState/*.json
+# Session-capture instructions — kept local.
+tests/e2e/auth/README.md
--- a/apps/customer-portal/webapp/tests/e2e/config/testData.ts
+++ b/apps/customer-portal/webapp/tests/e2e/config/testData.ts
@@ -1,0 +1,291 @@
+// Copyright (c) 2026 WSO2 LLC. (https://www.wso2.com).
+//
+// WSO2 LLC. licenses this file to you under the Apache License,
+// Version 2.0 (the "License"); you may not use this file except
+// in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+//
+// MASTER TEST DATA for the E2E suite — VALUES ONLY.
+//
+// Every spec sources its environment-specific data from here, not just
+// case creation. This file holds no locators, no helpers, and no page logic:
+// selectors live in `../utils/selectors.ts`, actions in `../pages/`.
+//
+// Pointing the suite at a different tenant means editing this file and nothing
+// else. Values below were captured against **staging**
+// (https://support-stg.wso2.com), which is what `.env.e2e` targets.
+//
+
+// ─────────────────────────────────────────────────────────────────────────────
+// Project fixtures
+// ─────────────────────────────────────────────────────────────────────────────
+
+/** Project types this suite exercises.
+ *
+ * The portal derives feature visibility from `GET /projects/{id}/features`
+ * rather than from these labels, but the type still drives a few rules —
+ * deployment filtering, case/SR product category, and which severities are on
+ * offer — so each type needs its own dedicated project. */
+export const ProjectType = {
+  SUBSCRIPTION: "Subscription",
+  MANAGED_CLOUD_SUBSCRIPTION: "Managed Cloud Subscription",
+  CLOUD_SUPPORT: "Cloud Support",
+} as const;
+
+export type ProjectType = (typeof ProjectType)[keyof typeof ProjectType];
+
+/** A project the suite runs against, plus the option labels its forms offer.
+ *
+ * `deployment` and `productVersion` are per-project rather than shared: those
+ * dropdowns are populated from the project's own deployments, so the labels
+ * differ between projects even for the same project type. */
+export interface ProjectFixture {
+  /** Project id as it appears in the URL (`/projects/<id>/...`).
+   *
+   * An empty string means "not yet captured for this environment" — specs must
+   * skip rather than run against the wrong project. */
+  id: string;
+  /** Project name exactly as displayed in the portal. */
+  name: string;
+  /** The project's type label, for specs that assert type-driven behaviour. */
+  type: ProjectType;
+  /** Deployment option label, exactly as rendered in the Deployment dropdown.
+   * Empty when `autoSelectsDeployment` is true — there is nothing to pick. */
+  deployment: string;
+  /** Sysid of the deployment above. The case form selects by label, so this is
+   * only needed by specs that assert against the API payload or filter by
+   * deployment in a URL. */
+  deploymentId: string;
+  /** True when the case form hides the Deployment field and locks it to the
+   * project's primary production deployment. Cloud Support and Cloud Evaluation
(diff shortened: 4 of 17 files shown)


apps/customer-portal/webapp/.env.e2e
```diff
@@ -0,0 +1,30 @@
...
+# Environment the specs run against.
+#
+# This MUST match the `origin` of the captured session bundle: the storage
+# replay is scoped to that origin, so a mismatched run restores nothing and the
+# app boots signed-out. withSession() detects this and skips with a message
+# naming both origins.
+#
+# Set to http://localhost:3000 if you recapture session.json against the local
+# dev server (and drop E2E_NO_WEBSERVER so Playwright boots it for you).
+E2E_BASE_URL=http://localhost:3000
```
chanukaranaba, line 21: **Should fix — committed default contradicts the docs, and the out-of-the-box run skips everything.**

This file ships `E2E_BASE_URL=http://localhost:3000` + `E2E_NO_WEBSERVER=1`, but:

- the README says "Default in `.env.e2e` is staging";
- the comment above says the URL *must* match the captured session's origin, and the session was captured against staging;
- with a localhost target, no dev server booted, and a staging-captured session, a fresh `pnpm run test:e2e` skips all 8 tests on the origin check — the opposite of "works with no env vars".

The staging URL isn't secret (it already appears in a `testData.ts` comment), so either commit `E2E_BASE_URL=https://support-stg.wso2.com` as the team default, or rewrite the README/comments to describe the localhost default honestly.

apps/customer-portal/webapp/.gitignore
```diff
@@ -39,3 +39,5 @@ config.js
 /playwright/.cache/
 # Captured auth sessions hold real tokens — never commit them.
 tests/e2e/storageState/*.json
+# Session-capture instructions — kept local.
+tests/e2e/auth/README.md
```
chanukaranaba, line 43: **Should fix — the capture instructions are referenced everywhere but excluded from the repo.**

`tests/e2e/auth/README.md` is git-ignored here, yet the e2e README's Layout table and both skip messages in `fixtures/test.ts` point at it ("See tests/e2e/auth/README.md"). Anyone cloning the repo hits that message and finds no such file — they can't capture a session, so they can never run the suite.

Please commit the capture instructions (redacting anything sensitive). This matters doubly since this setup is the template for the other apps.

apps/customer-portal/webapp/tests/e2e/config/testData.ts
```diff
@@ -0,0 +1,291 @@
...
+  logType: string;
+  /** Start Time, as whole days before today. */
+  startDaysAgo: number;
+  /** End Time, as whole days before today. */
+  endDaysAgo: number;
+  purpose: string;
+  description: string;
+}
+
+export const PRODUCT_LOGS_REQUEST_INPUT: ProductLogsRequestInput = {
```
chanukaranaba, line 251: **Consider:** `PRODUCT_LOGS_REQUEST_INPUT` (and its `ProductLogsRequestInput` interface, including `startDaysAgo`/`endDaysAgo`) has no consuming spec in this PR — the 8 tests don't include a product-logs request. Suggest dropping it until the spec lands, or adding the spec here; committed-but-unused fixtures tend to drift from the real catalog.

apps/customer-portal/webapp/tests/e2e/config/testData.ts
```diff
@@ -0,0 +1,291 @@
...
+  catalog: "Information Request",
+  catalogItem: "Request Product Logs",
+  logType: "Carbon",
+  startDaysAgo: 3,
+  endDaysAgo: 1,
+  purpose: "This is a test Inofrmation Request SR for MS subscription project",
```
chanukaranaba, line 257: Typo: "Inofrmation" → "Information" (here and on line 259). Worth fixing even though this block is currently unused — these descriptions are the suite's mechanism for keeping the permanent backend records identifiable, so they should be clean.

apps/customer-portal/webapp/tests/e2e/utils/caseFlows.ts
```diff
@@ -0,0 +1,138 @@
...
+// under the License.
+
+//
+// Multi-step case flows shared by more than one spec. These sit above the page
+// objects: they drive a whole user journey and assert it succeeded, so a spec
+// that merely needs a case to exist can call one instead of restating every
+// step.
+//
+
+import { expect, test, type Page } from "@playwright/test";
```
chanukaranaba: **Consider:** this imports `test`/`expect`/`Page` directly from `@playwright/test`, while the specs import them from `../fixtures/test` (which re-exports them, evidently for exactly this purpose). Harmless today because the fixture doesn't extend `base`, but the moment `withSession` becomes a proper `base.extend()` fixture, this direct import silently bypasses it. Suggest importing from the fixture path everywhere so that refactor stays safe.

apps/customer-portal/webapp/tests/e2e/pages/CaseDetailPage.ts
```diff
@@ -0,0 +1,97 @@
...
+   * Clicks "Close" and confirms the dialog, then waits for the PATCH to
+   * succeed.
+   *
+   * The button stays disabled until the case-states metadata resolves (the
+   * action needs a state key to patch with), so this waits for it to be
+   * enabled rather than clicking into a no-op.
+   *
+   * @returns The successful PATCH response for the caller to assert on.
+   */
+  async closeCase(): Promise<import("@playwright/test").Response> {
```
chanukaranaba: Nit: inline `import("@playwright/test").Response` — add `Response` to the type import on line 17 like the other types in this file.

.gitignore
```diff
@@ -34,3 +34,10 @@ handoffs/
 
 # Local working docs (PR reply/decision drafts, etc.) — never committed/pushed
 local-docs/
+
+# Local Claude Code config — kept out of version control.
+# Note other CLAUDE.md files in this repo (csm-portal, entity-service,
+# integrations/*) are deliberately tracked, so these are ignored by path rather
+# than by a blanket CLAUDE.md pattern.
+.claude/agents/e2e-test-writer.md
```
chanukaranaba, line 42: **Consider:** these two entries (`.claude/agents/e2e-test-writer.md`, `apps/customer-portal/CLAUDE.md`) are personal tooling config unrelated to the E2E framework — fine to have, but they'd travel better in their own housekeeping commit/PR so this one stays purely about the test framework.

chanukaranaba (commented):
Great foundation for the Playwright rollout — since this framework will be the template for the other apps, the structure here matters more than usual, and it's in good shape: clean data/selectors/pages/flows/specs layering, page objects with no assertions, backend-verified writes (asserting on the captured `POST /cases` response rather than just UI navigation), and engineered failure modes (fixture-gap skips that name the missing fields, origin-mismatch skips that name the remedy, the `toHaveCount(1)` guard on the dynamic input). License headers and secret hygiene (`storageState/*.json` ignored) are all correct.

Inline comments below — the first two are the ones I'd fix before merge, the rest are minor.

Two notes that don't anchor to a diff line:

1. **`.github/workflows/e2e-tests.yml` will go green with everything skipped.** It's `workflow_dispatch`-only today (safe), but a manual run has no `session.json` on the runner, so `withSession()` skips all 8 tests and the job passes — easily misread as "E2E passed". Worth either failing the workflow when no session bundle exists, or noting in it that it isn't runnable until CI session provisioning is solved. Something to settle before this pattern is copied to the other apps.

2. Process nit: the PR is opened from your fork's `main`. A feature branch per PR keeps your fork's `main` in sync with upstream and lets you open a second PR before this one merges.

SParaparan:
@coderabbitai review

apps/customer-portal/webapp/.env.e2e
SParaparan: The developer url is provided.

.gitignore
SParaparan: Yes. It is added in the gitignore for the same purpose.

apps/customer-portal/webapp/tests/e2e/config/testData.ts
SParaparan: The follow up test cases will use this.

apps/customer-portal/webapp/tests/e2e/config/testData.ts
SParaparan: The follow up test cases will use this.

apps/customer-portal/webapp/.gitignore
SParaparan: session.json is a temporary implementation. It will replace with real username password login.

chanukaranaba:
Thanks for the quick turnaround on the review — the updates land well:

- The CI guard step + the honest header in `e2e-tests.yml` is exactly the right treatment for the green-with-all-skips trap.
- All imports now route through `fixtures/test` (with the why documented), so a future `base.extend()` refactor stays safe.
- The new `create-case-validation.spec.ts` is a great addition — record-free by design, with the stayed-on-form assertion turning a silently broken validation rule into a visible failure.

The structure (config / selectors / pages / flows / specs) is solid as the template we'll replicate to the other apps — no changes needed there.

**Not blockers to merge** — please pick these up as follow-ups:

1. **Align the three places that describe the committed default target.** `.env.e2e` ships `E2E_BASE_URL=http://localhost:3000`, but the E2E README (lines 54 and 65) and the new `e2e-tests.yml` header both say the committed default is staging. If localhost is the intended default, update those docs; if staging is, update `.env.e2e`. As committed, a fresh `pnpm run test:e2e` skips every test on the origin check, which is worth clearing up for the next person who clones.
2. **Open a tracked issue for replacing session-capture with real username/password login** (per your comment on the `auth/README.md` thread), and link it in the E2E README — that keeps "temporary" from quietly becoming permanent, and until then documents that only the captured-session holder can run the suite.
3. Fix the "Inofrmation" → "Information" typo in `PRODUCT_LOGS_REQUEST_INPUT` before the follow-up specs consume it — those descriptions become permanent staging records verbatim.

Good to merge from my side once you're ready. 🚀

chanukaranaba (approved):

PR #1428, opened 2026-08-11
Add deployment E2E coverage

SParaparan:
### What

Adds E2E coverage for creating a deployment from the Deployments tab of the
project details page, reached through the side nav as a user would. **6 tests.**

| Suite | Tests | Creates records? |
|---|---|---|
| `Deployment` | 1 | Yes — one deployment |
| `Deployment — validation` | 5 | **No** (one submits, backend rejects it) |

Scoped to Managed Cloud Subscription: the Add Deployment button is withheld for a
Restricted project, and deployment access is a per-project feature flag.

### Files

| File | Purpose |
|---|---|
| `specs/project-details/add-deployment.spec.ts` | Happy path |
| `specs/project-details/add-deployment-validation.spec.ts` | Field gating and rejection cases |
| `pages/ProjectDeploymentsPage.ts` | Side-nav navigation, tab switch, modal |
| `config/testData.ts` | `DeploymentType`, `DEPLOYMENT_INPUT` |
| `utils/selectors.ts` | `PROJECT_DETAILS`, `ADD_DEPLOYMENT` |

### Two things the flow required getting right

**Names must be unique.** The backend rejects a duplicate with
`409 {"message":"A deployment with the same name already exists for the
project."}`, and there is **no delete endpoint** for deployments — so a fixed
name works exactly once. The spec appends a sortable timestamp
(`Automation Test Deployment 2026-08-11-15-40-11`), which also keeps the
accumulated records identifiable.

**The create POST must be matched exactly.** `POST
/projects/{id}/deployments/search` also fires on this page, so a substring URL
match can capture the wrong response. The predicate uses `pathname.endsWith`.

Relatedly, the status check sits in an **assertion** rather than in the response
predicate. Requiring 2xx to match means a rejected create never resolves, and the
test dies on an unexplained 180s timeout instead of reporting the server's
message — which is exactly how the duplicate-name failure first presented. Worth
copying to the case, service-request and security-report specs, which still bake
`isSuccess` into their predicates.

### Validation coverage

`AddDeploymentModal` gates its confirm button on
`name && deploymentTypeKey && description`, so an incomplete form cannot be
submitted at all — those cases assert **button state**, not error text.

| Test | Submits? |
|---|---|
| keeps submit disabled until every required field is filled | No |
| keeps submit disabled when only the description is missing | No |
| keeps submit disabled for a whitespace-only name | No |
| rejects a duplicate deployment name → 409 | Yes, rejected |
| cancelling closes the modal without creating a deployment | No |

The whitespace case exists because `isValid` trims before comparing. The cancel
case counts create requests with a `page.on("request")` listener and asserts
zero, rather than inferring it from the modal closing.

### Notes on the UI

- The modal has real ids and associated labels (`#deployment-name`,
  `#deployment-type`, `#deployment-description`), so no structural locators were
  needed here.
- Its confirm button shares an accessible name with the button that opens it, so
  the opener is scoped to `<main>` and the confirm to the dialog.
- The type options are `Development, QA, Staging, Stress, UAT, Primary
  Production` — there is no plain "Production".

### One product observation

**A failed create shows the user nothing.** On 409 the modal stays open with no
visible error and `role="alert"` empty — from the user's side the button simply
does nothing. Asserted as current behaviour, but it looks like a defect worth
fixing separately.

### Testing

`tsc -b` on the E2E tsconfig and `eslint tests/e2e` are clean.

- **Happy path: verified.** Passes on consecutive runs (10.1s, 8.0s) — the repeat
  being the case the duplicate-name fix addresses.
- **Validation suite: not verified.** The captured session expired mid-run and
  all five redirected to sign-in. Nothing in the output points at a defect in the
  specs, but they have not been seen green.

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end coverage for creating deployments from the project Deployments tab.
  * Added validation coverage for required fields, whitespace-only names, duplicate names, and cancellation.
  * Added checks confirming successful creation, deployment identifiers, and list visibility.
  * Added coverage for deployment type selection, descriptions, modal interactions, and project navigation readiness.

### DIFF
--- a/apps/customer-portal/webapp/tests/e2e/config/testData.ts
+++ b/apps/customer-portal/webapp/tests/e2e/config/testData.ts
@@ -235,6 +235,52 @@ export const SERVICE_REQUEST_INPUT: ServiceRequestInput = {
   description: "This is a test Generic Request SR for MS subscription project",
 };
 
+/** Deployment types offered by the Add Deployment modal. Verified live against
+ * staging — note there is no plain "Production": the production-like option is
+ * "Primary Production". */
+export const DeploymentType = {
+  DEVELOPMENT: "Development",
+  QA: "QA",
+  STAGING: "Staging",
+  STRESS: "Stress",
+  UAT: "UAT",
+  PRIMARY_PRODUCTION: "Primary Production",
+} as const;
+
+export type DeploymentType =
+  (typeof DeploymentType)[keyof typeof DeploymentType];
+
+/** Content submitted by the add-deployment flow. */
+export interface DeploymentInput {
+  /** Name *prefix*. The spec appends a timestamp because the backend rejects a
+   * duplicate name with 409 and there is no delete endpoint, so a fixed name
+   * only ever works on the very first run. */
+  namePrefix: string;
+  /** Name of a deployment that already exists on the project, for exercising the
+   * duplicate-name rejection. Durable as a fixture because deployments have no
+   * delete endpoint, so once created this name cannot disappear. */
+  existingName: string;
+  type: DeploymentType;
+  /** Goes in the field labelled "Description *" — the modal does not call it
+   * "Deployment Description". */
+  description: string;
+}
+
+/**
+ * Deployment to create on the Managed Cloud Subscription project.
+ *
+ * ⚠️ Creates a permanent record on every run — `POST /projects/{id}/deployments`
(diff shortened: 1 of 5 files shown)


SParaparan:
@coderabbitai resume

SParaparan:
@coderabbitai resume

SasmithaDilshan (approved):

PR #1432, opened 2026-08-12
[CSM][Web] Allow resolving call requests (reject/cancel/notes) on a closed case

Hesara2003:
## Purpose
A closed case's Call Requests tab treated every action as read-only, on the assumption that "case closed" meant "nothing here can change." That's only true for two of the five agent actions. Reject, Cancel, and Send call notes have no case-state restriction anywhere in the stack — the Go BFF (`PatchCallRequest`) and the entity-service (`updateCallRequest`) are both pure pass-throughs to ServiceNow's own PATCH, with no business logic of their own. The blanket UI gate was the *only* thing stopping an agent from resolving a call request whose case closed before the call itself got scheduled or rejected — leaving it stuck showing as active indefinitely, with no way to clear it from the portal.

> **Scope note:** Schedule/Reschedule/Create call request stay disabled on a closed case. Those genuinely are backend-restricted — `CALL_REQUEST_ALLOWED_CASE_STATES` documents mirroring ServiceNow's own `CallRequestUtils.scheduleCall` gate, and `closed` isn't one of its 5 allowed states. Enabling those in the UI would just replace a disabled button with a submit-then-fail round trip. Unblocking them for real needs a ServiceNow-side change, which is out of scope for this repo.

## Goals
- An agent can Reject / Cancel / Send call notes on a call request whose case has already closed.
- Schedule / Reschedule / Create stay blocked where the restriction is real (ServiceNow's own state gate), with the actual reason shown instead of a generic "read-only" message.
- No behavior change for a case that's still open.

## Approach
- `CallRequestsTable.tsx`: replaced the `isClosed` prop (which disabled every row action) with `scheduleBlockReason` — non-null exactly when the case's state doesn't accept a schedule (this already covers `closed`, since it's never one of the 5 allowed states). Only `schedule`/`reschedule` actions are disabled on it now; the reason string is the tooltip.
- `CallRequestsWidget.tsx`: dropped `isClosed` entirely. It was redundant with `stateBlockReason` (`callRequestCaseStateBlockReason(caseState)`) everywhere it was used — the Create button's disabled/tooltip logic and the auto-open-create effect now read `stateBlockReason` directly, which also fixes the Create button showing a generic "This case is closed" message instead of the more specific, already-correct reason.
- `CsmCaseDetailPage.tsx`: stopped passing the now-removed `isClosed` prop (it was `c.state === "closed"`, duplicating what `caseState` already carries).

## User stories
As a CS engineer, I can reject, cancel, or send notes on a call request even after its parent case has closed, so it stops sitting in the queue as if it were still active.

## Release note
Fixed: Call requests on a closed case can now be rejected, canceled, or concluded with notes. Scheduling/creating a new call request still requires an open case.

## Documentation

## Automation tests
- Unit tests: 6 new in `CallRequestsTable.test.tsx` — Reject/Cancel/Send call notes stay enabled with a block reason set (across the three states that offer each), Schedule is disabled when a reason is present, everything is enabled with no reason, and `onAction` fires with the right action/row.
- Verified the two new tests that matter most (Schedule-disabled, Cancel-stays-enabled) actually catch the regression: run against the pre-change component, both fail.
- Verified live against a real closed staging case with several call requests stuck in an active-looking state: with the old code, Reject was genuinely disabled; with this change, enabled — confirmed via a local-only Playwright script (not part of the committed suite), removed after verification. No mutating action (Reject/Cancel/etc.) was actually submitted against real data — only button enabled/disabled state was checked.

## Security checks

## Test environment
- `tsc -b`, `pnpm run lint` (no new findings beyond one pre-existing, unrelated warning), `pnpm run test` all passing locally.
- No API contract change — every mutation this enables was already accepted end-to-end by the Go BFF and entity-service before this PR; only the frontend's own UI gate changes.

## Summary by CodeRabbit

* **New Features**
  * Scheduling and rescheduling actions are disabled only when a scheduling restriction applies.
  * Restriction details are shown in action tooltips.
  * Reject, cancel, and send-notes actions remain available regardless of case status.

* **Bug Fixes**
  * Call request creation and automatic opening now follow the applicable case-state restrictions more accurately.

* **Tests**
  * Added coverage for action availability and rejection behavior.

### DIFF
--- a/apps/csm-portal/webapp/src/features/csm-cases/components/CallRequestsTable.test.tsx
+++ b/apps/csm-portal/webapp/src/features/csm-cases/components/CallRequestsTable.test.tsx
@@ -1,0 +1,102 @@
+// Copyright (c) 2026 WSO2 LLC. (https://www.wso2.com).
+//
+// WSO2 LLC. licenses this file to you under the Apache License,
+// Version 2.0 (the "License"); you may not use this file except
+// in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+import { fireEvent, render, screen } from "@testing-library/react";
+import { describe, expect, it, vi } from "vitest";
+import "@testing-library/jest-dom/vitest";
+import { CallRequestsTable } from "@features/csm-cases/components/CallRequestsTable";
+import type { BeCallRequestView } from "@api/backend/types";
+
+function callRequest(overrides: Partial<BeCallRequestView> = {}): BeCallRequestView {
+  return {
+    id: "cr-1",
+    number: "CSTASK0000001",
+    reason: "Discuss upgrade",
+    state: { id: "pending_on_wso2", label: "Pending on WSO2" },
+    createdOn: "2026-08-01 10:00:00",
+    updatedOn: "2026-08-01 10:00:00",
+    ...overrides,
+  };
+}
+
+describe("CallRequestsTable — closed-case gating", () => {
+  it("keeps Reject enabled on a `pending_on_wso2` request even when the case is closed", () => {
+    render(
+      <CallRequestsTable
+        requests={[callRequest()]}
+        onAction={vi.fn()}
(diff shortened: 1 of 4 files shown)


s-sandali:
@coderabbitai review

Hesara2003:
@rksk — flagging something found while working on this, for visibility.

The closed-case block on call requests turned out to be split across two different layers, only one of which this PR can fix:

- **Reject / Cancel / Send call notes** — fixed here. Nothing in our own stack (the Go BFF or the entity-service) restricted these by case state; it was only this frontend's own blanket UI gate.
- **Schedule / Reschedule** — still blocked, but that restriction lives entirely on the ServiceNow side, not in any of our repos. The case-state allow-list used when scheduling a call doesn't include `closed`.

Could we get `closed` added to that allow-list on the SN side? Right now it blocks scheduling an already-pending call request just because its case happened to close first.

PR #1440, opened 2026-08-12
acp-closure-service: real notice content (subject/body templates, structured recipients)

Ananth-Abi:
Closes wso2-enterprise/digiops-cs#2748

Follow-up to #1330, covering everything on this branch since that PR
merged. Reworks ACP Phase 1 notice content from development-shaped
logging (a `Kind` label + a single `Recipient` string) into real,
Chamara-approved subject-line and email-body templates for every window —
so wiring in real email sending later is a drop-in change, not a
redesign.

## What changed

- **Real subject/body templates** for every notice type: internal
  day-count reminder (90/60/30/15/7), internal day-0 suspension notice,
  customer day-count notice (15/7), customer day-0 notice, and the
  no-business-contact urgent notice — confirmed against real examples,
  wording verified character-for-character.
- **Structured `Recipients`** (Account Owner, Renewal Manager, Technical
  Owner, Customer) replacing the single `Recipient` string — closes a gap
  where Renewal Manager/Technical Owner were fetched from entity-service
  but silently dropped by `accountDTO` having no field for them.
- **Internal and customer notices sent separately** per window, not
  bundled into one — their content genuinely differs, not just their
  recipient list. The `[ACP]` prefix now correctly applies to every
  internal-audience copy (90 through 0), not just 90/60/30.
- **No-business-contact notice now reaches all three internal
  recipients** (Account Owner + Renewal Manager + Technical Owner), not
  Account Owner alone.
- **Dropped the raw "would update project" write-confirmation log
  line** — only real notice/email-style log lines are emitted now, so
  dry-run output reads as actual email content end to end.
- **Fix:** the project key field is `key` on the live `GetProject`
  response, not `projectKey` as `csm-integration-service`'s own
  `openapi.yaml` documents — was silently always empty before this fix,
  caught via a real Postman response.

## Verification

- Full test suite (`go vet` + `-race`) green throughout, TDD one seam at
  a time, matching this component's existing conventions.
- A Standards+Spec two-axis code review was run against this diff before
  opening the PR; findings (a duplicated `Notice`-construction block, a
  stale doc comment, missing recipient names in the log) were applied.
- Confirmed against real staging data via the Choreo POC component,
  including a real (non-dry-run) write against the dedicated test
  project — subject, body, and recipients all matched the confirmed
  templates exactly.

## Summary by CodeRabbit

* **New Features**
  * Added distinct internal, customer, and no-business-contact notifications for closure reminders.
  * Notifications now include project details, dates, tailored subjects, message content, and appropriate recipients.
  * Internal notices include account owner, renewal manager, and technical owner contacts.
  * Customer notices are sent when contacts are available; otherwise, an internal alert is issued.
  * Added clear day-of-closure messaging and suspension status logging.

* **Bug Fixes**
  * Improved handling of project, account, key, date, and contact information.
  * Dry-run updates now complete without misleading update logs.

### DIFF
--- a/integrations/acp-closure-service/CLAUDE.md
+++ b/integrations/acp-closure-service/CLAUDE.md
@@ -221,6 +221,19 @@ wrong answer:
   `internal/sweep/types.go`'s `project.Account` has always expected the
   nested `{id, name}` shape; only the doc comment needed correcting once the
   broader `SearchProjects` gap closed.
+- **Project key field name.** `csm-integration-service`'s own `openapi.yaml`
+  documents this field as `projectKey` on the `Project` schema. The real,
+  live `GetProject` response actually names it `key` (confirmed directly by
+  the user via Postman against the dedicated test project — the response
+  had `"key": "APPSUB"`, no `projectKey` field at all). `internal/sweep/
+  types.go`'s `project.ProjectKey` was tagged `json:"projectKey"` for a
+  while as a result — silently, always empty on every real response, since
+  the tag never matched anything on the wire. Caught only because the
+  notice-content redesign started actually reading and logging the value;
+  before that, nothing exercised it. Now tagged `json:"key"`, confirmed
+  against the real response. If this ever gets "corrected" back to
+  `projectKey` by an openapi.yaml update, verify live behavior again before
+  copying it — don't just trust the spec.
 
 ## Open dependencies
 
--- a/integrations/acp-closure-service/CLAUDE.md
+++ b/integrations/acp-closure-service/CLAUDE.md
@@ -262,7 +262,10 @@ wrong answer:
 - TDD throughout: red before green, one seam at a time. Seams under test:
   `closure.Decide`, `recipients.ResolveCustomerContact` /
   `AccountManagerEmail`, `suspensionstate.LastNoticeWindow` /
-  `WithSubscriptionEndDateState`, `sweep.processProject`, `sweep.Run`.
+  `WithSubscriptionEndDateState`, `sweep.processProject`, `sweep.Run`, and
+  the pure subject/body builders (`internalNoticeSubject`,
+  `customerNoticeSubject`, `internalNoticeBody`, `customerNoticeBody` in
+  `sweep.go`) tested directly rather than only through `processProject`.
   `main.go` and the two logging/no-op implementations
   (`notify.LoggingNotifier`, `sweep.DryRunProjectUpdater`) are deliberately
   untested, matching this repo's convention that wiring-only code and
--- a/integrations/acp-closure-service/CLAUDE.md
+++ b/integrations/acp-closure-service/CLAUDE.md
@@ -76,8 +76,9 @@ decision logic cheaply testable without mocks.
 have exactly two side-effecting dependencies — `projectUpdater` (writes) and
 `notifier` (sends) — expressed as small interfaces. `main.go` decides which
 concrete implementation to inject based on `DRY_RUN`:
-`sweep.DryRunProjectUpdater` (logs, never calls `UpdateProject`) vs. the real
-`*entity.Client`. Reads (`SearchProjects`, `GetAccount`,
+`sweep.DryRunProjectUpdater` (silently no-ops, never calls `UpdateProject`
+— see the notice-content-redesign section below for why it deliberately
+doesn't log) vs. the real `*entity.Client`. Reads (`SearchProjects`, `GetAccount`,
 `SearchProjectContacts`, `SearchAccountContacts`) are never dry-run-gated —
 fetching and deciding has no write effect to protect against.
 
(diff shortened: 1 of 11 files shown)


integrations/acp-closure-service/CLAUDE.md
```diff
@@ -92,22 +95,80 @@ loop's `offset := 0` line. This is what backs safe testing against a single
...
+  both. The internal notice fires unconditionally for every window
+  (90/60/30/15/7/0); a second notice (customer, or the no-business-contact
+  nudge) fires only for 15/7/0.
+- **`Subject`** has no single template — four distinct ones
+  (`internalNoticeSubject`/`customerNoticeSubject` in `sweep.go`), confirmed
+  against multiple real examples from Chamara:
+  - Internal, day-count (90/60/30/15/7): `"[ACP] {N} Days Reminder of
+    Project for {ProjectName} of {AccountName}"`. The `[ACP] ` prefix marks
+    "this is the internal-audience copy" and applies to **every** window,
+    including 15/7 — not just 90/60/30. (An earlier version of this logic
```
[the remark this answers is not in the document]
Ananth-Abi: Fixed in 6e55d9d — removed the space inside the code span.

integrations/acp-closure-service/internal/notify/notify.go
```diff
@@ -24,40 +24,58 @@ package notify
...
+	ProjectKey  string
+	StartDate   time.Time
+	EndDate     time.Time
+	Window      closure.NoticeWindow
+	// Subject is the notice's title line — either the day-count reminder
+	// ("{N} Days Reminder of Project for {ProjectName} of {AccountName}",
+	// [ACP]-prefixed only for the internal-only 90/60/30 windows) or the
+	// no-business-contact notice's fixed "[Urgent] [ACP] No Business
+	// Contacts Specified for Project {ProjectName}".
+	Subject string
```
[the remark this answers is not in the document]
Ananth-Abi: Good catch — that comment was left over from before the [ACP] prefix rule
was corrected to apply to every internal window (90 through 0), not just
90/60/30. Rewrote it in 6e55d9d to describe all five subject variants
(internal day-count, internal day-0, customer day-count, customer day-0,
no-business-contact) instead of just the stale one.

integrations/acp-closure-service/internal/sweep/sweep.go
```diff
@@ -83,101 +83,288 @@ func needsCustomerAudience(window closure.NoticeWindow) bool {
...
+	internalNotice.Subject = internalNoticeSubject(window, proj.Name, accountName(proj))
+	internalNotice.Body = internalNoticeBody(window, proj, contacts.AccountOwner.Name)
+	internalNotice.Recipients = internalRecipients
+
 	if !needsCustomerAudience(window) {
 		return ntf.Send(ctx, internalNotice)
 	}
 
-	projectContacts, accountContacts, err := fetchContacts(ctx, reader, proj)
+	if err := ntf.Send(ctx, internalNotice); err != nil {
```
rksk: **The internal notice is sent (a real, logged email) before fetchContacts is called for the 15/7/0 windows, so a transient fetchContacts failure leaves the internal notice already sent but the window's suspensionProcessState un-persisted, causing a duplicate internal-notice send on the next sweep.**

At a 15/7/0-day window, ntf.Send(ctx, internalNotice) succeeds at line 280, then reader.SearchProjectContacts or SearchAccountContacts fails transiently at line 284 (fetchContacts). notifyForWindow returns an error, processProject (line 41) skips recordNoticeSent, so suspensionProcessState is never advanced. The next daily sweep re-evaluates the same window as still due and calls resolveAccountContacts + ntf.Send(internalNotice) again — a second real day-count reminder email for the same window. Pre-diff, fetchContacts ran before any Send, so a failure there aborted with zero notices sent; this diff's reordering introduces the duplicate-send window.

integrations/acp-closure-service/internal/sweep/dryrun.go
```diff
@@ -16,23 +16,24 @@
...
+// direction, the only log output that matters for a dry run is the
+// notify.LoggingNotifier "notice" line (the actual email content) — a
+// separate "would update project" line describing the raw PATCH body is
+// noise once every window produces a real notice log, and stays noise once
+// real email sending replaces LoggingNotifier.
+type DryRunProjectUpdater struct{}
 
-// UpdateProject logs the write that would have happened and always succeeds.
+// UpdateProject always succeeds without doing anything.
 func (u *DryRunProjectUpdater) UpdateProject(ctx context.Context, id string, body []byte) ([]byte, error) {
```
rksk, line 37: **DryRunProjectUpdater.UpdateProject no longer logs anything, and on the day-0 path where ShouldNotify is already false (prior run recorded the notice) but ShouldSuspend is still true, notifyForWindow is never called either — so a dry run produces zero log output for a project that is about to be suspended.**

A project reaches day-0, gets notified and recordNoticeSent persists lastNoticeWindow<=0 on run N, but suspend() (sweep.go:441) fails or the process is interrupted before completing. On run N+1, closure.Decide (internal/closure/decide.go) returns ShouldNotify=false (alreadyNotified) and ShouldSuspend=true. processProject (sweep.go:56-67) skips the notify branch entirely (no notify.LoggingNotifier log line) and calls suspend(), whose UpdateProject is now a silent no-op with no logging at all. An operator running DRY_RUN=1 to preview what would change sees nothing for this project, even though a live run would suspend it.

integrations/acp-closure-service/internal/sweep/sweep.go
```diff
@@ -83,101 +83,288 @@ func needsCustomerAudience(window closure.NoticeWindow) bool {
...
-// same window in the same run (see shouldSuppressInternalNotice).
+// internalNoticeSubject builds the always-sent internal (Account Owner/
+// Renewal Manager/Technical Owner) notice's title line. Every internal copy
+// is [ACP]-prefixed regardless of window — that prefix marks "this is the
+// internal-audience copy," not "this is a 90/60/30 window" (confirmed
+// against real examples from Chamara spanning 90 through 0; an earlier
+// version of this function had the rule backwards). Day-0 gets distinct
+// "suspension notice" wording instead of "N Days Reminder" — there's no
+// "days remaining" left to report once a project is actually suspended.
+func internalNoticeSubject(window closure.NoticeWindow, projectName, accountName string) string {
```
rksk, line 107: **internalNoticeSubject (and customerNoticeSubject's account-naming sibling) produces a real outbound email subject with a dangling "of " when the project has no linked account, since accountName(proj) returns "" for a nil Account.**

A project with proj.Account == nil fires a 90/60/30 window: internalNoticeSubject(window, proj.Name, "") returns "[ACP] 90 Days Reminder of Project for {ProjectName} of " — a live email subject line ending in a trailing "of ". The PR's own sweep_test.go hardcodes and asserts this exact malformed string as the expected output instead of guarding against it.

integrations/acp-closure-service/internal/sweep/sweep.go
```diff
@@ -83,101 +83,288 @@ func needsCustomerAudience(window closure.NoticeWindow) bool {
...
+Start Date: %[4]s
+
+End Date: %[5]s
+
+Since the project is suspended, kindly take the remedial actions to reinstate the subscription support. We appreciate your prompt attention to this matter.
+
+Best Regards,
+WSO2 Team`
+
+func internalNoticeBody(window closure.NoticeWindow, proj project, accountOwnerName string) string {
```
rksk, line 172: **internalNoticeBody's greeting/Account Owner line goes blank when the Account Owner has no name on file, since contacts.AccountOwner.Name flows unguarded into the %[1]s "Dear" placeholder.**

A project's account has an Account Manager assigned with no Name on file (or none assigned at all — contactFromPersonRef returns a zero Contact). internalNoticeBody produces "Dear \n\n..." and "Account Owner: \n" in a real outbound email body; the same empty-name substitution recurs in noBusinessContactBody's %[3]s Account Owner field.

integrations/acp-closure-service/internal/sweep/sweep.go
```diff
@@ -83,101 +83,288 @@ func needsCustomerAudience(window closure.NoticeWindow) bool {
...
-		}
-		return ntf.Send(ctx, nudgeNotice)
+	resolution := recipients.ResolveCustomerContact(projectContacts, accountContactsList)
+
+	if !resolution.NeedsAMNudge {
+		customerNotice := baseNotice(proj, window)
+		customerNotice.Subject = customerNoticeSubject(window, proj.Name)
+		customerNotice.Body = customerNoticeBody(window, proj)
+		customerNotice.Recipients = internalRecipients
+		customerNotice.Recipients.Customer = resolution.CustomerContact
```
rksk, line 325: **Removing shouldSuppressInternalNotice means the Account Owner (and now also Renewal Manager/Technical Owner) unconditionally receives both the day-count/suspension internal notice and the separate no-business-contact urgent notice in the same run, restoring the double-send the deleted function existed to prevent.**

A project has an Account Manager but no resolvable project/account customer contact, at a 15/7/0-day window. The same three internal recipients get two emails back to back in one sweep pass: the ordinary "N Days Reminder"/suspension notice, then immediately the "[Urgent] [ACP] No Business Contacts" notice for the identical project/window — noise the removed shouldSuppressInternalNotice/TestProcessProject_SuppressesInternalNoticeWhenNudgeGoesToSameRealRecipient test previously guarded against, with no replacement test asserting the new behavior either way.

integrations/acp-closure-service/internal/sweep/sweep.go
rksk, line 107: **The day-0-vs-reminder branch (if window == closure.NoticeWindow0) is independently re-implemented in four separate functions (internalNoticeSubject, customerNoticeSubject, internalNoticeBody, customerNoticeBody) instead of resolved once.**

A future change to day-0/terminal-window semantics requires four synchronized edits instead of one; the component's own doc comment on internalNoticeSubject already notes "an earlier version of this function had the rule backwards" for this exact branch shape, evidence it's error-prone to keep re-deriving. A closure.NoticeWindow.IsSuspension() bool method or a single per-window template table would resolve it once.

integrations/acp-closure-service/internal/sweep/sweep.go
```diff
@@ -83,101 +83,288 @@ func needsCustomerAudience(window closure.NoticeWindow) bool {
...
-		Window:    window,
-		ProjectID: proj.ID,
-		Recipient: amEmail,
+	internalRecipients := notify.Recipients{
+		AccountOwner:   contacts.AccountOwner,
+		RenewalManager: contacts.RenewalManager,
+		TechnicalOwner: contacts.TechnicalOwner,
 	}
 
+	internalNotice := baseNotice(proj, window)
```
rksk, line 298: **The three notice-construction blocks (internal/customer/nudge) each repeat baseNotice(...) + four manual field assignments with only Subject/Body/Recipients/ResolvedVia differing, and the !needsCustomerAudience early-return duplicates the same ntf.Send(internalNotice) call already made at line 280.**

Maintenance cost: adding a fifth notice type or a new shared field on Notice requires editing three near-identical assignment blocks instead of one constructor call; a newNotice(base, subject, body, recipients, resolvedVia) helper would collapse ~9 duplicated lines and remove the duplicate ntf.Send call site at line 277 vs 280.

integrations/acp-closure-service/internal/sweep/types.go
```diff
@@ -36,9 +36,21 @@ import (
 // never read Account directly, so callers don't need to duplicate the nil
 // check.
 type project struct {
-	ID      string             `json:"id"`
-	Account *projectAccountRef `json:"account"`
-	EndDate *time.Time         `json:"endDate"`
+	ID   string `json:"id"`
+	Name string `json:"name"`
```
rksk, line 40: **The project struct's new Name/ProjectKey/StartDate fields are regression-tested only against a GetProject-shaped fixture; the production sweep path unmarshals the same struct from SearchProjects's leaner item shape, whose field parity for these three fields is unverified.**

Run()'s default (TEST_PROJECT_ID unset) path pages through /projects/search and unmarshals each item into project. If SearchProjects's response item omits name/key/startDate the way it previously omitted account before being independently confirmed via Postman, every notice sent from an unattended production sweep gets ProjectName="", ProjectKey="", StartDate=nil — blank subjects and bodies ("Project Key: ", "Start Date: ") for every real project, not just the manually-tested one. This is the same class of wire-vs-spec mismatch this PR already caught once (projectKey→key) but here left unverified for the code path that actually runs unattended.

integrations/acp-closure-service/internal/sweep/sweep.go
Ananth-Abi: Good catch — you're right, this reordering introduced a real regression.
Fixed by moving fetchContacts + ResolveCustomerContact back before the
internal notice send, so a transient fetch failure now sends zero
notices again instead of a partial send with no state record. Added a
regression test (TestProcessProject_CustomerAudienceWindowFetchContactsFailureSendsNothing)
covering exactly this.

integrations/acp-closure-service/internal/sweep/dryrun.go
Ananth-Abi: Confirmed, this was a real gap — that retry path never calls
notifyForWindow at all, so removing DryRunProjectUpdater's logging left
it completely silent. Fixed by having suspend() itself log
unconditionally (real or dry-run) right before attempting the write,
matching Run's own existing always-on logging pattern rather than
reintroducing the raw write-body dump.

integrations/acp-closure-service/internal/sweep/sweep.go
Ananth-Abi: Confirmed and fixed — added a projectOfAccount helper that omits the
" of {AccountName}" clause entirely when there's no linked account,
instead of leaving a trailing "of " with nothing after it. Added
explicit regression cases for both the day-count and day-0 subject
variants with no account.

integrations/acp-closure-service/internal/sweep/sweep.go
Ananth-Abi: Confirmed this can render blank when the Account Manager has no name on
file. Discussed with the team — leaving this as-is for now rather than
inventing a fallback greeting; a nameless-AM account is expected to be
rare in practice, and any fallback text would be a content decision
worth its own sign-off rather than something to guess at here.

integrations/acp-closure-service/internal/sweep/sweep.go
Ananth-Abi: This one's accurate about the mechanism but not an oversight — sending
both the internal notice and the no-business-contact notice in this
case was a deliberate simplification we decided on earlier, rather than
carrying forward the old shouldSuppressInternalNotice logic. It's
recorded in CLAUDE.md's "Notice audience matrix and content" section
with the reasoning, including an explicit note to revisit if it proves
too noisy in practice. Appreciate you pushing on it independently either
way — happy to reconsider if it turns out to be a real problem once
this is live.

integrations/acp-closure-service/internal/sweep/sweep.go
Ananth-Abi: The construction duplication is now reduced via a baseNotice() helper
for the shared project-identity fields. The duplicate Send call for the
internal notice is trickier — collapsing it to one call site conflicts
with fixing your first comment above (contacts need to resolve before
any send, but internal-only windows should still skip the contact-fetch
entirely) — so there are still two call sites, but both now wrap errors
identically rather than differing. Documented the tradeoff in a comment
on notifyForWindow.

integrations/acp-closure-service/internal/sweep/types.go
Ananth-Abi: Verified against a real /projects/search response — name/key/startDate
are all present there too, not just on GetProject. Added
TestProject_ParsesNameKeyAndStartDateFromRealSearchProjectsResponse
using that exact response as the fixture, so this is now covered the
same way GetProject's shape already was.

rksk:
@coderabbitai resume

rksk (approved):

PR #1456, opened 2026-08-13
[CSM] Live case-activity updates via Server-Sent Events

Hesara2003:
## Purpose
Real-time updates for the case Activities tab (new comments, status changes) — viewers currently rely on a 10s staleTime/manual refresh; this pushes updates live over Server-Sent Events instead.

> **Scope note:** header-based auth only, no separate ticket/token-exchange step — the stream connects with the same auth headers every other backend call already uses, via a fetch-backed EventSource polyfill (native `EventSource` can't set custom headers).

## Goals
- Case Activities tab reflects a new comment or status change without a manual refresh
- Works across multiple backend replicas (Kafka fan-out, not a single-process pub-sub)
- No new auth mechanism — reuses the existing JWT header convention
- Graceful degradation: unset config falls back to the existing polling behavior

## Approach
- **Backend**: comment creation / status change publish `case.comment_added`/`case.status_changed` to Azure Event Hub (async, non-blocking). `internal/caseevents.Handler` (Kafka consumer, ephemeral per-replica consumer group so every replica sees every event) fans matching events out to a new in-process `internal/stream.BroadcastHub`. `GET /cases/{id}/activities/stream` serves the SSE connection on a dedicated `:9092` listener (own `http.Server`, `WriteTimeout`/`IdleTimeout: 0`) since the main `:8080` listener's timeouts would kill a long-lived connection. New `internal/middleware/cors.go` handles CORS for that listener specifically (the main API relies on the Choreo gateway for this).
- **Webapp + microapp**: `useCaseActivityStream(caseId)` hook in each, using `@sanity/eventsource` (a cross-environment EventSource polyfill supporting custom headers) to connect with the same auth headers the existing API clients already send. Listens for `case_updated` and invalidates the comments/activities query keys. Reconnects with a freshly-fetched token on error rather than relying on the polyfill's own retry.
- **Infra**: second Choreo REST endpoint (`.choreo/component.yaml`, `openapi-stream.yaml`) for the `:9092` listener.
- Both the stream URL config and the Event Hub wiring itself are optional — unset, everything falls back to the existing staleTime-based polling.

## User stories
As a CSM agent viewing a case, I see a new comment or status change from a teammate appear in the Activities tab without refreshing the page.

## Release note
Case Activities tab now updates live when a comment is added or the case's status changes, instead of relying on a manual refresh.

## Documentation
N/A — internal CSM portal UI/BFF, no external doc surface affected. OpenAPI spec for the new endpoint is included in-repo (`openapi-stream.yaml`).

## Automation tests
- Backend: new unit tests for `BroadcastHub`, the SSE handler, `caseevents` broadcast wiring, and CORS middleware — `make test` (vet + race) passing, `gosec -fmt=text ./...` reports 0 issues.
- Webapp: 6 new tests for `useCaseActivityStream` (connect/no-op/reconnect/invalidate/cleanup) — passing; full suite shows the same pre-existing failures as unmodified `dev-app-csm-portal` (verified via stash-and-compare), no regressions.
- Microapp: no test runner wired up yet (pre-existing gap) — validated via `tsc -b`/`eslint`.
- Verified against staging: real end-to-end test against the real staging Azure Event Hub (publish → per-replica consumer → broadcast → SSE → a real authenticated browser session → real React Query invalidation), confirmed working.

## Security checks
- Ran FindSecurityBugs plugin and verified report? N/A — this Go backend uses `gosec` instead; ran `gosec -fmt=text ./...`, 0 issues

## Test environment
- Backend: `go vet ./...`, `go test -race ./...`, `gosec -fmt=text ./...` all passing locally.
- Webapp: `eslint`, `vitest`, `tsc -b` all passing locally.
- Microapp: `eslint .`, `tsc -b` passing locally (no test runner configured).

## Summary by CodeRabbit

* **New Features**
  * Added live, authenticated case-activity updates to case detail pages.
  * Case comments and status changes now refresh automatically, with automatic stream recovery.
  * Added validation for supported pasted and uploaded image formats.
  * Improved authentication token recovery and sign-in handling.
* **Bug Fixes**
  * Prevented dashboard widgets from failing when optional filters are missing.
  * Added specific feedback for unsupported or oversized pasted images.
* **Configuration**
  * Added optional stream URL and CORS settings; polling remains available when streaming is unavailable.
* **Documentation**
  * Documented streaming authentication, CORS, heartbeats, and availability behavior.


s-sandali:
@coderabbitai review

YasiruHarinda:
@coderabbitai review

apps/csm-portal/backend/internal/handler/cases.go
```diff
@@ -93,12 +97,49 @@ type entityCaseClient interface {
...
+// publishAsync fires eventpublisher.Publish on its own goroutine so the HTTP
+// response doesn't wait on a Kafka round-trip. It deliberately does NOT use
+// r.Context() for the goroutine's own lifetime: net/http cancels a request's
+// context the moment the handler returns, and this goroutine is specifically
+// meant to keep running after that — so it detaches via context.WithoutCancel
+// first (same technique eventpublisher.Publish already uses internally for
+// its own failure-recording call), then applies its own bounded timeout.
+// A no-op when h.publisher is nil (Event Hub not configured — see
+// cmd/server/main.go).
+func (h *CaseHandler) publishAsync(ctx context.Context, eventType events.Type, entityID string, payload any) {
```
rksk, line 133: **Correctness:** the event publish runs in a fully detached goroutine (`publishAsync`) with no `WaitGroup` tracking, and `main.go` calls `eventPublisher.Close()` before draining in-flight HTTP requests on shutdown.

A status-change `PATCH` can return 200 right as `SIGTERM` arrives; its untracked goroutine is still running when `Close()` shuts the writer, so the event never reaches `caseevents.Handler` — no SSE broadcast — even though the state change was durably saved. This isn't a Close()/Shutdown()-ordering fix away either, since the goroutine is never awaited at all.

apps/csm-portal/backend/internal/handler/case_stream.go
```diff
@@ -0,0 +1,133 @@
...
+	// Nginx/Choreo-gateway hint to disable response buffering for this
+	// endpoint; harmless (ignored) on stacks that don't recognise it.
+	w.Header().Set("X-Accel-Buffering", "no")
+	w.WriteHeader(http.StatusOK)
+	flusher.Flush()
+
+	ch := h.hub.Register(caseID)
+	defer h.hub.Unregister(caseID, ch)
+
+	ctx := r.Context()
```
rksk, line 112: **Correctness:** this loop only exits on `r.Context().Done()`, but `http.Server.Shutdown()` does not cancel actively-running handlers' request contexts — it waits for them to return on their own, up to the shutdown timeout.

On a routine rolling deploy, if one engineer has a case-detail tab open (SSE alive on `:9092`), `streamSrv.Shutdown(shutdownCtx)` blocks for the full 15s budget, leaving no time for `srv.Shutdown()` on the main API — which can hit `main.go`'s `os.Exit(1)` path if it also has an in-flight request at that instant. A graceful drain turns into an abrupt kill.

apps/csm-portal/backend/cmd/server/main.go
```diff
@@ -67,7 +71,41 @@ func main() {
...
+		// each replica sees 100% of events instead of Kafka load-balancing
+		// them across replicas — a shared group would mean an SSE client
+		// connected to replica B never learns about an event a shared
+		// group happened to hand to replica A. See internal/stream.
+		// BroadcastHub: it only knows about subscribers on its own
+		// process, so every replica must independently observe every event
+		// to be able to broadcast it to whichever subscribers it happens to
+		// be holding the connection for.
+		consumerGroupBase := envOrDefault("EVENT_HUB_CONSUMER_GROUP", "csm-portal-backend")
+		consumerGroup := fmt.Sprintf("%s-replica-%s", consumerGroupBase, newReplicaID())
```
rksk, line 114: **Correctness:** each replica derives a consumer group from its hostname (`newReplicaID`), which starts at `FirstOffset` — so every ordinary rolling deploy (not just crash-restarts) replays the entire retained event-topic history, with no dedup/timestamp filtering before broadcast.

Right after any routine deploy, every open Activities tab gets a burst of spurious `case_updated` events for potentially days-old activity. This also leaks an unbounded number of stale, never-cleaned-up consumer groups on the broker over the service's lifetime.

apps/csm-portal/backend/cmd/server/main.go
```diff
@@ -220,20 +240,41 @@ func main() {
...
+		// every response, including a CORS preflight — CORS runs next,
+		// still ahead of Auth (see middleware.CORS's doc comment): a
+		// preflight OPTIONS carries no x-jwt-assertion for Auth to accept.
+		// In a real deployment Choreo's gateway supplies CORS itself, so
+		// this is a no-op there; it matters when the gateway isn't in the
+		// path (local development, where the browser calls this listener
+		// directly). CORS_ALLOWED_ORIGINS is a comma-separated allow-list;
+		// unset allows any origin (see middleware.CORS on why that's safe
+		// here).
 		Handler: middleware.SecurityHeaders(
```
rksk, line 284: **Security:** this PR adds `middleware.CORS(...)` to the main `:8080` server's handler chain for the first time — previously it had no in-app CORS and relied solely on the Choreo gateway. With `CORS_ALLOWED_ORIGINS` unset (the default), the middleware reflects any `Origin` header.

Any Choreo/production deployment that doesn't explicitly set `CORS_ALLOWED_ORIGINS` now lets any website read authenticated-looking cross-origin responses from the CSM backend — safe today only because auth is header-based and the attacker needs a valid `x-jwt-assertion`/`x-user-id-token`, but a silent regression the moment any future change exposes that token to page-readable JS. Worth defaulting `CORS_ALLOWED_ORIGINS` to a required/fail-closed value rather than open-by-default.

apps/csm-portal/webapp/src/features/csm-cases/api/useCaseActivityStream.ts
```diff
@@ -0,0 +1,122 @@
...
+
+      const url = `${apiConfig.streamUrl}/cases/${encodeURIComponent(caseId)}/activities/stream`;
+      source = new EventSourcePolyfill(url, {
+        headers: {
+          "x-jwt-assertion": token,
+          "x-user-id-token": idToken,
+        },
+      });
+
+      source.addEventListener("case_updated", () => {
```
rksk, line 132: **Reliability:** reconnect uses a fixed 3s retry with no exponential backoff, jitter, or attempt cap (the microapp variant additionally forces a real IdP token refresh on every reconnect regardless of whether the failure was auth-related).

Combined with the UUID-validation issue flagged on `case_stream.go` (or any sustained outage/misconfiguration), every open case-detail tab hammers the stream endpoint every 3s indefinitely with no cap.

rksk (commented):
Automated review pass (6 confirmed findings, correctness/security only).

apps/csm-portal/webapp/src/features/csm-cases/api/useCaseActivityStream.ts
```diff
@@ -0,0 +1,122 @@
...
+ * opens a fresh one with newly-fetched tokens after RECONNECT_DELAY_MS,
+ * rather than relying on that built-in retry.
+ *
+ * A no-op when `caseId` is unset or `apiConfig.streamUrl` isn't configured
+ * (Event Hub — and therefore this endpoint — is optional on the backend);
+ * callers fall back to the comments/activities queries' own staleTime.
+ */
+export function useCaseActivityStream(caseId: string | undefined): void {
+  const queryClient = useQueryClient();
+  const { getAccessToken, getIdToken } = useAsgardeo();
```
rksk: **Auth handling:** this hook imports `useAsgardeo()` directly and manually attaches `getAccessToken()`/`getIdToken()` as headers, bypassing `useAuthApiClient.ts` — the webapp's single point for token handling, which also does a trusted-origin check and a token-expiry/silent-reauth/redirect recovery flow that this hook has none of.

When a user's IdP session actually expires, every other part of the app silently re-auths or redirects to sign-in; this hook just retries forever every 3s with no recovery, so the Activities tab silently stops updating with no user-visible signal. Worth routing this through the existing client's token-recovery logic (or wrapping it) rather than pulling tokens out of the hook directly, given EventSource's construction-time-header constraint is the only reason to go around it.

rksk (approved):

PR #1467, opened 2026-08-14
Add case attachments, the cases lists

SParaparan:
Adds three Playwright suites covering the Support area, and fixes two defects in
the shared page objects that the new coverage exposed.

All suites are parameterised over the three project types — Subscription,
Managed Cloud Subscription and Cloud Support — and driven from config, so a
fourth project is a config entry rather than a spec change.

### Case attachments (18 tests)

The full lifecycle on a case: upload, list, expand, collapse, download, delete.

Two cases per project, deliberately: a **kept** case holds one permanent PNG that
the list, expand, collapse and download tests all depend on (expand needs
specifically an image), and a **delete** case runs the upload-then-remove round
trip so it ends where it started. Sharing one case would have the delete take
away the other tests' fixture. The kept upload is guarded on the file name, so a
run only uploads when the file is missing — otherwise every run would stack
another copy with nothing removing them.

Net effect on staging: three new permanent attachments, one per kept case.

### Cases lists (9 tests)

Side nav → Support Center → the Outstanding Cases card's footer buttons, into
both All Cases and My Cases.

My Cases is filtered by the backend on `filters.createdByMe`, and the heading
merely reports it — so the specs match that flag on the wire as well as checking
what renders. A heading reading "My Cases" over an unfiltered result set would
otherwise pass. A third test compares totals against All Cases (47 of 48, 12 of
13, 19 of 20), which is what proves the filter narrows by contents rather than by
label.

### Support Center cards (15 tests)

The four stat cards, and the list each one opens: Outstanding Cases, Active
Chats, Resolved Cases (Last 30d), Resolved via Chat (Last 30d).

Each card navigates whatever its count, and several read 0 on this tenant, so the
assertions are about the destination — URL, heading, description — plus the trip
back, which only offers "Back to Support Center" because the card passed a
`returnTo`.

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end coverage for Support Center cards and navigation across project types.
  * Added validation for All Cases and My Cases lists, filters, headings, totals, and creator information.
  * Added case attachment coverage for uploading, listing, previewing, downloading, and deleting files.
  * Improved response validation, navigation checks, filtering, and download verification for more reliable test results.
  * Added coverage for attachment counts, file metadata, image previews, and cleanup after deletion tests.


rksk:
@coderabbitai resume

Rashmika998 (approved):

PR #1476, opened 2026-08-17
Add call requests  related e2e and the dashboard menu items

SParaparan:
Adds two Playwright suites — call requests on a case, and the project dashboard —
for 65 tests in total. Both are driven from config and parameterised over project
type where the feature is not project-specific.

### Request Call (14 tests)

The call lifecycle on a case's Calls tab, plus field gating across all three of
its dialogs.

- **Happy path (4)** — request a call for tomorrow; reschedule it to the day
  after for an hour; dismiss the Edit modal and confirm the change is discarded;
  cancel a request outright.
- **Validation (8)** — submit gated on a reason and on a preferred time, a time
  before the allowed floor snapping back up, up to three preferred times with the
  first not removable, exactly four durations, no Reason field when rescheduling,
  and Confirm gated on a cancellation reason.
- **Tab availability (2)** — the Calls tab is withheld entirely on a case whose
  status is outside `CALL_SCHEDULABLE_CASE_STATUSES`.

⚠️ Creates permanent records: `POST /cases/{id}/call-requests` has no delete —
cancelling leaves the record behind. The suite files a request only when its own
earlier one is not already on the tab, keyed on a fixed reason, so repeated runs
do not stack them. The cancellation test keeps a *separate* request for the same
reason the attachment suite keeps a separate case: cancelling is terminal and
would leave the reschedule test with a dead button.

The validation half writes **nothing**, and proves it — each test counts the
`POST`/`PATCH` calls that left the browser and asserts zero.

### Dashboard (51 tests)

The project dashboard reached through the side nav, read-only throughout.

- **Stat cards** — all four present, exactly four, and which ones are clickable;
  each of the three clickable cards opens its list; Avg. Response Time opens
  nothing, asserted both by role and by clicking it anyway.
- **Sections** — Outstanding Support Cases and Outstanding Engagements on every
  project; Outstanding Operations only where SR/CR access exists, with expected
  visibility declared per project and its absence asserted where it should not
  render.
- **Severity donut** — all four legend entries and all four pie slices drill down
  to their severity list, each checked against its own entry.
- **Operations donut** (MCS only) — both legend entries and both pie slices drill
  down to the outstanding SR and CR lists.
- **Cases table** — open a case from a row, switch to My Cases, change page size
  to 10, page forward and back, filter by severity and clear it.

### Notes on what the assertions check

Several of these flows change nothing in the URL, so the tests assert the request
on the wire as well as what renders:

- My Cases in the cases table re-queries in place — matched on
  `filters.createdByMe`, with the dashboard URL asserted **unchanged**.
- Page size and paging are matched on `pagination.limit` / `pagination.offset`.
- The severity filter is matched on `filters.severityIds`, and clearing on its
  absence.
- Reschedule and cancel share one endpoint, so `stateKey` and
  `durationInMinutes` in the PATCH body are what tell them apart.
- The operations lists carry `outstandingOnly` in router state, not the URL, so
  the "Outstanding …" heading is the only evidence the chart passed it.

Pie slices are clicked by dispatching the event, not by clicking a point: a slice
is an arc, so its bounding-box centre falls in the donut's hole and both a normal
and a forced click land on nothing (verified live).

### Verification

All 65 pass — Request Call 14, Dashboard 51 (6.2m). `tsc --noEmit` and `eslint`
clean on every touched file.

## Summary by CodeRabbit

- **Tests**
  - Added comprehensive end-to-end coverage for dashboard cards, charts, filters, pagination, case navigation, and project-specific visibility.
  - Added end-to-end coverage for scheduling, rescheduling, and cancelling case calls, including validation, duration selection, preferred times, modal behavior, and request status.
  - Added shared test utilities and selectors to improve reliability when validating dashboard and case-call workflows.
  - Expanded search testing for severity filters, pagination, and empty severity results.

### DIFF
--- a/apps/customer-portal/webapp/tests/e2e/config/testData.ts
+++ b/apps/customer-portal/webapp/tests/e2e/config/testData.ts
@@ -342,6 +342,87 @@ export interface CaseViewSet {
  * per-severity naming.
  */
 /**
+ * The case a call is requested against, and what to request.
+ *
+ * The reason is fixed rather than stamped per run, which is what lets the spec
+ * recognise its own earlier request and not file another: a call request cannot
+ * be removed — the delete action cancels it and leaves the record — so an
+ * unguarded create would add one on every run.
+ *
+ * `durationLabel` is the default duration as the card renders it; the spec
+ * leaves the duration select untouched.
+ */
+export const CALL_REQUEST_INPUT = {
+  projectType: ProjectType.SUBSCRIPTION,
+  caseId: "3f18677f3ba6c35091404c6aa5e45ae0",
+  reason: "This is a test call request from Automation Test",
+  durationLabel: "30 minutes",
+  /** Time of day to request, on tomorrow's date. Late enough in the morning to
+   * clear the earliest-allowed time the modal computes from the case severity,
+   * which is a floor measured from now. */
+  preferredTimeOfDay: "10:00",
+  /** What a reschedule moves the request to: the day after tomorrow, at the
+   * same time of day, for an hour.
+   *
+   * The two duration strings differ because the modal and the card disagree on
+   * how to say it — the Meeting Duration option reads "1 hour", while the card
+   * always renders the raw minutes. */
+  rescheduledDurationOption: "1 hour",
+  rescheduledDurationLabel: "60 minutes",
+  /**
+   * The cancellation test's own request, kept separate from the one above.
+   *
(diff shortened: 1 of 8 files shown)


SParaparan:
@coderabbitai resume

rksk:
@coderabbitai resume

rksk (approved):

PR #1477, opened 2026-08-17
[CSM] Add deployment attachment list and upload routes to backend-v2

SasmithaDilshan:
Resolves wso2-enterprise/digiops-cs#2792 (epic wso2-enterprise/digiops-cs#2751).

## Summary

**Four frontend hooks call routes backend-v2 never implemented**, so deployment documents can currently be neither listed nor uploaded:

| Hook | Call |
|---|---|
| `useGetDeploymentDocuments` | `GET /deployments/{id}/attachments` |
| `useInfiniteDeploymentDocuments` | `GET /deployments/{id}/attachments?limit&offset` |
| `usePostDeploymentAttachment` | `POST /deployments/{id}/attachments` |

Only `PATCH /deployments/{deploymentId}/attachments/{attachmentId}` existed — an attachment could be renamed, but never created or read.

**No new upstream capability was needed.** entity-service stores attachments through its generic reference-keyed API, `entity.ReferenceTypeDeployment` already exists, and `POST /attachments` + `POST /attachments/search` were already wired. These two routes are just the deployment-scoped URL shape the frontend expects, mirroring `CaseHandler.SearchCaseAttachments`/`CreateCaseAttachment`.

- `SearchDeploymentAttachments` / `CreateDeploymentAttachment` on `DeploymentHandler`
- `dto.MapDeploymentAttachments` returns `totalRecords` (not `total`) to match the frontend's shared pagination envelope
- README, CLAUDE.md route inventory (105 → 107), `openapi.yaml`

## The reference can't be spoofed

`dto.CreateDeploymentAttachmentRequest` deliberately has **no** `referenceId`/`referenceType` field. Both are injected from the `{deploymentId}` path value, so a client cannot attach a file to a different entity — the same rule as the case-attachment path and the project-scoping rule in CLAUDE.md. A test sends a body containing `referenceId` and `referenceType: "case"` and asserts the path value still wins.

## Test plan

Container (no local Go toolchain):

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt -l .` — clean
- [x] `go test -race ./...` — all pass, incl. 3 new `*DeploymentAttachment*` tests
- [x] `gosec -fmt=text ./...` — **0 issues**
- [x] `openapi.yaml` parses; both methods and both new schemas present
- [ ] Staging: open a deployment's Documents tab, confirm the list loads, then upload a file

## Note on the spec

The two new paths use inline error responses referencing `ErrorPayload`, matching the rest of this file — my first attempt referenced a `components/responses` section that doesn't exist here, which would have produced an unresolvable spec.

## Summary by CodeRabbit

* **New Features**
  * Added deployment-scoped attachment endpoints for listing and uploading attachments.
  * Added pagination support for attachment listings using optional limit and offset parameters.
  * Added validation to ensure deployment identifiers are valid and attachment uploads use the deployment specified in the URL.
* **Documentation**
  * Updated API documentation and route inventory with the new attachment capabilities.
* **Tests**
  * Added coverage for deployment scoping, identifier validation, pagination, and upload behavior.

### DIFF
--- a/apps/customer-portal/backend-v2/CLAUDE.md
+++ b/apps/customer-portal/backend-v2/CLAUDE.md
@@ -6,7 +6,7 @@ responses for the frontend. This is a rewrite of the existing backend at
 `apps/customer-portal/backend`, modeled on `apps/csm-portal/backend`'s conventions — read that
 backend's own CLAUDE.md too if something here is underspecified.
 
-**Status: in progress.** 105 routes are wired up so far, across seven upstream services:
+**Status: in progress.** 107 routes are wired up so far, across seven upstream services:
 entity-service, the WSO2 Updates service, SCIM, the AI chat agent, the product-consumption
 service, the registry (robot-account) service, and the project-contact onboarding service (see
 "The AI chat agent", "The product-consumption service", "The registry service", and "The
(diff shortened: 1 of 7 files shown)


SasmithaDilshan:
Superseded by https://github.com/wso2-open-operations/cs-tools/pull/1479 — identical commit, but headed from the upstream branch so the stacked chain (#1478 bases on it) retargets correctly on merge. Closing to avoid two PRs for the same change.

PR #1478, opened 2026-08-17
[CSM] Decode missing case-detail fields (closedBy, engagement dates, SLA response time)

SasmithaDilshan:
Resolves wso2-enterprise/digiops-cs#2808 (epic wso2-enterprise/digiops-cs#2751).

## Summary

Same class as the deployment-count bug in #1475: the Ballerina entity-service declares these on `CaseResponse`, the Go `snCase` struct didn't, so `encoding/json` discarded whatever ServiceNow sent. The frontend's `CaseDetails` type declares five of them.

**entity-service** — declare and map, with `sysidToUUID()` on the refs like every other inbound ID:
`slaResponseTime`, `closedBy`, `hasAutoClosed`, `engagementStartDate`, `engagementEndDate`, `acknowledgedBy`, `engagementPaymentType`

**backend-v2** — decode all seven, expose the five the frontend declares (`features/support/types/cases.ts:272,290,291,296,298`).

## The exposure boundary is deliberate

| Field | Exposed? | Why |
|---|---|---|
| the five above | **yes** | frontend's `CaseDetails` declares them |
| `acknowledgedBy`, `engagementPaymentType` | **no** | decoded for parity, no frontend consumer — CLAUDE.md's "restrict, don't mirror" keeps them trimmed |
| fix-ETA quartet | **no** | CSM-internal; must never reach a customer |

A test asserts the trimmed fields don't leak, so widening exposure has to be deliberate rather than accidental. All fields are nullable end to end and **omitted rather than nulled** when absent — a case with no engagement dates shouldn't grow keys.

## Caveat worth reviewing

**Ballerina declaring a field proves it's *expected*, not that ServiceNow populates it.** The existing fix-ETA fields sit in `snCase` with comments saying ServiceNow doesn't send them today and they always unmarshal to `nil` — precedent for exactly this.

For `deployedProductCount` (#1475) we had hard proof: the working page. Here the evidence is circumstantial — the frontend reads these and works against the Ballerina backend. Declaring them is safe either way (absent stays `nil`, and the frontend handles missing keys), but don't expect all five to light up until verified against a real staging case response.

## Test plan

- [x] `go build ./...` — entity-service, backend-v2
- [x] `go vet ./...` — both
- [x] `gofmt -l .` — clean, both
- [x] `go test -race ./...` — all pass, incl. 3 new `TestMapCaseDetails_*`
- [x] `gosec -fmt=text ./...` — **0 issues**, both
- [x] `entity-service/openapi.yaml` parses with the new `CaseView` properties
- [ ] Staging: a closed case for `closedBy`/`hasAutoClosed`; an engagement case for the date fields

## Deployment order

**entity-service first, then backend-v2.** Unlike the `createdBy` change, deploying backend-v2 alone is harmless here — the fields just stay absent, as today.

### DIFF
--- a/apps/customer-portal/backend-v2/internal/dto/case.go
+++ b/apps/customer-portal/backend-v2/internal/dto/case.go
@@ -361,7 +361,16 @@ type CaseDetails struct {
 	ResolutionNotes       *string                      `json:"resolutionNotes,omitempty"`
 	WatchList             []CaseWatchListUser          `json:"watchList,omitempty"`
 	FixEta                *time.Time                   `json:"fixEta,omitempty"`
-	Tags                  []CaseTag                    `json:"tags,omitempty"`
+	// Exposed because the frontend's CaseDetails type declares them
+	// (features/support/types/cases.ts). AcknowledgedBy and
+	// EngagementPaymentType are decoded upstream but deliberately NOT exposed:
+	// no frontend consumer, so per CLAUDE.md they stay trimmed until one exists.
+	SLAResponseTime     *string   `json:"slaResponseTime,omitempty"`
+	ClosedBy            *Ref      `json:"closedBy,omitempty"`
+	HasAutoClosed       *bool     `json:"hasAutoClosed,omitempty"`
+	EngagementStartDate *string   `json:"engagementStartDate,omitempty"`
+	EngagementEndDate   *string   `json:"engagementEndDate,omitempty"`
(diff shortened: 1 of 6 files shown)


rksk:
@coderabbitai resume

rksk:
@coderabbitai resume

cloby99 (approved):

PR #1490, opened 2026-08-18
Add deployment-product and improve deployment e2e coverage

SParaparan:
Extends the deployments coverage from "can a deployment be created" to the full
lifecycle of a deployment and of the products on one — 27 tests across two
suites, up from 6.

Scoped to Managed Cloud Subscription for anything that writes, since the Add
Deployment button is withheld on a Restricted project and deployment access is a
per-project feature flag.

### Deployment (add-deployment.spec.ts) — 17 tests

- **Lifecycle**: add; edit the description; edit the name and type; create a
  non-production deployment and check the type select offers all six types; page
  the list and change its page size; expand and collapse a card; delete.
- **Validation** (8): every required field gated in turn, description-only
  missing, whitespace-only name, whitespace-only description, duplicate name
  rejected with 409, cancel creates nothing, an unchanged edit sends nothing and
  a cancelled edit is discarded, and Go Back from the delete dialog keeps the
  deployment.
- **Access** (2): the Deployments tab and its list are reachable on Subscription
  and Managed Cloud Subscription. Read-only, so it runs on both.

`add-deployment-validation.spec.ts` is folded into this file as a nested
`validation` describe, matching the product suite's shape.

### Deployment Product (add-deployment-product.spec.ts) — 10 tests

- **Lifecycle**: add a product; manage it (description, then an update-history
  entry, including the current-level readout); edit core count and TPS; edit and
  delete an update entry; add a second entry and confirm the level select stops
  offering the level already reached; delete the product.
- **Validation** (4): the Add Product form gated on product and version with the
  numeric and description fields optional, cancel creates nothing; the Manage
  Product form — an unchanged save sends nothing, Add Update gated on level and
  date, Close saves nothing; an invalid core count; Go Back keeps the product.

### Behaviours worth knowing, found while writing this

**"Delete" is a deactivation.** Both deleting a deployment and deleting a product
PATCH `{ active: false }` to the same endpoint their edit modals use — there is no
DELETE verb. The record leaves the list but is not removed, so the tests assert
`active === false` on the wire: the endpoint alone cannot tell a delete from an
edit.

**A non-production deployment leaves the first page.** The list orders production
first, so a deployment created as — or retyped to — Development is not among the
first ten. The type-change assertions are therefore on the request body; paging to
find the card would assert the sort order rather than the edit.

**Create and edit disagree on a field name**: the create form sends
`deploymentTypeKey`, the edit modal sends `typeKey`.

**Two labels per product.** The Add Product select offers `API Manager` while the
deployment's product list renders `WSO2 API Manager`, so config carries both.

**The Manage Product modal seeds its fields once, on open.** Reading a saved value
back needs the modal reopened, not a longer wait inside the open one — the tests
poll by reopening.

### Two probable defects, asserted as they behave today

- **An invalid core count is not validated in the form.** A negative is turned
  into `undefined`, dropped by `JSON.stringify`, and the resulting empty update is
  refused by the API with `400 {"message":"At least one of cores or tps or
  description or updates should be provided…"}`. The user sees that, not "core
  count must be positive".
- **A duplicate deployment name surfaces no error.** The 409 is asserted, and the
  modal stays open — from the user's side the button simply does nothing. Both are
  flagged in comments rather than written as if intended.

### Records these suites leave behind

A full run creates seven deployments on the MCS project and disposes of one (the
delete test's). Every test makes its own deployment so it can run alone, which is
the trade-off for that. Deployment names carry a millisecond timestamp because the
backend rejects duplicates with 409.

### Verification

All 28 project-details tests pass — Deployment 17 (2.1m), Deployment Product 10
(2.6m), plus the existing overview test. `tsc --noEmit` and `eslint` clean on every
touched file.

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end coverage for creating, editing, paginating, expanding, and deactivating deployments.
  * Added coverage for deployment products, including creation, validation, updates, deletion, and cancellation.
  * Expanded read-only access and persistence checks across supported project types.
  * Added assertions for API responses, modal behavior, list updates, and prevention of unintended changes.
  * Consolidated deployment creation flows and introduced shared test utilities and selectors.


SParaparan:
@coderabbitai resume

SParaparan:
@coderabbitai resume

Rashmika998 (approved):

rksk (approved):

PR #1497, opened 2026-08-18
[Customer Portal v2] Port the WebSocket side-channel and Converted-state handling

SasmithaDilshan:
## Purpose

Four behaviours the Ballerina backend gained were never ported to backend-v2. Their Ballerina commits are **not on `dev-app-csm-portal`**, which is why this backend never picked them up: `15eda8597`, `0c459a561`, `5048bb148`, `6167db970`.

## Side-channel messages no longer stall the connection

A `feedback` (answer rating) or `token_increase_request` message fell through into the chat-turn path, where `StreamChat` reads until a `final` event. The agent answers these with a `*_ack` and **never** sends a `final`, so the read ran to the five-minute idle deadline and the customer got `"Failed to process message."` — even though the rating had been stored.

**This was worse than the Ballerina symptom.** There, the streaming lock at least rejected the next message with an explicit error. Here the browser-facing read loop is strictly sequential — it does not read the next frame until the current one is handled — so the whole connection stalled for those five minutes and the customer's next message was never read off the socket at all. The chat looked dead while the socket was healthy.

Both types are now dispatched **before** the conversation guard, the streaming call, and the post-stream bookkeeping, none of which applies to them:

- persisting the "user query" makes no sense for a rating, which carries no `message`
- the resolved/auto-resolve bookkeeping does not apply either

`SendSideChannel` is a separate client method with its own **30-second** bound (vs the 5-minute chat idle timeout), terminating on `feedback_ack`, `token_request_ack` or `error`. A read timeout is reported as **success**: the write already succeeded, so the rating may well have been stored, and telling the customer it failed would be worse than saying nothing.

## `requestedBy` is stamped from the session, never the browser

The requester on a token-increase request was forwarded **verbatim from the client**, so a caller could name anyone as the actor on a durable audit row.

It is now rewritten unconditionally from the authenticated session. When there is no session email to stamp, a client-supplied value is **removed** rather than forwarded — otherwise that one path would preserve the same hole. The field is therefore either this session's user or absent; the upstream falls back to the account when absent.

## `CONVERTED` is no longer downgraded to `RESOLVED`

The resolved branch set `RESOLVED` unconditionally, so the agent reporting an issue solved silently overwrote the state of a conversation a case had been created from — and the chat stopped being attributable to the case it produced. It now checks the current state first and skips the transition when already converted.

A failed lookup deliberately does **not** block the transition: resolving is the safer default when the current state is unknown, and this branch is best-effort bookkeeping.

## A case created from a chat marks its conversation Converted

`POST /cases` forwards `conversationId` to entity-service, but entity-service does not act on it, so the source conversation kept counting as an active chat. The handler now marks it Converted after a successful create — non-blocking, since the case already exists and a conversion failure must not turn a successful creation into an error. Same as the Ballerina backend, which forwards the id and still performs this update explicitly.

`PATCH /conversations/{id}` already accepted `"converted"` (`3003eb6fe`'s equivalent), so no change was needed there.

## Testing

The `requestedBy` stamping and the converted check are extracted as **pure helpers**, so both are unit-testable without a live socket. Tests cover:

- the audit-trail substitution (a browser-supplied `requestedBy` never survives)
- the no-email removal path
- verbatim forwarding of every other field
- case / whitespace / nil handling of the state comparison

`go build`, `go vet`, `gofmt -l`, `go test -race ./...` all pass; `gosec -fmt=text ./...` reports **0 issues** (103 files).

### UI verification

1. Rate an answer in the Novera chat, then send another message — it must be answered normally rather than the chat going silent
2. Request a token increase; the audit row shows the signed-in user, not whatever the page sent
3. Create a case from a chat, then let the agent resolve it — the conversation stays **Converted**, not **Resolved**

## Deployment

**backend-v2 only** — no entity-service change. Independent of #1487 and #1489.

## Summary by CodeRabbit

* **New Features**
  * Added support for feedback and token-increase requests during live chat.
  * Automatically records when a chat conversation is converted into a case.
  * Preserves converted conversation status during automatic resolution.
  * Forwards side-channel responses through the live chat connection.

* **Security**
  * Ensures feedback requests use the authenticated session identity.

* **Tests**
  * Added coverage for request identity, payload handling, and converted conversation states.

### DIFF
--- a/apps/customer-portal/backend-v2/internal/handler/websocket.go
+++ b/apps/customer-portal/backend-v2/internal/handler/websocket.go
@@ -450,6 +450,19 @@ func (h *WebSocketHandler) handleMessage(ctx context.Context, conn *websocket.Co
 		// A failed lookup deliberately does not block the transition — the
 		// pre-existing behaviour (resolve) is the safer default when the current
 		// state is unknown, and this whole branch is best-effort bookkeeping.
+		//
+		// This read-then-write is NOT atomic, and cannot be made so from here:
+		// entity.UpdateConversationRequest carries a state and nothing else, and
+		// entity-service's conversation API has no conditional-update
+		// precondition (no expected-state field, no If-Match/ETag). So a case
+		// created from this chat in the window between the read and the write
+		// still gets downgraded to RESOLVED. The window is short and one-
+		// directional — a CONVERTED write landing after this one wins correctly —
+		// and the result is a mis-stated state, recoverable via
+		// PATCH /conversations/{id}, which accepts "converted". Closing it
+		// properly needs a conditional update upstream; do not simulate one with
+		// a re-read-and-correct loop here, which can lose the same way and would
+		// fight a legitimate concurrent update.
 		current, err := h.entity.GetConversation(ctx, conversationID)
 		if err != nil {
 			slog.WarnContext(ctx, "entity GetConversation failed before auto-resolve; proceeding", "userID", user.UserID, "conversationID", conversationID, "err", summarizeErr(err))
--- a/apps/customer-portal/backend-v2/internal/handler/websocket.go
+++ b/apps/customer-portal/backend-v2/internal/handler/websocket.go
@@ -35,6 +35,7 @@ import (
 // WebSocketHandler.
 type wsStreamer interface {
 	StreamChat(ctx context.Context, sessionID, payload string, caller aichatagent.BrowserConn) (map[string]json.RawMessage, error)
+	SendSideChannel(ctx context.Context, sessionID, payload string, caller aichatagent.BrowserConn) error
 }
 
 // wsMaxMessageBytes bounds the size of a single WebSocket frame this handler
@@ -54,12 +55,27 @@ const wsIdleTimeout = 5 * time.Minute
 type entityCommentCreator interface {
 	CreateComment(ctx context.Context, req entity.CreateCommentRequest) (entity.CreateCommentResponse, error)
 	UpdateConversation(ctx context.Context, id string, req entity.UpdateConversationRequest) (entity.UpdateConversationResponse, error)
+	GetConversation(ctx context.Context, id string) (entity.ConversationDetails, error)
 	GetProject(ctx context.Context, id string) (entity.ProjectDetailsView, error)
 }
 
 // wsAutoResolveState is entity-service's ConversationState value used to
 // auto-resolve a conversation when the AI agent reports the issue solved.
 const wsAutoResolveState = "RESOLVED"
+
+// msgTypeFeedback/msgTypeTokenIncreaseRequest are the inbound side-channel
+// message types — a thumbs up/down on an answer, and a request asking support to
+// raise a token limit. Mirrors MSG_TYPE_FEEDBACK/MSG_TYPE_TOKEN_INCREASE_REQUEST
+// in apps/customer-portal/backend's modules/ai_chat_agent/constants.bal.
+const (
+	msgTypeFeedback             = "feedback"
+	msgTypeTokenIncreaseRequest = "token_increase_request"
+)
+
+// requestedByField names the actor on a token-increase request. It is always
+// rewritten server-side from the authenticated session — never trusted from the
+// browser — because it lands in a durable audit trail. See handleSideChannel.
+const requestedByField = "requestedBy"
 
 // wsSubprotocol is the WebSocket subprotocol this endpoint negotiates. The
 // frontend offers it (see WS_CUSTOMER_PORTAL in the webapp's apiConstants.ts)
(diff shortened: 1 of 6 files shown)


shayanmalinda:
🤖 **Claude review complete** — 3 comment(s) posted ([view review](https://github.com/wso2-open-operations/cs-tools/pull/1497#pullrequestreview-4964461822))

🟠 Medium: 1 · 🟡 Minor: 2

apps/customer-portal/backend-v2/internal/handler/websocket_test.go
```diff
@@ -270,6 +270,14 @@ type stubEntity struct {
...
 	err     error
 	calls   int
+
+	// conversation/conversationErr script GetConversation, which handleMessage
+	// consults before downgrading a CONVERTED conversation to RESOLVED.
+	conversation    entity.ConversationDetails
+	conversationErr error
+	// updatedStates records every state UpdateConversation was asked to set, so
+	// a test can assert the transition was skipped rather than merely reordered.
+	updatedStates []string
```
shayanmalinda, line 280: **🟠 Medium: This scaffolding is written but nothing reads it**

`updatedStates`, `conversation` and `conversationErr` are all populated by the stub, and the comment says they're here so a test can assert the transition was skipped rather than merely reordered — but no test asserts any of them, so the CONVERTED-outranks-RESOLVED rule ships uncovered. The stub is already doing the recording, so the test is a few lines: set `conversation.State` to `"CONVERTED"`, drive `handleMessage` with a resolved payload, and check `updatedStates` stayed empty.

> 📖 **Reference ([dead-code](https://refactoring.guru/smells/dead-code)):** "Unused functions, exports, params, branches, or orphaned writes (a key written but never read). Delete it."

---

apps/customer-portal/backend-v2/internal/handler/websocket.go
```diff
@@ -399,12 +432,93 @@ func (h *WebSocketHandler) handleMessage(ctx context.Context, conn *websocket.Co
...
+	if conversationID == "" || !uuidRe.MatchString(conversationID) {
+		// The client knows which answer it is rating even when this connection
+		// has not carried a turn yet, so a missing id is a client bug, not a
+		// resumable state. Dropped rather than surfaced: a failed rating must not
+		// interrupt the chat.
+		slog.ErrorContext(ctx, "discarding side-channel message: no usable conversation id", "userID", user.UserID, "msgType", msgType)
+		return
+	}
+
+	payload, dropped, err := stampRequestedBy(parsed, user.Email)
```
shayanmalinda, line 500: **🟡 Minor: `requestedBy` is stamped onto ratings too**

`stampRequestedBy` runs for both side-channel types, so every feedback payload now gains a `requestedBy` field — but the constant's own comment a few lines up says it "names the actor on a token-increase request". Either gate the stamping on `msgType == msgTypeTokenIncreaseRequest`, or widen that comment to say it's stamped on every side-channel message.

---

apps/customer-portal/backend-v2/internal/handler/websocket_side_channel_test.go
```diff
@@ -0,0 +1,135 @@
...
+	"testing"
+)
+
+// TestStampRequestedBy_UsesTheSessionNotTheBrowser is the security assertion for
+// the token-increase audit trail: requestedBy becomes the authenticated user's
+// email regardless of what the browser claimed.
+func TestStampRequestedBy_UsesTheSessionNotTheBrowser(t *testing.T) {
+	parsed := map[string]any{
+		"type":           msgTypeTokenIncreaseRequest,
+		"conversationId": "11111111-1111-1111-1111-111111111111",
```
shayanmalinda, line 30: **🟡 Minor: The dispatch that fixes the stall isn't covered**

These tests cover the two pure helpers well, but nothing asserts the actual fix — that a `feedback` or `token_increase_request` message reaches `SendSideChannel` and never `StreamChat`. That's the branch whose absence caused the five-minute connection stall, and it's the one most likely to be lost in a future refactor of `handleMessage`; a fake `wsStreamer` recording which method was called would pin it down.

---

shayanmalinda (commented):
Read the whole change: the new `SendSideChannel` client method, the dispatch in `handleMessage`, the `requestedBy` stamping, the CONVERTED guard, and the case-creation hook. Left 3 comments — 1 medium, 2 minor, all about test coverage or a comment that's drifted from the code. No correctness or security problems found.

Start with the one on `websocket_test.go`: the stub was extended to record exactly what the CONVERTED rule needs asserted, but the assertion never got written.

Two things I checked that are fine. Skipping the chat path's guards for side-channel messages doesn't open an authorization hole — that path only validates the conversation id's format too, and the real gate is the project-access check at upgrade time. And `SendSideChannel` dialling its own upstream connection matches what `StreamChat` already does per call. Build, vet and tests are clean.

cloby99 (approved):

PR #1512, opened 2026-08-19
[CSM] Add a Ballerina-to-Go response-field parity check

SasmithaDilshan:
Closes wso2-enterprise/digiops-cs#2812.

## Why

Every field gap this migration has hit was **invisible**. Ballerina's records are open (`json...;`); the `sn*` structs are closed by omission. A field ServiceNow sends that no Go struct declares is dropped by `encoding/json` with no error, and a screen simply renders empty.

`deployedProductCount` blanked Usage & Metrics. Ten project fields went missing. Several case-detail fields were lost. **Nothing failed to compile and no test noticed** — each was found by debugging staging.

This turns that class into a build failure.

## How

A frozen inventory of the Ballerina service's response fields — **420 names across 164 response types**, from digiops-cs `204778dea` — compared against the `sn_*.go` decode layer. Anything absent and not accounted for in `known_gaps.json` fails.

The inventory is frozen deliberately: Ballerina lives in another repository and is being decommissioned, so the reference had to be captured **while it still exists**. `extract_ballerina_fields.py` regenerates it from a checkout.

### Two layers, because one wasn't enough

A negative test is what exposed this, and it's the part worth reviewing:

1. **Scanning the whole module was useless.** Removing `json:"deployedProductCount"` from the sn struct — the exact regression that blanked Usage & Metrics — **did not fail**, because `domain/entity.go` still declared the name and a flat set counted it as understood.
2. **Scoping to the decode layer wasn't sufficient either.** `deployedProductCount` is decoded in *both* `sn_deployment_service.go` and `sn_project_stats_service.go`, so deleting either one alone still slipped through.

So `critical_fields.json` pins the decode **site** for fields already known to break a page. Both layers are verified by removing a tag, confirming the failure, and restoring:

```
A. drop deployedProductCount from ONE of its two sites
   --- FAIL: TestCriticalFieldsStayAtTheirDecodeSite
       ../service/sn_deployment_service.go no longer decodes: deployedProductCount
B. drop a field with a single site
   --- FAIL: TestBallerinaResponseFieldParity      ... decoded nowhere: onboardingStatus
   --- FAIL: TestCriticalFieldsStayAtTheirDecodeSite
```

## Relevance to the pending `main → dev` merge

That merge conflicts in **19 entity-service files**, including `sn_case_service.go`, `sn_project_service.go` and `domain/entity.go` — every field fix we have made. Resolving one by taking a side wholesale deletes struct tags, and *nothing else would tell you*. This is the guard for exactly that; CLAUDE.md now says to resolve those files by **union**.

## known_gaps.json

The 14 current gaps, each with a reason: four tracked under digiops-cs#2811, `fixEta` deliberately excluded as CSM-internal, five with no portal consumer, three assembled in the service layer rather than decoded, one likely name collision.

The tests also fail on a **stale** entry (implemented since) and on a missing reason, so the list cannot quietly become a permanent exemption.

Ballerina's `closedAt`/`openedAt`/`resolvedAt` are normalised to this service's `closedOn`/`openedOn`/`resolvedOn` — a documented convention here, so naming differences rather than gaps.

## A workflow had to come with it

**There was no CI for either Go module** — `go test` ran only when someone remembered. A guard against a regression that produces no compile error and no failing behaviour is worthless as a manual step. `go-checks.yml` runs gofmt, build, vet and `test -race` for entity-service and backend-v2 on any PR touching them.

## Testing

- `go build`, `go vet`, `gofmt -l`, `go test -race ./...` — all pass
- Both guard layers negative-tested as shown above
- `gosec` (with `GOTOOLCHAIN=auto`) reports the same single pre-existing G704 as before this change — none added


SasmithaDilshan:
Closing at the author's request — the parity check is not worth its maintenance cost for the coverage it gives.

Recording the reasoning so it does not get re-argued from scratch:

- It compares field **names** only, so it never covered the type/shape mismatches (the `createdBy` object-vs-string 500). Realistically about half of the gaps found during end-to-end testing.
- The module-wide comparison needed a hand-maintained `known_gaps.json`, and the per-file guard needed `critical_fields.json` — two lists to keep honest.

The Go CI workflow is closed with it, so `gofmt`/`build`/`vet`/`test -race` remain unautomated for entity-service and backend-v2; Go PRs continue to rely on local runs.

**One thing that does not come back:** the frozen inventory (420 response fields across 164 response types, from digiops-cs `204778dea`) was captured because the Ballerina service is being decommissioned. Once that repository is gone it cannot be regenerated. It stays retrievable in this PR's diff — `entity-service/internal/parity/ballerina_response_fields.json` — and the branch is left in place rather than deleted, so nothing is lost if it is ever wanted.

PR #1513, opened 2026-08-19
[Customer Portal v2] Send project type as an {id, label} ref

SasmithaDilshan:
## Why the Operations nav item was missing

`SideBar.tsx` filters it out unless `isProjectTypeResolved`:

```js
const projectTypeLabel = selectedProject?.type?.label ?? projectDetails?.type?.label;
const isProjectTypeResolved = projectTypeLabel != null && String(projectTypeLabel).trim() !== "";
```

backend-v2 sent the project type only as a plain `subscriptionType` string, so `type` was `undefined` on **both** the search item and the detail, and the item was always dropped.

Same shape of bug as the case status/severity labels: the value was present, under a key and shape the frontend does not read. `hasSR`/`hasCR` were fine — those come from `/features` and are correct.

## The nav item was the visible symptom, not the whole problem

Those label strings are compared **for equality** against the frontend's `ProjectType` enum (`src/types/permission.ts`) and drive behaviour:

- **Development Support** → case severity locked to S4
- **Cloud Support** / **Cloud Evaluation Support** → auto-pick the primary product on case and service-request creation
- **anything other than Managed Cloud Subscription** → S0 (Catastrophic) filtered out of the severity options

With `type` absent, every one of those silently took its *"not that type"* branch — so severity options and product defaults have been wrong for those project types, not merely a nav item hidden.

A wrong label would be worse than a missing one: it would silently change which severities a customer can choose. So the labels are **pinned byte-for-byte against `permission.ts` by a test**, not derived. entity-service builds its enum mechanically from the ServiceNow name (lowercase, spaces → underscores), but reversing that transform would mangle any future value containing an acronym, so an explicit table is used.

## Design notes

- **`id`** carries the enum value. entity-service normalises the type and exposes no numeric choice-list id; nothing in the portal reads project `type.id` (only *deployment* `type.id`), and a stable value beats inventing a fake number.
- **`subscriptionType` is kept** alongside `type`, so anything already reading it is unaffected.
- **An empty upstream value omits the object** rather than sending a blank one, preserving the frontend's `??` fallback from the search item to the detail.

## Testing

- `go build`, `go vet`, `gofmt -l`, `go test -race ./...` — all pass
- `gosec -fmt=text ./...` (with `GOTOOLCHAIN=auto`) — **0 issues** (104 files)
- Tests pin all seven labels against the frontend enum, cover the empty and unrecognised cases, and assert the emitted **object shape** on both the detail and the search item — the shape is what the nav gate depends on, not just the value
- `openapi.yaml` updated for both schemas; I initially referenced a non-existent `IDLabelRef` schema and corrected it to the existing `IdLabelRef`, then verified the spec has **no dangling `$ref`s** anywhere

### UI verification

Open a project: the **Operations** item appears in the side nav. On a Development Support project, case creation should offer only S4; on a non-Managed-Cloud-Subscription project, S0 should not be offered.

## Summary by CodeRabbit

* **New Features**
  * Project summaries and details now include a `type` field with a standardized ID and display label.
  * Recognized project types show consistent frontend-friendly labels.
  * Unknown project types remain available, while empty values are omitted.
  * Existing subscription type information remains supported.

* **Documentation**
  * Updated the API specification to document the new project type field.

### DIFF
--- a/apps/customer-portal/backend-v2/internal/dto/project.go
+++ b/apps/customer-portal/backend-v2/internal/dto/project.go
@@ -29,14 +29,17 @@ import (
 // complianceViolationDate and suspensionProcessState — these are WSO2-internal
 // closure/compliance workflow fields not meant for customer display.
 type ProjectSummary struct {
-	ID               string     `json:"id"`
-	Name             string     `json:"name"`
-	Key              string     `json:"key"`
-	SubscriptionType string     `json:"subscriptionType"`
-	StartDate        *time.Time `json:"startDate,omitempty"`
-	EndDate          *time.Time `json:"endDate,omitempty"`
-	CreatedOn        time.Time  `json:"createdOn"`
-	ClosureState     *string    `json:"closureState,omitempty"`
+	ID               string `json:"id"`
+	Name             string `json:"name"`
+	Key              string `json:"key"`
+	SubscriptionType string `json:"subscriptionType"`
+	// Type is what the frontend actually reads (project.type.label); see
+	// projectTypeRef. subscriptionType is kept for existing consumers.
+	Type         *IDLabelRef `json:"type,omitempty"`
+	StartDate    *time.Time  `json:"startDate,omitempty"`
+	EndDate      *time.Time  `json:"endDate,omitempty"`
+	CreatedOn    time.Time   `json:"createdOn"`
+	ClosureState *string     `json:"closureState,omitempty"`
 	// No omitempty: the frontend's ProjectListItem types activeCasesCount as a
 	// required number, so a project with no active cases must still send 0.
 	ActiveCasesCount int `json:"activeCasesCount"`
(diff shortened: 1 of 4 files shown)


SasmithaDilshan:
Rebased onto `dev-app-csm-portal` to clear the conflict with #1511, which merged in between. Both PRs added properties to `ProjectSummary`/`ProjectDetails` and to the same two OpenAPI schemas.

Resolved as a **union**, keeping both sets:

- `ProjectSummary`: #1511's `activeCasesCount` **and** this PR's `type`
- `ProjectDetails`: #1511's ten engagement fields **and** this PR's `type`
- `openapi.yaml`: both property sets under both schemas, verified with no dangling `$ref`s

Worth noting this is precisely the resolution hazard #1512 is built to catch: taking one side wholesale here would have silently deleted either the onboarding fields or the type ref, with **nothing failing to compile and no test noticing**.

Re-validated after the rebase with both feature sets together: build, vet, gofmt, `go test -race` all pass (8 project tests across both features), gosec 0 issues.

cloby99 (approved):

PR #1525, opened 2026-08-20
[CSM][Web] Fix dashboard notIn filters click-through sign flip; add Team bar control

Rashmika998:
## Summary

**Bug fix** — reported live: clicking a dashboard widget's "Work In Progress" pie slice (from a widget whose base query includes `projectOnboardingStatus notIn ["In-Progress"]`) landed on the Cases list showing an "Onboarding: In-progress" **INCLUDE** chip/filter — the exact opposite of the widget's own filter.

Root cause: `widgetResourceConfig.ts`'s `translateCaseDashboardFilters` read `projectOnboardingStatus`'s values op-blind (ignoring whether the widget said `in` or `notIn`) and always wrote them into the same "include" slot. Per the backend's `POST /cases/search` contract, `state`, `tag`, and `projectOnboardingStatus` are the only 3 case-search fields that accept `notIn` at all — `tag` was already handled correctly (via `tags`/`excludeTags`); `state` and `projectOnboardingStatus` had the identical latent bug (state`s just hadn't been reported yet). Both are now read op-aware, mirroring `tag`'s existing `in`/`notIn` → two-distinct-fields pattern (`states`/`excludeStates`, `onboardingStatuses`/`excludeOnboardingStatuses`), plumbed through the URL codec and the `/cases/search` payload builder.

Added a "Hide onboarding in progress" checkbox to the filter bar (same simple pattern as `IncidentsFilterBar`'s "SLA violated" checkbox) so the one real-world exclude value has more than just a chip.

**UI change** — removed the "Work state" bar control (narrow, rarely hand-picked — `workStates` still round-trips losslessly via the URL/search payload, now surfaced as a removable chip instead) and replaced its slot with a "Team" control (CS team, by `creGroupId`), restoring a control that had been deliberately removed earlier as bar clutter.

## Test plan
- [x] `tsc -b` — clean
- [x] `eslint .` — 0 errors (2 pre-existing warnings, unrelated files)
- [x] Full `vitest run` — **179/179 test files, 1767/1767 tests pass**, including new regression tests reproducing the exact reported bug (`projectOnboardingStatus notIn` decoding to `excludeOnboardingStatuses`, never `onboardingStatuses`) at the translation, URL-codec, search-payload, and chip-rendering layers, plus new tests for the "Team" control and the "Hide onboarding in progress" checkbox

### DIFF
--- a/apps/csm-portal/webapp/src/features/csm-cases/components/CasesFilterBar.test.tsx
+++ b/apps/csm-portal/webapp/src/features/csm-cases/components/CasesFilterBar.test.tsx
@@ -167,72 +167,187 @@ describe("CasesFilterBar — removed bar controls fall back to chips", () => {
 
 
   /**
-   * The CS-team and tag bar controls were removed as clutter, so a chip is now
-   * the ONLY way these filters are visible or clearable after a dashboard
-   * click-through. If these break, a user lands on a filtered list with no way
-   * to see or undo why.
+   * The tag bar controls (and, briefly, CS team's) were removed as clutter,
+   * so a chip is the ONLY way these filters are visible or clearable after a
+   * dashboard click-through. If these break, a user lands on a filtered
+   * list with no way to see or undo why. CS team has its own "Team" bar
+   * control again (see the describe block below) and is deliberately NOT
+   * chipped, to avoid showing the same selection twice.
    */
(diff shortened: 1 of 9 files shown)


Rashmika998:
Superseded — these changes were merged into #1523 instead.

PR #1532, opened 2026-08-21
feat(entity-service): conversation counts grouped by state

SasmithaDilshan:
## What

`POST /conversations/stats/search` — how many of the matching conversations sit in each state.

```jsonc
// request — same filters as /conversations/search, minus states
{ "filters": { "projectIds": ["…"], "createdBy": ["…"], "searchQuery": "" } }

// response
{ "totalCount": 412,
  "stateCounts": [ {"state":"ACTIVE","count":31},   {"state":"RESOLVED","count":250},
                   {"state":"CONVERTED","count":88}, {"state":"ABANDONED","count":43},
                   {"state":"CLOSED","count":0} ] }
```

## Why

The Novera analytics dashboard needs deflection and abandonment rates across the estate. Today it calls `GET /projects/{id}/conversations/stats` **once per project** and sums the results, so the cost grows with the project list and one slow project holds up the whole figure. This collapses it to a single request.

## How

Built over the existing `POST /conversations/search` rather than a new upstream endpoint: one probe per state, issued concurrently, asking for the smallest page and reading `totalRecords`.

Counts are exact — the page size bounds the **rows returned**, not the total, so a state with 4,000 conversations still reports 4,000. Five upstream requests regardless of how many conversations or projects are involved.

**This needs no change in the ServiceNow integration service**, which is why it can ship independently. A native aggregate (one grouped query instead of five probes) would be better, but that has to be exposed by the integration first — the contract here would not change, only its implementation.

## Two decisions worth reviewing

**Failure is all-or-nothing.** `errgroup`'s derived context cancels the siblings as soon as one state fails, so a partial tally is never returned. A caller cannot distinguish a partial count from a complete one, and a wrong deflection rate is worse than an error — the caller can retry, but it cannot detect a silently-missing state.

**Zero rows are always present.** `stateCounts` carries one entry per state including zeros. A state absent from the list and a state with no conversations are different claims, and a caller computing a rate needs the zero rows to build a denominator it can trust.

`states` is deliberately absent from the filters — the response is grouped by state, so filtering on it would only ever zero the other rows.

## Not included

**No date filtering.** `startDate` / `endDate` would make conversation outcomes reportable over time, which is the thing consumers actually want and cannot currently get — these figures are a live snapshot, so today's number is unrecoverable tomorrow. Neither this endpoint nor `/conversations/search` can offer it until the integration service does; adding a parameter here that the upstream ignores would be worse than not having one.

## Verification

`go build ./...`, `go vet ./internal/...` and `gofmt -l` are all clean. `openapi.yaml` parses and the new path and four schemas resolve.

Not exercised against a live ServiceNow instance — the counts depend on the upstream honouring `stateKeys` and returning an accurate `totalRecords`, which the existing search already relies on, but a staging call against one project would confirm it before anyone builds a KPI on it.

## Summary by CodeRabbit

- **New Features**
  - Added an endpoint to retrieve conversation statistics grouped by state.
  - Supports existing conversation search filters, excluding state filtering.
  - Responses include counts for each state, including states with zero results, plus an aggregate total.

### DIFF
--- a/entity-service/internal/domain/entity.go
+++ b/entity-service/internal/domain/entity.go
@@ -2495,7 +2495,7 @@ type ProjectContact struct {
 	// than treating it as an error. A pointer keeps "no id" distinguishable from an
 	// empty id internally; on the wire both a nil and an absent id are omitted, so the
 	// published shape is unchanged.
-	ID                   *string  `json:"id,omitempty"`
+	ID *string `json:"id,omitempty"`
 	// Name is nil when the row has no contact record linked -- the name is only ever
 	// known from that record.
 	Name *string `json:"name"`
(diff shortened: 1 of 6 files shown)


SasmithaDilshan:
**Do not merge as-is — this has a counting bug, found after the PR was raised.**

`conversationStatsOrder` enumerates five states and sums their counts into `totalCount`:

```go
ConversationStateActive, ConversationStateResolved, ConversationStateConverted,
ConversationStateAbandoned, ConversationStateClosed
```

Those map to SN state keys 2–6. Measured against the live instance, there is also **state key 1, "Open"**, and it is the largest single state:

| key | label | count |
|---:|---|---:|
| **1** | **Open** | **600** |
| 2 | Active | 434 |
| 3 | Resolved | 9 |
| 4 | Converted | 172 |
| 5 | Abandoned | 583 |
| 6 | Close | 9 |
| | **total** | **1807** |

So this endpoint would report `totalCount: 1207` — a third of all conversations missing — and any rate a caller computes against that total is inflated accordingly. The domain enum has no `OPEN` member, so `snConversationStateKeyMap` cannot even express key 1.

The same mistake was in the Novera chatbot's own mapping (wso2-enterprise/digiops-cs#2887) and is fixed there by folding Open into Active: both mean the conversation is in flight with no outcome yet, so splitting them adds a bucket without adding an answer.

**There is also a better implementation now.** wso2-enterprise/digiops-cs#2886 adds `POST /conversations/stats` to the Ballerina entity-service, using ServiceNow's Aggregate API — one grouped query rather than the five probes here, and it supports `startDate`/`endDate`, which this cannot offer at all. Date filtering is the capability that actually matters: conversation outcomes are currently a live snapshot, so today's number is unrecoverable tomorrow.

Two options, and I would like a steer before spending more on it:

1. **Close this**, and have the Go service pass through to the Ballerina endpoint once #2886 lands. The response contract here does not change, only who computes it.
2. **Keep it as the fallback** — in which case it needs an `OPEN` enum member, a key-1 entry, and the fold into Active, so it agrees with the other two layers.

I would take (1): the five-probe approach was designed to avoid touching the Ballerina layer, and that constraint no longer applies.

PR #1550, opened 2026-08-24
[Entity Service] Case SLA-breach/account-escalation filters + incident madeSla filter

rksk:
A CSM dashboard rebuild effort needs to numerically match a live ServiceNow reference dashboard. Two of its tiles ("SLA Violations", "Escalated Cases" on the case side, and an SRE incident-SLA tile) currently use approximate signals that diverge from what the SN tiles actually count. The corresponding ServiceNow and Ballerina entity-service layers now expose the exact underlying signals; this service does not yet surface them.

Resolves: N/A -- tracked internally.

- Expose `slaBreached` (case search): whether a case has a currently-breached SLA against a specific named-SLA-definition set.
- Expose `accountEscalationActive` (case search): whether the case's account currently has an active escalation record -- a different, account-level mechanism from the existing case-level `escalation`/`isEscalated` filter.
- Expose `madeSla` (incident search): ServiceNow's raw `made_sla` field, kept deliberately separate from the existing, more reliable `slaViolated` filter.

**Two independent, additive filters, bundled into one PR (same effort, landed same session).**

**`slaBreached` / `accountEscalationActive`** -- added as `*bool` fields on the case-search filter domain type, mirroring the existing `escalation`/`isEscalated` field's shape. Neither fits the existing isEmpty/isNotEmpty reference-presence pattern `escalation` uses, so a new `parseCaseFilterBool` helper was added (mirroring the incident side's equivalent, for consistency across resource types). Both are explicitly rejected on the Postgres-backed path and inside `anyOf` filter groups, matching the existing pattern for other ServiceNow-only case filters (`escalation`, `taskSLABusinessElapsedPercent`) -- these are ServiceNow-only capabilities and are not silently allowed to leak into paths that can't support them. `accountEscalationActive`'s doc comment explicitly distinguishes it from the case-level `escalation` field so a future reader doesn't conflate the two different mechanisms. Wire payload uses `*bool` (not plain `bool`) so an explicit `false` still reaches ServiceNow, matching the existing `IsEscalated` precedent.

**`madeSla`** -- added as a direct `*bool` sibling of the existing `SlaViolated` field at all three of its call sites (domain filter struct, allowed-field parser, outbound ServiceNow payload builder for both search and aggregate). Doc comment explicitly cross-references `slaViolated` and states `madeSla` is the less-reliable raw signal, kept only for exact ServiceNow-dashboard-parity purposes -- so nobody mistakes it for a newer/better version of `slaViolated`.

`openapi.yaml` updated for both.

- As a CSM dashboard widget, I can filter cases by whether they have a breached SLA against the same named-SLA-definition set ServiceNow's own dashboard uses.
- As a CSM dashboard widget, I can filter cases by account-level escalation state, distinct from case-level escalation.
- As a CSM dashboard widget, I can filter incidents by ServiceNow's raw SLA-met field for exact dashboard parity, without disturbing the existing, more reliable `slaViolated` filter.

Entity service: new case-search filters `slaBreached` and `accountEscalationActive`; new incident-search filter `madeSla`.

`openapi.yaml` updated with all three new filter fields and their descriptions, including the distinguishing notes vs. their similarly-named siblings.

## Training
N/A -- no training content covers this internal service.

## Certification
N/A -- no certification exam covers this internal service.

## Marketing
N/A -- internal infrastructure.

## Automation tests
 - Unit tests
   > Code coverage information

`go build ./...`, `go vet ./...`, `go test ./...` all pass clean from the `entity-service` module root. New tests: case-side field translation (true/false, and a check that `accountEscalationActive` doesn't leak into the existing `escalation`/`HasActiveEscalation` field), 6 new rejection cases (bad op / non-boolean / multi-value, one pair per new field), an `anyOf`-group rejection test, a Postgres-path rejection test, and a wire-level test pinning the exact JSON keys and proving an explicit `false` is transmitted rather than dropped. Incident-side: parse tests (true/false + 3 rejection cases) and a passthrough test proving `madeSla`/`slaViolated` are independent.

 - Integration tests

No automated integration suite covers these paths (consistent with this service's existing convention). A real end-to-end run against ServiceNow DEV was not performed for this PR: the corresponding Ballerina entity-service change (a separate PR, not yet merged) is a prerequisite for a genuine round-trip test, and a peer development session was concurrently holding this environment's local-stack ports during this work. This is disclosed here rather than overstated; a live smoke test against DEV is recommended once the Ballerina PR merges, before this is considered fully proven end-to-end.

## Security checks
 - Ran FindSecurityBugs plugin and verified report? N/A -- `gosec` was not available in this environment for this PR; no security-sensitive surface is touched (three new boolean filter fields following existing patterns), but this should still be run before merge per this service's normal convention.

## Samples
N/A

A corresponding ServiceNow scripted-API change and a corresponding Ballerina entity-service change are tracked separately (not in this repo / a separate PR in the private entity-service repo).

## Migrations (if applicable)
N/A -- no schema/data migration involved.

Go entity-service, macOS, local build/test only (see Automation tests above for the honest scope of verification).

Two of the three new fields don't fit either of this service's existing boolean-filter idioms cleanly (isEmpty/isNotEmpty for reference-presence checks like `escalation`) -- rather than force a mismatched pattern, a small `parseCaseFilterBool` helper was added, mirroring the incident side's existing equivalent, so plain-boolean filters have one consistent idiom across resource types.

### DIFF
--- a/entity-service/internal/domain/entity.go
+++ b/entity-service/internal/domain/entity.go
@@ -3549,6 +3549,13 @@ type SearchIncidentsFilters struct {
 	//     case search's own "createdOn" filter.
 	//   - "slaViolated" (op eq): a single boolean value; restricts to
 	//     incidents with (or without) at least one breached SLA record.
+	//   - "madeSla" (op eq): a single boolean value; restricts to incidents
+	//     matching ServiceNow's raw `made_sla` field. Deliberately kept
+	//     separate from "slaViolated" above: "slaViolated" is derived from
+	//     task_sla.has_breached and is the reliable signal; "madeSla" is
+	//     ServiceNow's own less-reliable raw field, kept only for exact
+	//     parity with SN's native incident dashboards. Prefer "slaViolated"
+	//     unless dashboard parity is the explicit goal.
 	//   - "productName" (op in): one or more product names, matched as a
 	//     union against the incident's backing business_service name.
 	// See service.ParseIncidentFieldFilters.
--- a/entity-service/internal/domain/entity.go
+++ b/entity-service/internal/domain/entity.go
@@ -1448,6 +1448,22 @@ type ParsedCaseFilters struct {
 	// escalation, from the "escalation" filter field's isEmpty/isNotEmpty op
 	// (optional; nil means no filter on this field).
 	HasActiveEscalation *bool
+	// HasBreachedSLA filters cases to those with a currently-breached SLA
+	// against the 10 named SLA definitions, from the "slaBreached" filter
+	// field's eq value (optional; nil means no filter on this field). Wire
+	// field on the SN payload: "slaBreached". Requires ServiceNow data source.
+	HasBreachedSLA *bool
+	// HasActiveAccountEscalation filters cases to those whose parent ACCOUNT
+	// has an active escalation (active_account_escalation.state IN 100,101),
+	// from the "accountEscalationActive" filter field's eq value (optional;
+	// nil means no filter on this field). Wire field on the SN payload:
+	// "accountEscalationActive". Requires ServiceNow data source.
+	//
+	// Distinct from HasActiveEscalation above: HasActiveEscalation filters by
+	// whether the CASE ITSELF carries an active escalation; this field
+	// filters by whether the case's parent ACCOUNT has one, regardless of
+	// whether this specific case is escalated.
+	HasActiveAccountEscalation *bool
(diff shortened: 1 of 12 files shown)


rksk:
Closing — bundled into #1548 instead, per request. Same two commits (slaBreached/accountEscalationActive + madeSla), now on that PR's branch.

PR #1579, opened 2026-08-26
[CSM Portal] Persist recent time-card approvers in localStorage

rksk:
Opening the "Log time" dialog re-derives the signed-in engineer's previously-selected
approvers from a backend query every time, even though that list rarely changes between
opens. A corresponding entity-service change (tracked separately) was also needed for the
underlying sort-by-date query this feature depends on to succeed against the live backend.

Cache the recent-approvers result in the browser across dialog opens and page reloads, so
it isn't re-queried on every single "Log time" open.

Added a versioned, per-signed-in-engineer `localStorage` cache
(`csm-portal:time-cards:recent-approvers:v1:<engineerId>`) to `useRecentApprovers`
(`useTimeSheets.ts`). A fresh (<24h) cache entry is fed back as the query's `initialData`
and the query is disabled entirely on a hit — no network call. Every real fetch re-persists
its result. The cache is dropped on any time-card create/decide/bulk-approve via the
existing `invalidateTimecards` hub (already called from those mutations). All reads/writes
are wrapped in try/catch so private browsing or a storage error falls back to the existing
always-fetch behavior rather than breaking the dialog. The separate, live per-open
eligibility re-check (approver role/active status via `/users/search`) is untouched and
still runs on every open, since that can drift independently of the engineer's own approver
history.

As a CS engineer logging time repeatedly against different cases, I don't want to trigger a
network query every time I open the dialog just to see the same recent approvers I already
picked before.

The "Log time" dialog's recently-selected approvers are now cached in the browser, avoiding
a repeat backend query on every open.

N/A — no visible workflow change to the `/help` guide; the recent-approvers feature itself
already exists and is already documented (if at all) — this only changes its performance
characteristics, not its behavior.

## Training
N/A — no training-content impact.

## Certification
N/A — no certification-exam impact.

## Marketing
N/A — internal engineer-portal performance improvement, no marketing surface.

## Automation tests
 - Unit tests
   > 19 new cases added to `useTimeSheets.test.tsx`: fresh-cache read with no network call,
   fetch+write on cache miss, TTL-expired entry refetches, malformed/garbage JSON falls back
   to refetch without crashing, `localStorage` throwing (private browsing) falls back to
   fetch, cache invalidation on mutation. Full `csm-timecards` feature suite: 102/102 pass.
 - Integration tests
   > None added — not applicable to a client-side caching change with no backend contract
   change on this side.

## Security checks
 - Ran FindSecurityBugs plugin and verified report? N/A — not a JVM/Java component

## Samples
N/A

## Related PRs
A corresponding entity-service change (tracked separately, not in this repo) was needed for
the underlying sort-by-date query this feature depends on to stop being rejected by the
backend.

## Migrations (if applicable)
N/A

## Test environment
`npx vitest run` and `npx tsc -b`, both clean/passing locally. Not yet screenshot-verified
in a running browser.

## Learning
N/A

### DIFF
--- a/apps/csm-portal/webapp/src/features/csm-timecards/api/useTimeSheets.test.tsx
+++ b/apps/csm-portal/webapp/src/features/csm-timecards/api/useTimeSheets.test.tsx
@@ -19,9 +19,12 @@ import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
 import { describe, expect, it, vi, beforeEach } from "vitest";
 import type { ReactNode } from "react";
 import {
+  clearPersistedRecentApprovers,
+  invalidateTimecards,
   mapTimeCard,
   searchTimeCards,
   useBulkApproveCards,
+  useRecentApprovers,
 } from "@features/csm-timecards/api/useTimeSheets";
 import { BackendApiError, type BackendApi } from "@api/backend/client";
 import type {
@@ -40,6 +43,7 @@ import type {
 // those tests — only the useBulkApproveCards tests further down actually
 // exercise this mocked patch.
 const patchMock = vi.fn();
+const postMock = vi.fn();
 vi.mock("@api/backend/client", () => ({
   BackendApiError: class BackendApiError extends Error {
     status: number;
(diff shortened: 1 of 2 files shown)


rksk:
Superseded — folded into #1580 (bundled with this session's other cs-tools change).

PR #1581, opened 2026-08-26
[Customer Portal v2] Batch bulk product-vulnerability fetches

SasmithaDilshan:
Fixes `400 {"message":"limit cannot exceed 50"}` on `POST /products/vulnerabilities/search`.

## Cause

The frontend asks for the whole advisory set in one call — `PRODUCT_VULNERABILITIES_ALL_FETCH_LIMIT = 5000` — and the Ballerina backend answers it with all **1265** records:

```json
{ "limit": 5000, "offset": 0, "totalRecords": 1265, "productVulnerabilities": [ … ] }
```

entity-service rejects any limit above 50, and its own comment says why:

> `maxLimit` is 50 **because the backing data source rejects anything above 50** with an opaque validation error.

So the cap **cannot** be raised upstream, and the request cannot be forwarded as sent. It has to be split — which is exactly what the Ballerina backend already does (`service.bal:57`: *"In-memory cache for bulk product-vulnerability fetches (limit > 50)"*).

## Approach — ported from the Ballerina implementation

One count call to size the work → sequential batches of 50 → aggregate cached 12h, keyed on the filter combination → local slice to the caller's offset/limit. The response echoes the caller's own `limit`/`offset` and reports the true total, so a client that paginates keeps working while one that asked for everything gets everything.

**Requests at or below 50 pass straight through, unchanged.**

## One deliberate difference: authorization

Ballerina reads its **shared** cache *before* contacting entity-service at all. A warm cache therefore serves data without re-checking access — a caller who has since lost permission still gets results.

Here the count call runs on **every** request, before any cached data is returned. It sizes the batch loop and re-authorizes at the same time. A cache hit costs **1** upstream call instead of 27, and still 403s when it should. There is a test for precisely that:

```
TestFetchAllVulnerabilities_AuthorizesOnEveryRequest
  → warms the cache, then asserts a 403 caller is still rejected
```

The cache is shared rather than per-user because the vulnerability list is an org-wide published advisory set, identical for every caller — and with authorization no longer delegated to it, sharing is safe. It is bounded to 20 filter combinations, and an entry is discarded when the upstream total disagrees with it, since that means the set moved regardless of TTL.

## Timeout

Twenty-six sequential upstream calls can outlast the server's global `WriteTimeout` even when each is fast, so this path extends its own write deadline — the same structural reason the license-provisioning flow does.

## Testing

- `gofmt -l`, `go build`, `go vet`, `go test -race ./...` — all pass
- `gosec -fmt=text ./...` — **0 issues** (110 files)

Nine tests, including a stub that **enforces the real 50 cap** so a batching mistake fails the same way production did:

| Test | |
|---|---|
| batching | 1265 records via 1 count + 26 pages, no call above the cap |
| authorization | warm cache still 403s |
| cache hit | second request costs 1 call, not 27 |
| staleness | changed upstream total discards the entry |
| empty corpus | count call only, no batch loop |
| slicing | whole corpus, exact page, overrun, offset past end, negative offset |
| key separation, TTL expiry, capacity bound | |

> Validation ran on a local Go 1.27 toolchain with locally installed gosec — Docker Desktop is not running in this environment.

## Deployment

**backend-v2 only** — no entity-service change.

## Summary by CodeRabbit

* **New Features**
  * Expanded vulnerability searches beyond upstream record limits through batched retrieval.
  * Added pagination and total-result counts for large searches.
  * Improved response times for repeated searches with short-term result caching.
  * Preserved filtering and authorization across bulk searches.

* **Bug Fixes**
  * Improved handling of expired, empty, and changing vulnerability results.
  * Added safeguards for canceled requests and upstream errors.
  * Improved consistency when multiple searches request the same uncached results.
  * Added bounded caching to maintain reliable performance over time.

### DIFF
--- a/apps/customer-portal/backend-v2/internal/handler/product_vulnerability_bulk.go
+++ b/apps/customer-portal/backend-v2/internal/handler/product_vulnerability_bulk.go
@@ -69,10 +69,52 @@ type vulnerabilityCacheEntry struct {
 type vulnerabilityBulkCache struct {
 	mu      sync.Mutex
 	entries map[string]vulnerabilityCacheEntry
+	// inflight coalesces concurrent cold-cache fetches for the same filter
+	// combination. Without it, N callers arriving together after a restart or a
+	// TTL expiry each run the whole batch loop — N x 26 upstream calls, which is
+	// the load this cache exists to prevent, just arriving a different way.
+	inflight map[string]*vulnerabilityFetch
+}
+
+// vulnerabilityFetch is one in-progress aggregate fetch that later arrivals wait
+// on rather than duplicating.
+type vulnerabilityFetch struct {
+	done  chan struct{}
+	items []dto.ProductVulnerability
+	err   error
 }
 
 func newVulnerabilityBulkCache() *vulnerabilityBulkCache {
-	return &vulnerabilityBulkCache{entries: make(map[string]vulnerabilityCacheEntry)}
+	return &vulnerabilityBulkCache{
+		entries:  make(map[string]vulnerabilityCacheEntry),
+		inflight: make(map[string]*vulnerabilityFetch),
+	}
+}
+
+// lead registers the caller as the fetcher for key, or returns the fetch already
+// in progress for it.
+//
+// leader is true for exactly one caller per key at a time; that caller must call
+// finish. Everyone else receives the in-flight fetch and waits on it.
+func (c *vulnerabilityBulkCache) lead(key string) (f *vulnerabilityFetch, leader bool) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if existing, ok := c.inflight[key]; ok {
+		return existing, false
+	}
+	f = &vulnerabilityFetch{done: make(chan struct{})}
+	c.inflight[key] = f
+	return f, true
+}
+
+// finish publishes the result to any waiters and clears the in-flight slot.
+func (c *vulnerabilityBulkCache) finish(key string, f *vulnerabilityFetch, items []dto.ProductVulnerability, err error) {
+	c.mu.Lock()
+	delete(c.inflight, key)
+	c.mu.Unlock()
+
+	f.items, f.err = items, err
+	close(f.done)
 }
 
 // get returns the cached aggregate for key, if present and unexpired.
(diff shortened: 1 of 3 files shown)


shayanmalinda:
🤖 **Claude review complete** — 2 comment(s) posted ([view review](https://github.com/wso2-open-operations/cs-tools/pull/1581#pullrequestreview-5037161996))

🟡 Minor: 2

apps/customer-portal/backend-v2/internal/handler/product_vulnerability_bulk.go
```diff
@@ -0,0 +1,215 @@
...
+		mapped := dto.MapSearchProductVulnerabilities(page)
+		if len(mapped.ProductVulnerabilities) == 0 {
+			// The upstream returned fewer records than it claimed. Stop rather than
+			// loop forever against a total that will never be reached.
+			break
+		}
+		all = append(all, mapped.ProductVulnerabilities...)
+	}
+
+	cache.put(key, all, now)
```
shayanmalinda: **🟡 Minor: A short aggregate gets cached but can never be read back**

When the loop above breaks early because upstream returned fewer records than it claimed, `all` is shorter than `total`, and the read guard on line 188 requires `len(cached) == total` — so this entry is unusable the moment it's written, and it can evict a good entry from the 20-slot cache on the way in. Worth only caching a complete set, e.g. `if len(all) == total { cache.put(key, all, now) }`.

> 📖 **Reference ([dead-code](https://refactoring.guru/smells/dead-code)):** "Dead Code — unused functions, exports, params, branches, or orphaned writes (a key written but never read). Delete it."

---

apps/customer-portal/backend-v2/internal/handler/product_vulnerability_bulk.go
```diff
@@ -0,0 +1,215 @@
...
+
+	key := vulnerabilityCacheKey(req.Filters)
+	// Only trust the cache when it still agrees with the upstream count; a changed
+	// total means the advisory set moved and the entry is stale regardless of TTL.
+	if cached, ok := cache.get(key, now); ok && len(cached) == total {
+		return cached, nil
+	}
+
+	all := make([]dto.ProductVulnerability, 0, total)
+	for offset := 0; offset < total; offset += vulnerabilityUpstreamMaxLimit {
```
shayanmalinda, line 285: **🟡 Minor: Concurrent callers on a cold cache each run the whole batch loop**

Nothing coordinates two requests that miss the cache at the same moment, so ten users opening the security page after a restart or a TTL expiry means ten independent runs of ~26 sequential upstream calls. That's the load pattern this PR exists to avoid, just arriving a different way. A per-key `sync.Once` or `golang.org/x/sync/singleflight` around the fetch would collapse them into one — was that considered and left out deliberately?

---

shayanmalinda (commented):
Read through the batching path, the cache, and the tests. Two minor things inline — start with the `cache.put` one, since that entry can never be read back.

SasmithaDilshan:
Both addressed in `66148f02f`.

**Short aggregate cached but unreadable** — correct, and it was worse than dead weight: the entry could also evict a usable one from the 20-slot cache. Now only a complete set is cached (`len(all) == total`), while the partial result is still returned to the caller, since showing part of the list beats showing none. I moved the page walk into `fetchVulnerabilityPages` so the completeness check reads clearly at the call site rather than being buried after the loop.

**Concurrent cold-cache callers** — to answer directly: **no, it was not considered and left out.** It was an oversight, and your framing is right — it reintroduces exactly the load this PR exists to remove, just by a different route.

Fetches for the same filter combination now collapse into one: the first caller runs the batch, the rest wait on its result. I used a per-key in-flight map rather than `golang.org/x/sync/singleflight` because this backend is deliberately stdlib-only apart from `gorilla/websocket` (per its CLAUDE.md) and the coalescing is about thirty lines; happy to switch to `singleflight` if you would rather take the dependency.

Two things I was careful to preserve:

- **Authorization stays per-caller.** The count call runs for every request *before* any coalescing, so a shared fetch is never served to someone entity-service would reject. The existing `TestFetchAllVulnerabilities_AuthorizesOnEveryRequest` still passes.
- **A waiter honours its own context**, so a client that gives up does not block on another caller's fetch.

New tests: ten simultaneous cold callers collapsing to a single batch of ten pages (asserting the call count, not just the results), and an incomplete aggregate being returned but not cached. The stub is now mutex-guarded since the concurrency test drives it from several goroutines, and `go test -race` is clean.

Verified: build, vet, gofmt, `go test -race` all pass; gosec 0 issues across 110 files.

shayanmalinda (approved):

PR #1582, opened 2026-08-27
[Notification-Service] fix broken email logo, garbled non-ASCII names, and CTA styling

Rashmika998:
## Summary
- **Broken logo**: the inline base64 `data:` URI logo was being stripped by Gmail (and most webmail clients) as a security measure, regardless of correct encoding — not fixable by re-encoding. `cid:` inline attachments aren't an option either since the SMTP client behind the email-sending API only supports `Content-Disposition: attachment`, not inline/`Content-ID`. Switched to WSO2's own CDN-hosted logo URL (`wso2.cachefly.net`) with corrected `width`/`height` matching the real asset's aspect ratio.
- **Garbled names**: a display name with a non-ASCII trailing marker (e.g. `"Jane Doe Ⓦ"`) rendered as a literal `"?"` — the external email-sending service doesn't reliably preserve non-ASCII bytes through its own send path. `escapeHTML` now converts any non-ASCII rune to a numeric HTML character reference, which survives as ASCII and renders correctly.
- **CTA styling**: "Add Comment" changed from a solid button to a plain colored text link (matching "View Case") — the email can't actually submit a comment, only deep-link into the portal. Accent color changed from `#3b5bdb` to `rgb(71,96,146)` throughout.
- **`case.assigned` field rename**: `CaseAssignedPayload.AssignerName/AssignerEmail` → `AssigneeName/AssigneeEmail` (see events.go/validate.go) — a wire-format change. This is safe on its own: nothing in the monorepo currently publishes `case.assigned` (`csm-portal-backend` only declares the type constant, no `Publish` call site exists anywhere for it). It does pair with [#1583](https://github.com/wso2-open-operations/cs-tools/pull/1583), which makes entity-service the first-ever publisher, already using the new field names — no required merge order between the two, since there's no live producer using the old names to break.

## Test plan
- [x] `go build`/`go vet`/`gofmt -l` clean
- [x] `go test -race -count=1 ./...` clean, new tests for `escapeHTML`'s non-ASCII entity encoding
- [x] `gosec -fmt=text ./...` — 0 issues
- [x] `govulncheck ./...` — 0 vulnerabilities


shayanmalinda:
🤖 **Claude review complete** — 1 comment posted ([view review](https://github.com/wso2-open-operations/cs-tools/pull/1582#pullrequestreview-5037662383))

❓ Question: 1

integrations/csm-notification-service/internal/events/events.go
```diff
@@ -177,8 +177,8 @@ type StatusChangedPayload struct {
 // CaseAssignedPayload is TypeCaseAssigned's payload. See CaseCreatedPayload's
 // doc comment for why Recipients is here, and for why ProjectID is required.
 type CaseAssignedPayload struct {
-	AssignerName  string `json:"assignerName"`
-	AssignerEmail string `json:"assignerEmail"`
+	AssigneeName  string `json:"assigneeName"`
```
shayanmalinda, line 180: **🟡 Question: worth pinning the deploy order against #1583**

Renaming the JSON tags makes this a wire-format change, not just a Go-field rename. The consumer's `Validate` runs `decodeStrict`, which calls `DisallowUnknownFields`, so the two field names have to change on both sides in the same deploy — a payload carrying `assignerName` against the new struct, or `assigneeName` against the old one, is rejected outright and the event is dropped rather than partially decoded. Nothing publishes `case.assigned` today, so this side is safe to land first and #1583 must follow — but the reverse order, or a rollback of just one service, drops every assignment notification silently. Worth calling the ordering out in the description so whoever deploys knows the two go together.

---

shayanmalinda (commented):
Went through the escaping change carefully since that's the risky part, and it's correct — `html.EscapeString` runs first and the non-ASCII pass only touches runes above 127, so the `&` in `&amp;` is never double-escaped and astral-plane characters come out as valid references. Every call site moved to `escapeHTML`, the only remaining `html.EscapeString` is the one inside it, and every placeholder in all four templates still has a substitution. Builds and tests clean. One deployment note inline.

Small thing not worth its own comment: the CDN logo is a remote image, which Outlook desktop and similar clients block by default — the `alt="WSO2"` you kept means that degrades to text rather than a broken icon, so nothing to fix, just worth knowing the logo still won't render everywhere.

integrations/csm-notification-service/internal/events/events.go
Rashmika998: Checked this directly: nothing in the monorepo publishes `case.assigned` today — `csm-portal-backend` only declares the `TypeCaseAssigned` constant, with zero actual `Publish(...)` call sites anywhere, and `customer-portal-backend` has no publisher either. So there's no live producer sending the old `assignerName`/`assignerEmail` field names that this rename could break — worst case if #1583 lands out of order (or not at all) is just "still no assignment notifications," not a regression of a working feature.

Added a note to the PR description calling out the pairing with #1583 for anyone merging, per your suggestion. No code change needed here.

rksk (approved):

PR #1583, opened 2026-08-27
[Entity-Service] publish case.assigned on assignee change

Rashmika998:
## Summary
- `UpdateCase` now publishes `case.assigned` when `req.AssigneeEmail` is set and the assignee actually changes (same pre-PATCH `GetCaseByID` no-op guard as `case.status_changed`'s state comparison, to avoid a false notification on a re-PATCH of the current assignee).
- Unblocks a feature that was previously documented here as blocked on identity resolution ("who performed the assignment") — that was the wrong question: the new assignee's own email is already on the update request (`req.AssigneeEmail`), and ServiceNow's PATCH response resolves their display name (`assignedTo.name`), so no extra identity/auth work is needed.
- Matches `csm-notification-service`'s companion PR renaming `CaseAssignedPayload.AssignerName/AssignerEmail` → `AssigneeName/AssigneeEmail` to reflect what's actually being published.

## Test plan
- [x] `go build`/`go vet`/`gofmt -l` clean
- [x] `go test ./...` clean — 3 new tests: happy path, no-watchers skip, unchanged-assignee skip
- [x] `gosec` (via `GOTOOLCHAIN=auto`, per this repo's own convention) — 0 issues, 112 files scanned
- [x] `govulncheck ./...` — 0 vulnerabilities

### DIFF
--- a/entity-service/internal/service/sn_case_service.go
+++ b/entity-service/internal/service/sn_case_service.go
@@ -54,6 +54,11 @@ const publishCommentAddedTimeout = 5 * time.Second
 // Publish call afterward.
 const publishStatusChangedTimeout = 5 * time.Second
 
+// publishCaseAssignedTimeout bounds the same two things as
+// publishStatusChangedTimeout, for UpdateCase's assigneeEmail path instead
+// of its state path.
+const publishCaseAssignedTimeout = 5 * time.Second
+
 // watchListEmails extracts the non-empty emails from a case's watch list —
 // Recipients for every case.* event this file publishes is the case's
 // WatchList emails only (an explicit, deliberate decision — this service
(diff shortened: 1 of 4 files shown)


shayanmalinda:
🤖 **Claude review complete** — 1 comment posted ([view review](https://github.com/wso2-open-operations/cs-tools/pull/1583#pullrequestreview-5037662233))

❓ Question: 1

entity-service/internal/service/sn_case_service.go
```diff
@@ -1116,6 +1121,62 @@ func (s *snCaseService) publishStatusChanged(ctx context.Context, caseID, newSta
...
+	ctx, cancel := context.WithTimeout(ctx, publishCaseAssignedTimeout)
+	defer cancel()
+
+	recipients := watchListEmails(before.WatchList)
+	if len(recipients) == 0 {
+		slog.InfoContext(ctx, "sn update case: case.assigned not published, case has no watchers to email", "caseId", caseID)
+		return
+	}
+
+	payload, err := json.Marshal(events.CaseAssignedPayload{
```
shayanmalinda, line 1159: **🟡 Question: this can't deploy before the notification-service change**

`case.assigned` isn't published at all today, so this is the first producer — but it emits `assigneeName`/`assigneeEmail` while the deployed consumer still expects `assignerName`/`assignerEmail`. The consumer's `Validate` runs `decodeStrict`, which calls `DisallowUnknownFields`, so the two field names have to change on both sides in the same deploy — a payload carrying `assignerName` against the new struct, or `assigneeName` against the old one, is rejected outright and the event is dropped rather than partially decoded. So if this ships before #1582 the feature is silently dead on arrival: events publish, validation rejects every one, and no notification is sent. Worth stating the ordering in the description, or landing #1582 first.

---

shayanmalinda (commented):
Read the publish path against the existing `publishStatusChanged` it mirrors, and it holds up — the pre-PATCH `GetCaseByID` doubles as both the no-op guard and the enrichment source, the unchanged-assignee case is compared case-insensitively, a nil `AssignedEngineer` correctly falls through to publishing, and the enrich-failure path skips publishing exactly the way the status block does. Builds, vet and tests are clean. One deployment note inline — it's about ordering against #1582, not about this code.

entity-service/internal/service/sn_case_service.go
Rashmika998: Notification service related PR is already sent: https://github.com/wso2-open-operations/cs-tools/pull/1582.
Since this is a bug where assignee events are not configured to send emails and since this is still not in production yet it's fine to emit events.

rksk (approved):

PR #1618, opened 2026-08-30
[CSM Portal] add in-app tabbed view for opening multiple cases at once

rksk:
CS engineers reviewing multiple cases today have to rely on separate browser windows/tabs, with no in-app way to tell which browser tab holds which case and no way to switch between several open records without losing scroll position, an in-progress comment draft, or refetching data. This adds an in-app tabbed view — similar to a browser's own tab strip — so a CS engineer can have several cases, incidents, change requests, engagements, service requests, announcements, or security reports open at once inside the CSM portal itself.

- An in-app tab strip above the routed page content, with a permanent, non-closable "current location" tab plus up to 10 record tabs.
- Opened tabs stay genuinely mounted (not remounted) when switching between them, so scroll position, an open comment draft, and expanded sections are preserved.
- On by default, with a Preferences setting to turn it off and choose what happens when the tab limit is reached (replace the oldest tab or replace the last tab).
- Right-click support for closing all tabs or all-but-one, and a hover tooltip showing each tab's internal record id and subject.
- Existing ctrl/cmd/middle-click "open in a new browser tab" behavior is unaffected.

- A new `CaseTabsContext` (open tab list + reducer) and `CaseTabsBehaviorContext` (the on/off + cap-mode preference, persisted to `localStorage`) drive the feature; the open-tab list itself persists to `sessionStorage` (deliberately not `localStorage` — it's per-browser-session working state, not something that should follow a login across users on a shared machine), storing only each tab's record id and kind.
- Each record type's detail route (`/cases/:caseId`, `/operations/incidents/:id`, `/operations/change-requests/:id`, and four other case-like routes) is now wired through a small `CaseDetailRouteSync` component that opens/activates a tab for the matched route instead of mounting the page directly.
- Keep-alive rendering: every open tab's page component stays mounted (hidden via CSS) rather than being unmounted on switch. Since a single React Router instance can only resolve one route's params at a time, each tab gets its own lightweight `CaseRouteOverrideContext` (a plain React context, not a second `<Router>` — an earlier iteration tried nesting `<Router>` per tab and hit React Router's hard "no nested `<Router>`" invariant) so each mounted page resolves its own record id independent of the browser's real address bar.
- A consolidated `PreferencesDialog`, reachable from the profile-picture menu, replaces the previous standalone theme-only header control and now also hosts the tab-behavior setting.

As a CS engineer, I want to open several cases (or incidents/change requests) at once and switch between them without losing my place, so I can cross-reference records during a review pass without juggling separate browser windows.

Added an in-app tabbed view to the CSM portal, letting CS engineers open multiple cases, incidents, change requests, and related records as tabs within the app instead of separate browser windows. On by default; can be turned off from Preferences (profile menu).

The CSM portal's in-app `/help` user guide should get a short new topic describing the tab strip and the Preferences toggle, since this now ships on by default rather than as an undocumented opt-in. Tracked as a follow-up, not blocking this PR.

## Training
N/A — no external training content exists for this portal.

## Certification
N/A — no certification exam covers this portal.

## Marketing
N/A — internal tool, no marketing content.

## Automation tests
 - Unit tests
   > New test files for the tab-state reducer/context, the behavior/preference context, route-pattern matching, the tab strip component (including its context menus and tooltip), the isolated per-tab router, the close-confirm hook, and the relocated Preferences dialog/profile menu — all under `apps/csm-portal/webapp/src/{context/case-tabs,features/case-tabs,components/header}`.
 - Integration tests
   > `CaseTabsIntegration.test.tsx` mounts the mechanism inside a real `BrowserRouter` (matching how `App.tsx`/`AppLayout` actually nest it) and covers: opening multiple tabs, the default-on behavior on a fresh session, an explicit opt-out still rendering plain full-page navigation, and the tab-limit replace-oldest/replace-last behaviors with correct content shown (not a stale tab).

## Security checks
 - Ran FindSecurityBugs plugin and verified report? N/A — TypeScript/React frontend, not a Java/Go target for FindSecurityBugs; `eslint` and `tsc` ran clean (see Automation tests / CI)

## Samples
N/A — not a library/SDK change.

## Related PRs
None.

## Migrations (if applicable)
N/A — client-side only; no data migration. This DOES change default behavior for users with no saved preference: the tab strip is now on by default rather than off. A user can opt out via Preferences at any time; the setting persists per-browser in `localStorage`.

## Test environment
Manually verified against the deployed staging backend (real authentication, real case/incident/change-request data) via a local dev server pointed at staging, and against a full local stack (real ServiceNow DEV + real Asgardeo). Browser: Chromium (via CDP automation) at 1700×1100. No JDK/OS/DB dependency — this is a frontend-only change.

## Learning
N/A.

---

**Note on PR size:** this touches many files across several rounds of iterative refinement driven by hands-on manual testing and two independent review passes (CodeRabbit + a human reviewer's AI-assisted pass) — each round caught and fixed a real bug (a React Router nesting crash, a stale-label bug, a cap-overflow logic bug, a context-menu positioning bug, a stale-ref race causing a false-positive capacity toast, a scroll-position fix that relied on non-guaranteed browser behavior, a reactivation bug, and an accessibility gap in the tab strip's keyboard support). It's a single cohesive feature that doesn't split cleanly along file lines.

## Summary by CodeRabbit

* **New Features**
  * Added browser-style tabs for cases, service requests, incidents, change requests, engagements, announcements, and security analyses.
  * Tabs support activation, closing, bulk actions, up to 10 open items, session restoration, and independent section filters.
  * Added Preferences for theme and case-tab behavior, with options to replace the oldest or last tab.
  * Added draft protection before closing tabs with unsent replies.
  * Replaced the header theme selector with a built-in color-scheme toggle.

* **Bug Fixes**
  * Improved navigation, scrolling, background tab handling, and anchor preservation.
  * Preserved full-page navigation when tabs are disabled.


apps/csm-portal/webapp/src/features/case-tabs/components/CaseTabStrip.tsx
```diff
@@ -0,0 +1,229 @@
...
+              onDelete={(e) => {
+                // Chip's onDelete already receives a synthetic event whose
+                // propagation stopping is handled by oxygen-ui internally;
+                // stopPropagation here too so a delete click never also
+                // triggers the Chip's own onClick (which would activate the
+                // tab that's about to close).
+                e.stopPropagation();
+                onRequestClose(tab.id);
+              }}
+              aria-label={`Close ${label}`}
```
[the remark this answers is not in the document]
rksk: Confirmed — the aria-label overrides the tab's accessible name for screen readers. Fix is queued locally (not yet pushed pending an internal go-ahead).

apps/csm-portal/webapp/src/features/case-tabs/components/CaseTabStrip.tsx
```diff
@@ -0,0 +1,229 @@
...
+      <Menu
+        open={Boolean(menuAnchorPosition)}
+        onClose={closeContextMenu}
+        anchorReference="anchorPosition"
+        anchorPosition={menuAnchorPosition ?? undefined}
+      >
+        {menuTarget?.kind === "tab" && (
+          <MenuItem
+            onClick={() => {
+              if (menuTarget.kind === "tab") onCloseOthers(menuTarget.tabId);
```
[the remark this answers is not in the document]
rksk: Valid — bulk-close bypassing the draft check was a disclosed simplification in the original implementation, but you're right it's a real correctness gap, not just a nicety. Fix (route bulk-close through the same draft-confirm protection) is queued locally, not yet pushed pending an internal go-ahead.

apps/csm-portal/webapp/src/features/case-tabs/hooks/useCurrentLocationTab.ts
```diff
@@ -0,0 +1,78 @@
...
+ * track of "the last non-case page" — that's the whole point of the pin.
+ */
+export function useCurrentLocationTab(): CurrentLocationTab {
+  const location = useLocation();
+  const isCaseRoute = matchCaseLocation(location.pathname) !== undefined;
+
+  const [lastNonCaseLocation, setLastNonCaseLocation] = useState(() =>
+    isCaseRoute
+      ? { pathname: "/dashboard", search: "" }
+      : { pathname: location.pathname, search: location.search },
```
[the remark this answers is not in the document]
rksk: Confirmed — the hash fragment is dropped from the pinned-tab location. Fix is queued locally, not yet pushed pending an internal go-ahead.

apps/csm-portal/webapp/src/features/csm-cases/pages/CsmCaseDetailPage.tsx
```diff
@@ -325,9 +328,34 @@ const CASE_TAB_IDS: readonly CaseTabId[] = TAB_DEFS.filter(
...
+  // The override exists because this page can be mounted several times at
+  // once (one per open tab, all kept alive in the background — see
+  // `CaseTabIsolatedRouter`), while there is only ever ONE real matched
+  // route/location for the app as a whole. Without it, every background
+  // tab's `useParams`/`useLocation` would resolve to whatever route is
+  // CURRENTLY on-screen, not the case this particular instance represents.
+  const routedCaseId = useNormalizedIdParam("caseId");
+  const routedNavigate = useNavTransition();
+  const routedLocation = useLocation();
+  const routeOverride = useCaseRouteOverride();
```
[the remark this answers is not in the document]
rksk: Confirmed — this duplicates a limitation we'd already flagged internally (the in-page section tabs read/write the real router query instead of the per-tab override). Fix is queued locally, not yet pushed pending an internal go-ahead.

apps/csm-portal/webapp/src/layouts/AppLayout.tsx
```diff
@@ -118,104 +123,122 @@ export default function AppLayout({ children }: AppLayoutProps): JSX.Element {
...
+                  <Suspense fallback={<RouteSuspenseFallback />}>
+                    {/* Every open case tab's page, kept alive and hidden via
+                        CSS unless it's both the active tab and the current
+                        route is a case-detail route — see
+                        CaseTabsContentHost's own doc comment. Sits alongside
+                        (not instead of) children/<Outlet/>: a case route's
+                        own element now renders nothing itself once its tab
+                        is open (see CaseDetailRouteSync), so there's no
+                        double-render. */}
+                    <CaseTabsContentHost />
```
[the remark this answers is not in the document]
rksk: Confirmed — the shared scroll container resets on every tab switch. Fix is queued locally, not yet pushed pending an internal go-ahead.

rksk:
@coderabbitai review

rksk:
@coderabbitai review

apps/csm-portal/webapp/src/features/case-tabs/components/CaseTabIsolatedRouter.tsx
```diff
@@ -0,0 +1,182 @@
...
+        // came back at the top, even though the tab's other state (drafts,
+        // open dialogs) correctly persisted (this component is kept alive,
+        // never unmounted, purely toggling `display`/`hidden`). Giving each
+        // tab its own scrolling element means each one's `scrollTop` is a
+        // property of ITS OWN never-unmounted node, which the browser
+        // preserves across a `display: none` toggle on its own — no manual
+        // save/restore needed, and `AppLayout`'s reset (which only ever
+        // touches `mainContentRef`, a different element) can no longer reach
+        // it at all.
+        overflowY: "auto",
```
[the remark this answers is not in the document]
rksk: Fixed in cb98a0420 — no longer relying on the browser to preserve `scrollTop` across the `display:none` toggle. Each panel now captures its scroll position continuously via an `onScroll` handler and explicitly restores it in a `useLayoutEffect` when it becomes visible again. Regression test deliberately zeroes `scrollTop` while hidden (simulating a non-preserving browser) and confirms the explicit restore still works.

apps/csm-portal/webapp/src/hooks/useSectionTabs.test.tsx
```diff
@@ -234,4 +238,81 @@ describe("useQueryParamTabs", () => {
...
+          const next =
+            typeof to === "string"
+              ? (to.split("?")[1] ?? "")
+              : typeof to === "number"
+                ? ""
+                : (to.search ?? "").replace(/^\?/, "");
+          search = next ? `?${next}` : "";
+          onNavigate(search);
+        },
+      };
```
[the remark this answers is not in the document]
rksk: Fixed in cb98a0420 — the fixture now backs `search` with real React state in a wrapper component and rebuilds the override value on every render, so `navigate` triggers a genuine rerender matching what `CaseTabIsolatedRouter` actually does.

rksk:
@coderabbitai resume

rksk:
@coderabbitai review

rksk:
@coderabbitai review

shayanmalinda:
🤖 **Claude review complete** — 4 comment(s) posted ([view review](https://github.com/wso2-open-operations/cs-tools/pull/1618#pullrequestreview-5063001633))

🟠 Medium: 3 · 🟡 Minor: 1

apps/csm-portal/webapp/src/features/case-tabs/components/CaseTabIsolatedRouter.tsx
```diff
@@ -0,0 +1,210 @@
...
+    const [pathnameAndSearch, hash = ""] = tab.path.split("#");
+    const [pathname, search = ""] = pathnameAndSearch.split("?");
+    return {
+      pathname,
+      search: search ? `?${search}` : "",
+      hash: hash ? `#${hash}` : "",
+    };
+  }, [tab.path]);
+
+  const [routeState, setRouteState] = useState<{
```
shayanmalinda, line 102: **🟠 Medium: A same-case reactivation updates the tab but not the mounted page**

`routeState` is seeded once and only ever changes via this tab's own in-tab `navigate`, so when an outside navigation reactivates an already-open case with a different `?tab=` section or `#hash`, the reducer's `OPEN_OR_ACTIVATE` updates `tab.path` but the kept-alive page keeps serving the old `override.search`/`hash` (read at `CsmCaseDetailPage.tsx:353` and `useSectionTabs.ts:264`). The address bar shows the new section while the page shows the old one — exactly the case the reducer comment says it exists to fix, just not reaching the page. Add an effect that re-derives `routeState` from `tab.path`/`tab.kind`/`tab.state` when they change.

---

apps/csm-portal/webapp/src/features/case-tabs/components/CaseTabStrip.tsx
```diff
@@ -0,0 +1,232 @@
...
+    // right-click means the strip's full-width container: the menu would
+    // always render at the strip's left edge regardless of where within it
+    // was actually clicked.
+    setMenuAnchorPosition({ top: e.clientY, left: e.clientX });
+    setMenuTarget(target);
+  };
+
+  return (
+    <Box
+      role="tablist"
```
shayanmalinda: **🟠 Medium: The tab strip claims a keyboard tab widget it doesn't implement**

The strip uses `role="tablist"` with `role="tab"` chips, which tells assistive tech to expect arrow-key movement between tabs and a `tabpanel` each tab controls — but there's no arrow-key handling (every chip is its own tab stop) and nothing carries `role="tabpanel"`/`aria-controls`. On top of that, close-all and close-others live only in the right-click menu, so keyboard-only users can't reach them at all. Either wire up the full pattern (roving `tabIndex`, arrow keys, `aria-controls` to the panel) plus a keyboard-reachable trigger for the bulk-close actions, or drop the tab roles for plain buttons that match the click-only behaviour.

> 📖 **Reference ([Accessibility](https://developer.mozilla.org/en-US/docs/Learn/Tools_and_testing/Client-side_JavaScript_frameworks/React_accessibility)):** "Ensure keyboard navigation works — all interactive elements must be focusable and operable."

---

apps/csm-portal/webapp/src/context/case-tabs/CaseTabsBehaviorContext.tsx
```diff
@@ -0,0 +1,185 @@
...
+ *    least-recently-viewed) closes to make room.
+ */
+export type CaseTabsCapMode = "evict-oldest" | "evict-newest";
+
+export const CASE_TABS_CAP_MODE_OPTIONS: { mode: CaseTabsCapMode; label: string }[] = [
+  { mode: "evict-newest", label: "Replace the last tab" },
+  { mode: "evict-oldest", label: "Replace the oldest tab" },
+];
+
+const DEFAULT_ENABLED = true;
```
shayanmalinda, line 63: **🟠 Medium: Ships on by default, but the description says "off by default (beta)"**

`DEFAULT_ENABLED = true` and the head commit ("flip case tabs to on by default") both make this enabled for every user with no saved preference, yet the PR description calls it "off by default (beta)" and the Preferences toggle is still labelled "Case tabs (beta)". Worth confirming which is intended — a beta feature going on for everyone is a bigger rollout decision than the description implies. If on-by-default is the call, drop the "(beta)" label and update the description; if not, flip the default back.

---

shayanmalinda (commented):
Went through the whole feature — the reducer and contexts, the isolated-router/override mechanism, the tab strip, and the three detail pages it hooks into. tsc, eslint, and all 102 new tests pass locally. Four comments (3 medium, 1 minor); start with the isolated-router one, where a same-case reactivation updates the tab but not the page you're looking at.

apps/csm-portal/webapp/src/features/csm-operations/pages/CsmChangeRequestDetailPage.tsx
```diff
@@ -214,14 +216,38 @@ const CHANGE_REQUEST_TAB_IDS: readonly ChangeRequestTabId[] = TAB_DEFS.map((t) =
...
+    | { from?: string }
+    | undefined;
   const backTarget = backState?.from ?? OPERATIONS_CR_PATH;
   const { data, isLoading, isError } = useGetChangeRequest(id);
+  // Same label computation this page's own `recordView` call below uses —
+  // The CR number as the short chip label (matching `CsmCaseDetailPage`'s
+  // own `caseNumber`-only report); change requests have no separate
+  // project-scoped id the way cases do, so the tooltip's `internalId` reuses
+  // the same number, with the subject alongside it.
+  useReportCaseTabMeta(id, {
```
shayanmalinda, line 247: **🟡 Minor: Closing a change-request or incident tab never warns about an open reply**

This page (and `CsmIncidentDetailPage`) tracks `composerOpen` but only calls `useReportCaseTabMeta`, not `useReportCaseTabDraft` — so unlike a case tab (`CsmCaseDetailPage.tsx:611`), its `hasDraft` stays false and closing the tab with a reply half-written skips the discard-confirm. Add `useReportCaseTabDraft(id, composerOpen)` on both pages to match the case page.

---

rksk:
@coderabbitai resume

apps/csm-portal/webapp/src/features/case-tabs/components/CaseTabIsolatedRouter.tsx
rksk: Confirmed — the reducer's OPEN_OR_ACTIVATE fix from earlier doesn't reach the mounted page since routeState is only ever updated by the tab's own in-tab navigate. Fix queued: re-derive routeState from tab.path/kind/state on change.

apps/csm-portal/webapp/src/features/case-tabs/components/CaseTabStrip.tsx
rksk: Agreed. We're implementing full keyboard support (roving tabIndex, arrow-key nav, aria-controls, a keyboard-reachable bulk-close path) rather than dropping the tab/tablist roles, to keep the existing role=tab test coverage valid. Queued.

apps/csm-portal/webapp/src/context/case-tabs/CaseTabsBehaviorContext.tsx
rksk: Good catch — the PR description was stale. Updated it to reflect on-by-default, and dropping the "(beta)" label from the Preferences dialog too.

apps/csm-portal/webapp/src/features/csm-operations/pages/CsmChangeRequestDetailPage.tsx
rksk: Confirmed — both pages track composerOpen but never call useReportCaseTabDraft, unlike CsmCaseDetailPage. Fix queued: add useReportCaseTabDraft(id, composerOpen) to both.

rksk:
@coderabbitai review

apps/csm-portal/webapp/src/features/case-tabs/components/CaseTabIsolatedRouter.tsx
```diff
@@ -0,0 +1,268 @@
...
+        prev.search === parsed.search &&
+        prev.hash === parsed.hash &&
+        prev.kind === tab.kind &&
+        prev.state === tab.state
+      ) {
+        return prev;
+      }
+      return { ...parsed, kind: tab.kind, state: tab.state };
+    });
+  }, [tab.path, tab.kind, tab.state]);
```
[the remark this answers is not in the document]
rksk: Fixed in 87fb41156 — replaced the useEffect with the render-time derived-state pattern already used elsewhere in this feature (useCurrentLocationTab). Compares tab.path/kind/state synchronously during render and calls setRouteState directly, not inside a hook.

rksk:
@coderabbitai review

apps/csm-portal/webapp/src/features/case-tabs/components/CaseTabStrip.tsx
rksk: Done in 180100302 — roving tabIndex, Left/Right arrow-key navigation with wrapping, Home/End, each chip's id/aria-controls wired to its role=tabpanel panel, and a visible kebab button ("More tab actions") as the keyboard-reachable path to Close all/Close other tabs, since those were previously right-click-only. 10 new tests cover this incl. real DOM focus-movement assertions.

shayanmalinda (approved):

cloby99 (approved):

PR #1621, opened 2026-08-31
[CSM][Web] Move column customizer next to tables, fix scroll, add drag-to-reorder

Rashmika998:
## Summary
- **Moved "Customise columns" next to the table it controls.** It used to live in the page header, far from the table, and wasn't available at all on the dashboard's case-family widget preview or its dashboard tiles (both reuse `CasesList` directly, bypassing `CsmIssuesView`'s header). `CasesList` now takes a `columnCustomizer` slot prop and renders it in a small toolbar row directly above its own table; each caller (`CsmIssuesView`, the dashboard "View more" preview, and now the dashboard tile itself) builds its own `ColumnCustomizerButton` and passes it through. Preferences are shared with the matching main list page where the column vocabulary matches (cases/service-requests/security-reports/engagements), so a viewer's choices carry over across all three surfaces.
- **Fixed a real horizontal-scroll bug.** `CasesList` is a hand-rolled CSS grid, not a real `<Table>`/`<TableContainer>` (unlike every other table in this app) — its wrapper's `overflow: hidden` had no scroll layer underneath it, so turning on several optional columns squished/clipped column content instead of scrolling. Fixed by wrapping the grid in its own `overflowX: auto` layer and giving the grid itself `minWidth: max-content`.
- **Replaced the up/down chevron pair with a drag handle** in `ColumnCustomizerButton`'s picker (native HTML5 drag-and-drop, no new dependency) so columns can be reordered by dragging. Added `reorderColumn` to `useColumnPreferences` for this — it moves a column directly to a target index in one atomic update, since the existing `moveColumn` closes over stale state and can't be safely chained for a multi-slot drag. The handle keeps arrow-up/down as a keyboard fallback (native drag-and-drop has no keyboard path of its own).
- **Extended column customization to dashboard tiles** (e.g. the "My Work" widget) — previously only the main list pages and the "View more" preview had it. Also fixes a latent inconsistency: Severity used to be hardcoded on for every case-family tile regardless of resourceType (even where it isn't a real concept, like service requests), only ever rendering "—" there; it's now gated the same way the other surfaces already gate it.
- Documented the underlying gotchas (CSS grid min-width trap, the `reorderColumn`-vs-`moveColumn` stale-closure issue, the columnCustomizer slot pattern) in `CLAUDE.md`.

## Test plan
- [x] `tsc -b` clean
- [x] `eslint .` clean (only 2 pre-existing warnings in untouched files)
- [x] `vitest run` — full suite passes; the only failures are 4 pre-existing, unrelated flaky tests (confirmed by re-running them in isolation and against a clean baseline)
- [x] Verified in a real Chromium browser (via temporary, since-removed harnesses): the column customizer renders directly above the table, dragging a column with the mouse moves it to the correct position (confirmed via order + drop-target highlight + dimmed dragged row), and the table now genuinely scrolls horizontally (`scrollWidth > clientWidth`, content shifts on scroll) instead of squishing

## Summary by CodeRabbit

* **New Features**
  * Added drag-and-drop column reordering, with keyboard controls as an accessible alternative.
  * Added customizable columns to case dashboards, widget previews, incidents, cases, accounts, announcements, and change requests.
  * Added Issue type and Reporter columns to case lists.
  * Added horizontal scrolling for wide case tables.
  * Column preferences are retained across related views.
* **Bug Fixes**
  * Improved resource-specific dashboard columns, including case severity.
  * Standardized issue-type labels across case creation and lists.


shayanmalinda:
🤖 **Claude review complete** — 3 comment(s) posted ([view review](https://github.com/wso2-open-operations/cs-tools/pull/1621#pullrequestreview-5064180162))

🟠 Medium: 2 · 🟡 Minor: 1

cloby99 (approved):

apps/csm-portal/webapp/src/features/csm-cases/components/CsmIssuesView.tsx
```diff
@@ -320,8 +320,10 @@ export default function CsmIssuesView({
   const availableOptionalColumns: CaseOptionalColumnId[] = [
     "product",
     "type",
+    "issueType",
     ...(showSeverityColumn ? (["severity"] as const) : []),
     "assignee",
+    "createdBy",
     "customer",
```
shayanmalinda, line 327: **🟠 Medium: The new "Customer" column is blank on every row**

This adds `customer` to the picker, but `mapCaseSearchViewToRow` in `caseSearchPayload.ts:316` hardcodes `customer: ""` and `CasesList.tsx:163` renders it with no fallback — so turning the column on shows an empty cell everywhere (unlike `createdBy`, which falls back to "—"). Either populate `customer` in the mapper, or drop it from `availableOptionalColumns` until the search view actually supplies it.

---

apps/csm-portal/webapp/src/features/csm-dashboard/config/widgetListConfig.tsx
```diff
@@ -144,23 +173,115 @@ const PREVIEW_COLUMN = { label: "Preview", width: "auto" };
...
+      "assignee",
+    ],
+    [showSeverityColumn],
+  );
+  const columnOptions = useMemo(
+    () => availableOptionalColumns.map((id) => ({ id, label: CASE_OPTIONAL_COLUMNS[id].label })),
+    [availableOptionalColumns],
+  );
+  const columnPrefs = useColumnPreferences({
+    viewId: `case-list:dashboard-tile-${resourceType}`,
```
shayanmalinda, line 238: **🟠 Medium: The dashboard tile doesn't actually share column prefs with the main list**

The description says a viewer's column choices carry across all three surfaces, but this tile keys on `case-list:dashboard-tile-${resourceType}` while the main list and the "View more" preview both use `case-list:cases` (etc.) — so a case tile gets its own bucket and a user's Cases-tab layout never shows up on it. The tile's own doc comment reads as if this separate bucket is intentional, so it's worth confirming which is right: point the tile at `case-list:${columnsViewIdForResourceType(resourceType)}` to actually share, or correct the description.

---

apps/csm-portal/webapp/src/hooks/useColumnPreferences.ts
```diff
@@ -231,9 +241,31 @@ export function useColumnPreferences({
...
 
+  const reorderColumn = useCallback(
+    (id: string, targetIndex: number) => {
+      const currentIndex = state.order.indexOf(id);
+      if (currentIndex === -1) return;
+      const clampedIndex = Math.max(0, Math.min(targetIndex, state.order.length - 1));
+      if (clampedIndex === currentIndex) return;
+      const order = [...state.order];
+      order.splice(currentIndex, 1);
+      order.splice(clampedIndex, 0, id);
```
shayanmalinda, line 252: **🟡 Minor: Dragging a column drops it on the wrong side depending on direction**

`handleDrop` passes the target's index in the full array, but this splices into the array after the dragged item is already removed — so dragging a column down lands it just below the row you dropped on, while dragging up lands it above. Same drop target, different result. Adjust the insert index when moving down so it's consistent: `const insertIndex = currentIndex < clampedIndex ? clampedIndex - 1 : clampedIndex;` then `order.splice(insertIndex, 0, id)`.

---

shayanmalinda (commented):
Went through the column-customizer move, the CasesList scroll fix and slot wiring, the useColumnPreferences hook, the drag-and-drop, and every surface's viewId. tsc, eslint, and the touched tests pass locally (610 across 52 files); the scroll fix and the shared-preference keying are both sound. Three comments (2 medium, 1 minor) — start with the customer-column one, since it ships a column that's blank on every row.

PR #1627, opened 2026-08-31
[CSM][Web] Fix /users/me hanging forever for a real not-authorized user

Rashmika998:
## Summary
- **Root cause of the not-authorized page never appearing for a real user**: after `useAuthApiClient`'s retry/silent-reauth/poll chain exhausted with `/users/me` still 401, it fell into the shared full-page sign-in redirect path, which always returns a never-resolving promise (by design — so nothing renders behind an in-flight navigation). But a genuinely-unauthorized user has a perfectly valid IdP session, so `signIn()` had nothing left to fix and could silently no-op — the request just hung forever, `isLoading` never settled, and the user was left on a normal-looking but silently degraded app indefinitely. Verified live via a temporary local-only test hook (not part of this diff) forcing a real 401 at the network layer for `/users/me` and watching the exact same hang reproduce.
- **Fix**: `useAuthApiClient` gains an opt-in `skipSignInRedirect` flag. `/users/me` (`useGetUsersMe.ts`) is the one caller that sets it — once the chain is exhausted, its own failure now surfaces immediately instead of risking that hang. Every other API call in the app keeps the existing redirect-on-persistent-401 behavior unchanged.
- **Separately**: `AuthorizedAppShell` now suppresses the routed page (dashboard, case list, etc.) for as long as `/users/me` is still resolving, not only once it's confirmed unauthorized. The routed page's own components fire their own API calls immediately on mount regardless of this outcome, so a caller about to be found unauthorized could see real response data (and use the sidebar to reach other real pages) during that window. Reuses the same `AppLayout` `children`-suppression mechanic this file already uses for `AuthPendingShell` (pre-sign-in).

## Test plan
- [x] `tsc -b` clean
- [x] `eslint` clean on all touched files
- [x] New `useAuthApiClient.test.ts` cases: `skipSignInRedirect` surfaces a persistent 401 immediately without ever redirecting (even on the first occurrence), and doesn't suppress the redirect for an unrelated call that didn't opt in
- [x] Updated `AuthGuard.test.tsx`: the routed page is suppressed (loading state shown, `minimalHeader` true) while `/users/me` is loading, not just once confirmed unauthorized
- [x] Manually verified locally end-to-end: forcing a real 401 at the network layer for `/users/me` now shows `NoPortalAccessPage` shortly after the retry/poll burst, with no other network calls firing and nothing else clickable beforehand (previously: hung forever on a silently-broken normal-looking app)
- [x] Targeted `vitest` sweep: 249/249 passing across `components`/`layouts`/`context`/`hooks`/`features/settings`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling for users without portal access by surfacing authorization failures promptly instead of potentially leaving the app stuck.
  * Prevented protected pages from appearing before access verification is complete.
  * Added a streamlined loading state with reduced navigation elements while authorization is pending.
  * Preserved normal sign-in redirect behavior for other authentication requests.


Rashmika998:
@coderabbitai review

rksk (approved):

PR #1631, opened 2026-08-31
[CSM Portal] fix case-tabs pre-auth crash from PR #1628

rksk:
## Purpose

PR #1628 (merged) shipped the CSM webapp as a single JS bundle to eliminate mid-session
route-chunk-fetch freezes. Immediately after that deploy, a real crash surfaced:
`useCurrentUser must be used within a CurrentUserProvider`, thrown on a cold reload of a
browser tab that had a case detail page open.

That fix was pushed to this same branch, but landed on the fork *after* #1628 had already
merged, so it never made it into `main`. This PR carries just the follow-up commits.

## Root cause

`AppLayout.tsx` rendered the open-case-tabs host (`CaseTabsContentHost`) unconditionally,
regardless of whether auth had resolved yet. `AuthPendingShell` (shown while sign-in is
still in flight, with no `CurrentUserProvider` ancestor) already suppressed the routed page
for exactly this reason, but the case-tabs host bypassed that suppression. On a cold
reload with a case tab restored from before the deploy, it tried to render that tab's page
immediately, which calls `useCurrentUser()` and throws.

This is a pre-existing architectural gap, not something #1628 introduced directly — but
#1628's static imports removed the chunk-fetch delay that used to give auth time to resolve
first, so the race went from rare to essentially guaranteed.

## Goals

- Stop the crash on cold reload with an open case tab.
- No change to case-tab behavior once auth has resolved normally.

## Approach

Added a `showCaseTabs?: boolean` prop to `AppLayout` (default `true`), following the same
pattern already used for the existing `minimalHeader` prop — applied synchronously in the
same render as `children`, not a flag settled a render later via an effect.
`AuthPendingShell` now passes `showCaseTabs={false}`. Chose unmount over hide: the case-tabs
host has no side effects that must keep running regardless of visibility, and a tab can't
have meaningful in-progress state to lose on a cold reload before auth even resolves.

Also reverts a temporary sourcemap-enabling change from the same branch, used only for a
one-off offline debugging session to demangle the crash's stack trace — never actually
deployed, and must not ship on a publicly-reachable staging site.

## Release note

Fixes a crash when reloading a browser tab that had a case page open, immediately after a
new deploy.

## Documentation

N/A — internal bug fix, no user-facing behavior or docs change beyond the crash no longer
happening.

## Security checks

- FindSecurityBugs: N/A (TypeScript/JS, not Go/Java) — `eslint` ran clean (0 errors, 2
  pre-existing warnings unrelated to this change)
- No secrets committed: yes

## Related PRs

Follow-up to #1628 (merged) — carries commits that were pushed to the same branch after
that PR had already merged, so they never landed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented restored case tabs from loading during authentication states where user information is unavailable.
  * Improved cold-start behavior for signed-out users, avoiding errors while authentication is pending.

* **Enhancements**
  * Added control over whether case tabs appear in the application layout, with tabs shown by default.

### DIFF
--- a/apps/csm-portal/webapp/src/layouts/AppLayout.test.tsx
+++ b/apps/csm-portal/webapp/src/layouts/AppLayout.test.tsx
@@ -1,0 +1,111 @@
+// Copyright (c) 2026 WSO2 LLC. (https://www.wso2.com).
+//
+// WSO2 LLC. licenses this file to you under the Apache License,
+// Version 2.0 (the "License"); you may not use this file except
+// in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+import "@testing-library/jest-dom/vitest";
+import { render, screen } from "@testing-library/react";
+import { MemoryRouter } from "react-router";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+// AppLayout pulls in a lot of chrome (header, sidebar, banners, the idle
+// timeout provider) that isn't relevant to the `showCaseTabs` gating under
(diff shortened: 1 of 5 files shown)


rksk:
@coderabbitai resume

Rashmika998 (approved):

PR #1633, opened 2026-09-01
[CSM Portal] CSM BFF: outages, incident handoff, verify CR/SR-metadata pass through free

rksk:
Second PR in a three-part layered rollout (entity-service, this PR: BFF, then webapp) adding
outage tracking, change-request field parity, incident specialist handoff, and service-request
catalog-variable metadata. This PR proxies the entity-service layer (already open in a prior PR)
through the CSM BFF that the webapp calls.

- Proxy all 7 outage entity-service resources through the BFF, with request-boundary validation
  (required fields, enum checks, UUID-format checks) but full-body passthrough of the actual
  payload.
- Proxy the incident specialist handoff operation, surfacing `githubIssueError` distinctly when
  the handoff itself succeeds but the follow-up GitHub issue could not be created — never folded
  silently into a clean success.
- Confirm (not assume) whether change-request field parity and catalog-variable metadata need any
  BFF code at all, given both were built as typed mappers one layer down.

No UI in this PR (BFF layer only). Outage handlers mirror the existing `problems.go` handler
pattern in this codebase exactly. The incident handoff handler forwards the response as a raw,
unmodified byte passthrough so `githubIssueError` always reaches the webapp intact, and
additionally does a partial decode of just that one field to log a warning server-side when it's
non-nil — matching an existing partial-decode-for-observability idiom already used elsewhere in
this BFF. Change-request field parity was traced end to end (interface signatures, client
methods, response writing) and confirmed to be a genuine raw-`[]byte` passthrough with no typed
struct anywhere in the BFF — zero code needed. Catalog-variable metadata was confirmed the same
way (`git diff` on the relevant handler is empty).

Same as the entity-service PR — this is the plumbing layer between that and the webapp PR that
follows.

Adds BFF support for outage tracking and incident specialist handoff; confirms change-request
field parity and catalog-variable metadata require no BFF changes.

## Documentation
N/A for this PR — no user-facing surface yet (BFF only). The webapp PR updates `/help`.

## Training
N/A — backend-only change.

## Certification
N/A — backend-only change.

## Marketing
N/A — internal SRE tooling.

## Automation tests
 - Unit tests
   > New `outages_test.go` (auth, body-size, validation, happy path, upstream-error mapping,
     all 7 handlers) and extended `incidents_test.go` (12 handoff subtests, including a dedicated
     test asserting `githubIssueError` passes through on an otherwise-successful handoff). `go
     test ./...` passes across all packages. `go test -race ./...` is clean in every package this
     PR touches; a pre-existing, unrelated race in `internal/entity` (confirmed present on
     unmodified `main` too) is untouched by this change.
 - Integration tests
   > Not run at this layer — this PR is unit-tested against a mocked entity-service client. Live
     verification happened one layer down (entity-service PR) and one layer up (webapp PR).

## Security checks
 - Ran FindSecurityBugs plugin and verified report? N/A — Go service, not Java

## Samples
N/A

## Related PRs
Part of a three-PR series: entity-service (`cs-tools#1632`), this PR (BFF), and a webapp PR to
follow, all for the same four features.

## Migrations (if applicable)
N/A

## Test environment
Go toolchain per `go.mod` in `apps/csm-portal/backend/`. No browser/UI involved in this PR.

## Learning
N/A

### DIFF
--- a/apps/csm-portal/backend/cmd/server/main.go
+++ b/apps/csm-portal/backend/cmd/server/main.go
@@ -93,6 +93,7 @@ func main() {
 	problemHandler := handler.NewProblemHandler(customerEntityClient)
 	incidentTaskHandler := handler.NewIncidentTaskHandler(customerEntityClient)
 	alertHandler := handler.NewAlertHandler(customerEntityClient)
+	outageHandler := handler.NewOutageHandler(customerEntityClient)
 
 	// Google Chat is not yet configured for every deployment, so its spaces
 	// are read with os.Getenv (never mustEnv) — a missing or malformed value
@@ -233,6 +234,7 @@ func main() {
 	mux.HandleFunc("POST /incidents/{id}/comments", incidentHandler.CreateIncidentComment)
 	mux.HandleFunc("POST /incidents/{id}/comments/search", incidentHandler.SearchIncidentComments)
 	mux.HandleFunc("POST /incidents/{id}/activities/search", incidentHandler.SearchIncidentActivities)
+	mux.HandleFunc("POST /incidents/{id}/specialist-handoffs", incidentHandler.HandOffIncidentToSpecialist)
 	mux.HandleFunc("GET /alerts/{id}", alertHandler.GetAlert)
 	mux.HandleFunc("GET /smart-alerts/{id}", alertHandler.GetSmartAlert)
 	mux.HandleFunc("POST /change-requests/{id}/comments", changeRequestHandler.CreateChangeRequestComment)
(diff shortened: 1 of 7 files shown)


rksk:
**Live verification complete** against real ServiceNow DEV (`wso2sndev.service-now.com`) via the local full-stack mode (webapp → BFF → Go entity-service → Ballerina → real SN DEV, real Asgardeo login).

Results:
- App load — pass
- Outages: list/search, create, detail, close/edit — pass, after fixing 4 real bugs found only by this live pass (none caught by `tsc`/`eslint`/`vitest`/`go test`/`go vet`, all clean throughout):
  1. Outage search request was nested under `filters`/`pagination`; the upstream contract is a flat closed record.
  2. Outage-communications search request had the same nesting bug in a sibling payload.
  3. `GET /outages/{id}` wasn't unwrapping the response's `{"outage": {...}}` envelope — silently returned an all-zero-value outage with a `200`.
  4. The outage detail page crashed on a legitimately-`null` (not empty-array) `affectedConfigurationItems` field.
- Incident specialist handoff — pass (real SN task + assignment-group update; GitHub issue creation 401'd as expected, no real credentials in local dev).
- Change-request field parity — pass (all 6 new writable fields, full read card).
- Service-request catalog-variable metadata — pass (mandatory/readOnly/maxLength enforcement confirmed).

This PR is part of a 3-PR split (entity-service `cs-tools#1632`, CSM BFF `cs-tools#1633`, CSM webapp `cs-tools#1644`) — the combined diff exceeded CodeRabbit's free-tier review guardrails (56 files, several >300 lines), so it was rebuilt as three focused, independently-buildable PRs by layer.

Rashmika998 (approved):

PR #1634, opened 2026-09-01
[CSM] Merge dev-app-csm-portal into main

SasmithaDilshan:
Merges `dev-app-csm-portal` into `main`.

418 commits ahead of `main`; `main` is 277 ahead of `dev`. This carries work from
several teams, not one:

| Area | Commits |
|---|---|
| `integrations/` | 113 |
| `apps/csm-portal` | 110 |
| `apps/customer-portal/backend-v2` | 87 |
| `entity-service` | 56 |
| `apps/customer-portal/webapp` | 13 |

The customer-portal side is the Ballerina→Go `backend-v2` migration: case detail
labels and the Calls tab, attachments, WebSocket feedback parity, project and
onboarding fields, case escalation/duration, time-card state and date validation,
product-vulnerability bulk fetch, and person-reference decoding across cases,
conversations, comments and activities.

## Conflicts — not yet resolved

7 files conflict. **None are in `apps/customer-portal`** (backend-v2, webapp, and
the Ballerina backend all merge clean). They fall into three groups:

**Pure additions — resolve by keeping both sides:**
- `entity-service/internal/domain/entity.go` — `HasSr` (main) + `ProjectEngagementFields` (dev)
- `entity-service/internal/service/sn_project_service.go` — `HasSr` (main) + the query/onboarding hour fields (dev)

**Stale side — take `main`:**
- `entity-service/internal/service/case_service.go` — dev rejects `type: "announcement"`
  on case creation as unsupported; main implements it (2026-08-24, six days after dev's
  rejection was written, including the SN payload-building case dev's comment says is
  missing). Dev's comment is out of date; main's implementation is the current intent.

**Needs a decision from the entity-service and csm-portal owners:**
- `entity-service/internal/server/routes.go`
- `entity-service/internal/handler/case_handler.go`
- `entity-service/internal/service/interfaces.go`
- `entity-service/internal/service/sn_case_service.go`

Both branches independently implemented `GET /attachments/{id}`, with different
contracts and a live consumer each:

| | `main` | `dev` |
|---|---|---|
| Method | `GetAttachment` | `GetAttachmentByID` |
| Returns | `domain.Attachment` (metadata) | `domain.AttachmentDetails` (metadata + base64 `content`) |
| Consumer | csm-portal backend | customer-portal backend-v2 |

Keeping both registers the same `GET /attachments/{id}` pattern twice, which makes
`net/http.ServeMux` panic at startup — entity-service would not boot. Taking dev's
side alone would start returning base64 file content on every metadata fetch to the
csm-portal backend, which passes the response straight through. Taking main's side
alone removes the content path backend-v2 depends on.

Resolving this properly likely means two distinct routes rather than one overloaded
one, which is an API decision for the owners of both consumers — so these four files
should be resolved together as one unit, not file by file.

A further 24 conflicting files under `apps/csm-portal` belong to that team.


SasmithaDilshan:
Superseded by #1636, which is the same merge opened from the fork (SasmithaDilshan:dev-app-csm-portal) rather than from the upstream branch, and carries the current ref state plus the entity-service conflict-resolution map.

PR #1636, opened 2026-09-01
[CSM] Merge dev-app-csm-portal into main

SasmithaDilshan:
Merges `dev-app-csm-portal` into `main`, from the fork. **All conflicts resolved**
in `a5aaad5`.

Work from several teams:

| Area | Commits |
|---|---|
| `integrations/` | 113 |
| `apps/csm-portal` | 110 |
| `apps/customer-portal/backend-v2` | 87 |
| `entity-service` | 56 |
| `apps/customer-portal/webapp` | 13 |

The customer-portal side is the Ballerina→Go `backend-v2` migration: case detail
labels and the Calls tab, attachments, WebSocket feedback parity, project and
onboarding fields, case escalation/duration, time-card state and date validation,
product-vulnerability bulk fetch, and person-reference decoding across cases,
conversations, comments and activities.

## Conflict resolution — 24 files

**None were in `apps/customer-portal`**; it merged clean. 8 were in `entity-service`,
16 in `apps/csm-portal`. Both branches' features are preserved. Three collisions
could not be kept literally:

**Same name, different feature — renamed so both survive:**
- `domain.CaseFeedback` — main's case-satisfaction rating keeps the name; dev's
  emoji/chips feedback becomes `CaseEmojiFeedback`. Different endpoints, no URL
  collision, no API change.
- `watchListEmails` — main's upstream-resolving helper keeps the name; dev's
  watch-list extractor becomes `watchListUserEmails`. Both have live callers.

**Duplicate declarations git merged silently** (no conflict marker — both sides added
them in different regions of one file, so only the compiler catches it):
`UpdateAttachmentRequest`, `UpdateAttachmentResponse`, `UpdatedAttachment`. Identical
copies deduped; main's request struct kept, since `json.RawMessage` keeps absent /
explicit-null / value distinguishable.

**Same thing implemented twice — superset kept:**
- `GET /attachments/{id}` → dev's `AttachmentDetails`. The csm-portal backend needs
  `storageKey`, which only dev's shape carries, so one implementation serves both
  consumers. Registering both handlers would panic `ServeMux` at startup.
- `UpdateAttachment` → main's request struct with dev's richer per-reference-type
  validation. An explicit `null` description still means "clear it", not "absent".
- Case type `announcement` → main's implementation; dev's rejection predates it and
  its explanatory comment is stale.

**Resolved in main's favour:**
- `useAuthApiClient` → main's version, which carries the 2026-08-31 fix for
  `/users/me` hanging for a not-authorized user. Dev's `useAuthTokens` extraction is
  not applied to this hook, and `useAuthTokens` itself stays in use by the
  case-activity SSE stream. **Worth a look from whoever owns both sides** — the clean
  end state is main's fix ported onto dev's extracted structure.

### Notification channels kept in the backend

Dev deleted `apps/csm-portal/backend/internal/notifications` (and
`internal/handler/notifications.go`) in `496facd83`, *"Extract notification channels
into a standalone csm-notification-service"* (2026-07-29). Git honoured that deletion
silently, since main had made no change to those files since the merge base — so the
merge initially dropped all seven files and main's `POST /notifications/google-chat/alerts`
route with them.

They have been **restored deliberately** (`3e7fa9ed3`): the package, the handler, both
test files, the `googleChatClient`/`notificationHandler` construction,
`parseGoogleChatSpaces`, and the route. Main keeps its in-backend Google Chat and email
channels rather than relying solely on `integrations/csm-notification-service`.

Consequence worth reviewing: the same `email.go`/`googlechat.go` logic now exists in two
places — `apps/csm-portal/backend/internal/notifications` and
`integrations/csm-notification-service/internal/notifications`. They are separate modules
and do not conflict at build time, but they will need to be kept in sync by hand until
one is retired.

Four tests were removed, each obsoleted by the other branch's feature rather than to
make the suite pass: two asserted the Postgres data source *cannot* serve attachments
(dev implements it), one asserted announcement creation is rejected (main implements
it), one asserted main's removed response shape.

## Verification

- `entity-service` and the csm-portal backend: build, vet, gofmt clean; **all 13 test
  packages pass** (including the restored `internal/notifications`). entity-service
  `gosec` 123 files / 0 issues, `govulncheck` clean. The csm-portal backend's `gosec`
  reports 4 findings (`dashboard/sections.go`, `dashboard/registry.go`,
  `sftpgo/client.go` x2) — all pre-existing and matching `origin/main`'s own baseline;
  none are in files touched by this merge.
- Tag-level diff against **both** branches: no struct tag dropped — the silent failure
  mode behind several field-loss bugs on this migration.
- OpenAPI path/schema diff against both branches: nothing lost except main's
  `group-by` → `aggregate` rename and dev's account-response unification, both
  deliberate. The routes↔spec contract test passes.

**Not verified:** the 8 resolved `apps/csm-portal/webapp` TypeScript files. No node
toolchain was available where the merge was performed, so those had structural checks
only and need CI or a local `pnpm test`. `pnpm-lock.yaml` was resolved to match the
auto-merged `package.json` (`0.13.1`) and is worth regenerating with
`pnpm install --lockfile-only`.


rksk:
**Bug: duplicate `CaseFeedback` OpenAPI schema key**

`entity-service/openapi.yaml:12974`

The merge renamed the emoji/chips feedback Go type to `CaseEmojiFeedback`, but left its OpenAPI schema name as `CaseFeedback` — colliding with the unrelated satisfaction-rating `CaseFeedback` schema already defined at line 9128.

Since `components/schemas` is a YAML map, the second `CaseFeedback:` key (emoji/chips shape: `id`, `emoji`, `chips`, `assessmentId`, `createdBy`, `createdOn`) silently overwrites the first (satisfaction-rating shape: `instanceId`, `caseId`, `rating`, `ratingLabel`, `comment`, `submittedAt`). Any YAML-parsing tool — codegen, redocly/swagger-ui rendering, client generation for `POST /cases/feedback/search` — ends up with the wrong response shape.

Suggest renaming the emoji/chips schema to `CaseEmojiFeedback` to match its Go type.

(Note: this file's diff is too large for GitHub inline comments, hence posting here instead of on the line directly.)

entity-service/internal/config/config.go
```diff
@@ -84,15 +105,27 @@ func getEnvOrDefault(key, defaultVal string) string {
...
+// DATA_SOURCE=servicenow, or if EVENT_HUB_BROKER/EVENT_HUB_CONNECTION_STRING/
+// EVENT_HUB_TOPIC are only partially set.
 func (c *Config) Validate() error {
 	switch c.DataSource {
 	case DataSourcePostgres, DataSourceServiceNow:
 		// valid
 	default:
 		return fmt.Errorf("invalid DATA_SOURCE %q: must be %q or %q", c.DataSource, DataSourcePostgres, DataSourceServiceNow)
 	}
+	if c.DBUser == "" {
```
rksk, line 131: This makes `DB_USER`/`DB_PASSWORD`/`DB_NAME` (and a reachable Postgres) unconditionally required, including when `DATA_SOURCE=servicenow`. Previously the pool was skipped in that mode, so existing DB-less deployments will crash-loop at boot with `DB_USER is required` after this rolls out.

If requiring the DB in all modes is intentional (event publish failure tracking), this needs:
1. `.env.example` updated — it still says the DB vars are only required when `DATA_SOURCE=postgres`.
2. An upgrade/deploy note so operators provision DB credentials before rolling this out.

entity-service/internal/domain/entity.go
```diff
@@ -289,20 +261,12 @@ type SearchAccountsRequest struct {
...
-
-// SNAccountView is the account view returned by the ServiceNow data source.
-// Timestamp fields are kept as strings to accommodate empty values from ServiceNow.
+// AccountView is the unified account search-result shape returned for all data
+// sources. Both Postgres and ServiceNow responses are mapped to this type so
+// callers receive the same fields regardless of which backend is active.
+// Fields not available from a given data source are left nil.
 // SupportTier is returned as a plain label string (no ID).
-type SNAccountView struct {
+type AccountView struct {
```
rksk, line 269: The unified `AccountView` renames `agentEnabled`/`kbReferencesEnabled` to `hasAgent`/`hasKbReferences` and replaces `tier` with `supportTier`, but the consumers were not updated to match:

- CSM webapp: `CsmAccountDetailPage.tsx:319` and `AccountPreviewDrawer.tsx:94` still read `a.agentEnabled` (now always `undefined`), so the Agent chip shows Disabled/No for every account.
- Microapp: `account.model.ts` copies `tier: dto.tier` with no `supportTier` fallback, so `AccountsTab.tsx:45` and `AccountDetailPage.tsx:51` render an empty tier chip.

Both mappers emit the new shape and the BFF is a passthrough, so this is wrong data in the UI under both data sources, not an edge case. The FE readers need updating to the new field names (or the view needs to keep emitting the old names alongside).

apps/customer-portal/backend-v2/internal/handler/ai_chat.go
```diff
@@ -0,0 +1,501 @@
...
+	if err != nil {
+		slog.ErrorContext(r.Context(), "entity CreateConversation failed", "userID", user.UserID, "projectID", projectID, "err", summarizeErr(err))
+		mapUpstreamError(w, err, "Failed to create a new conversation.")
+		return
+	}
+	conversationID := convResp.Conversation.ID
+
+	chatResp, err := h.ai.CreateChat(r.Context(), aichatagent.ChatPayload{
+		Message:        req.Message,
+		AccountID:      projectID,
```
rksk: `CreateConversation` (here) and `SendConversationMessage` (~line 342) pass `AccountID: projectID` without resolving the project's account, while the WebSocket path (`websocket.go:280-306`) resolves `project.Account.ID` for the same conversations — its tests assert that override precisely so the per-account token budget is billed to an account the caller belongs to.

Both REST routes are registered in `main.go` and gateway-reachable, so a REST caller gets a separate budget key per project (N projects → N+1 budgets per account) and split analytics. Either resolve the account here the same way the WS path does, or drop the REST routes if only the WS path is meant to ship.

apps/customer-portal/backend-v2/internal/dto/case_time_cards_audit_test.go
```diff
@@ -0,0 +1,120 @@
...
+func TestMapCaseTimeCardSearchResponse_EmitsCaseAuditFields(t *testing.T) {
+	out := MapCaseTimeCardSearchResponse(entity.SearchCaseTimeCardsResponse{
+		Cases: []entity.CaseTimeCardSummary{{
+			Case: entity.CaseTimeCardCaseRef{
+				ID:        "70d63bca-3bd2-0310-9140-4c6aa5e45a37",
+				Number:    "CS0440883",
+				Name:      "Combined-field PATCH test case title",
+				UpdatedOn: "2026-08-06 17:26:23",
+				CreatedOn: sp("2026-07-27 08:39:48"),
+				CreatedBy: sp("rashmika@wso2.com"),
```
rksk: These test fixtures (here and in `user_reference_test.go`, `comment_author_test.go`, plus a non-test comment in `user_reference.go` ~line 31) use real individuals' names and work emails. This is a public repository — fixtures must use dummy values (`jane.doe@example.com`, `Jane Doe`) instead of identifiable real people. None of these appear on `main` today, so this PR is what introduces them. Please scrub all occurrences before merge.

apps/customer-portal/backend-v2/internal/entity/types.go
```diff
@@ -0,0 +1,2435 @@
...
+// AccountSummary is a single search result item from POST /accounts/search —
+// a superset of entity-service's Postgres Account and ServiceNow SNAccountView.
+type AccountSummary struct {
+	ID     string  `json:"id"`
+	Name   string  `json:"name"`
+	Region *string `json:"region,omitempty"`
+	// Postgres-only.
+	SfID                *string `json:"sfId,omitempty"`
+	Tier                *string `json:"tier,omitempty"`
+	OwnerID             *string `json:"ownerId,omitempty"`
```
rksk, line 364: The entity structs here still decode `owner`/`ownerId`, but the unified entity-service response emits only `accountManager` under both data sources. As a result `GET /accounts/{id}` and `POST /accounts/search` map the account owner to null for every account, silently dropping a field this backend's own openapi.yaml documents. The decode needs updating to `accountManager` (or to accept both during transition).

apps/csm-portal/microapp/src/types/account.model.ts
```diff
@@ -44,7 +56,7 @@ export function toAccount(dto: AccountDto): Account {
     id: dto.id,
     sfId: dto.sfId,
     name: dto.name,
-    tier: dto.tier,
+    tier: resolveTier(dto),
```
rksk, line 68: `resolveTier` handles the tier rename, but `toAccount` still reads `dto.owner?.id`/`dto.owner?.name`/`dto.ownerId` a few lines below — keys the unified entity response no longer emits under either data source (it sends `accountManager` now, and the backend passes bytes through verbatim). So `ownerId`/`ownerName` are always null and the account detail page renders "—" for every account's owner. Same class of miss as the `hasAgent`/`tier` fixes this PR already made in the webapp; the owner mapping needs an `accountManager` fallback here too.

apps/csm-portal/webapp/src/hooks/useAuthTokens.ts
```diff
@@ -0,0 +1,218 @@
...
+// creates its own closures, so this lives at module scope to ensure only ONE
+// full sign-in redirect is triggered even when many concurrent callers
+// (across different hook instances) discover a dead refresh token at once.
+let signInInFlight = false;
+
+// Shared across every caller's hook instance for the same reason as
+// signInInFlight: many concurrent callers can discover a dead refresh token
+// at once, and they should all await the SAME hidden-iframe silent sign-in
+// attempt rather than each opening their own.
+let silentSignInInFlight: Promise<boolean> | null = null;
```
rksk: This hook keeps its own module-scope `signInInFlight`/`silentSignInInFlight` guards while `useAuthApiClient` and `AuthGuard` coordinate through the shared `trySilentSignInOnce` (`silentSignIn.ts`), whose documented invariant is at most one re-auth cycle app-wide. The two systems don't see each other: a dead refresh token discovered simultaneously by the SSE stream (this hook) and any REST fetch (`useAuthApiClient`) fires two concurrent hidden-iframe `signInSilently` cycles — the exact race `silentSignIn.ts` warns can drop an in-flight request such as a comment POST — and this hook's redirect path also bypasses `useAuthApiClient`'s `recentlyForcedSignIn` loop guard. The doc comment below ("useAuthApiClient.ts is this hook's primary consumer") no longer holds in the merged tree. Suggest routing this hook's re-auth through `trySilentSignInOnce` so there is a single flight.

apps/customer-portal/backend-v2/internal/aichatagent/ws.go
```diff
@@ -0,0 +1,270 @@
...
+			Scopes:       cfg.Scopes,
+		},
+	}
+}
+
+// dial opens a WebSocket connection to the upstream AI chat agent for the
+// given session ID, authenticated with a bearer token obtained via OAuth2
+// client credentials.
+func (c *WSClient) dial(ctx context.Context, sessionID string) (*websocket.Conn, error) {
+	token, err := c.oauth.Token(ctx)
```
rksk: `c.oauth.Token(ctx)` on `clientcredentials.Config` builds and discards a fresh token source per call, so every chat turn and side-channel message pays a full token-endpoint round trip before the upstream WebSocket handshake — added first-byte latency per message and one identity-provider grant per message across all users (throttling risk). The sibling REST client already caches (`NewClient` holds `cc.Client(tokenCtx)` once); the same applies here: store `cfg.TokenSource(...)` in `NewWSClient` and call `ts.Token()` in dial so the token is reused until expiry.

entity-service/internal/service/sn_case_service.go
```diff
@@ -2284,58 +3046,117 @@ func (s *snCaseService) DeleteCaseAttachment(ctx context.Context, req domain.Del
...
-// The GET /attachments/{id} response does not carry referenceType (mirroring the
-// /attachments/search response's snAttachment shape), so the returned domain.Attachment's
-// ReferenceType is left as the zero value -- callers that need it already know it from
-// context (e.g. the request that led them to the attachment id).
-func (s *snCaseService) GetAttachment(ctx context.Context, attachmentID string) (domain.Attachment, error) {
-	if err := validateUUIDs("id", []string{attachmentID}); err != nil {
-		return domain.Attachment{}, err
+// snAttachmentDetails mirrors the Choreo GET /attachments/{id} response
+// (Ballerina's AttachmentResponse — every Attachment field plus content).
+// ServiceNow's attachment-details lookup returns a bare createdBy string with
```
rksk: Minor: this comment (and its twin on the domain struct) asserts the backing data source's attachment-details lookup "returns a bare createdBy string with no createdByUser", which contradicts the fixture the merge deleted and the upstream change notes. Nothing user-visible depends on it today (search still resolves the person ref separately), but the comment invites a future "fix" aligning the search decode with a wire shape that doesn't exist — and the merge also removed `GetAttachmentByID`'s only happy-path test, so the details decode (createdOn parsing, id/referenceId mapping) now ships untested. Suggest correcting the comment and restoring a happy-path test.

apps/customer-portal/backend-v2/internal/dto/case_time_cards_audit_test.go
SasmithaDilshan: Scrubbed in 817a69289 — `jane.doe@example.com` / `john.smith@example.com` / "Jane Doe" across `user_reference_test.go`, `comment_author_test.go`, `case_time_cards_audit_test.go`, and the doc comment in `user_reference.go`. I also swapped the `a@wso2.com` / "A Person" placeholders in those files to `example.com` for consistency.

Two more of the same class turned up in the sweep but are already on `main`, so I left them out of this PR rather than widening its diff: `hesara@wso2.com` in `apps/csm-portal/microapp/src/types/case.dto.ts` and `paraparan@wso2.com` in `apps/customer-portal/webapp/tests/e2e/config/testData.ts`. Happy to do those in a follow-up if you want them gone too.

apps/customer-portal/backend-v2/internal/entity/types.go
SasmithaDilshan: Fixed in 817a69289. `AccountDetail`/`AccountSummary` now decode `accountManager` alongside `owner`, and `accountOwnerRef` prefers `owner` when present and falls back to `accountManager` — the "accept both" option, so the mapping holds whichever service deploys first.

Added `internal/dto/account_owner_test.go` covering the current shape, the legacy shape, both-present, and neither-present (stays nil rather than inventing an owner).

apps/csm-portal/microapp/src/types/account.model.ts
SasmithaDilshan: Fixed in 817a69289. Added `accountManager` to `AccountDto` and a `resolveOwner(dto)` helper — `dto.owner ?? dto.accountManager ?? null` — mirroring `resolveTier`, with `ownerId` still falling through to the bare `dto.ownerId` for PG-native accounts.

Flagging honestly: no node toolchain was available where I made this change, so it has not been typechecked or run. Worth a CI check specifically on this file.

apps/customer-portal/backend-v2/internal/aichatagent/ws.go
SasmithaDilshan: Fixed in 817a69289. `NewWSClient` now builds `cc.TokenSource(tokenCtx)` once and `dial` calls `ts.Token()`, so the token is reused until expiry instead of one grant per message. `tokenCtx` mirrors `NewClient`, bounded by `tokenFetchTimeout` with `noRedirect` — noted in a comment that the refresh is therefore bounded by that timeout rather than the triggering request's context, the same trade-off the REST path already makes.

entity-service/internal/service/sn_case_service.go
SasmithaDilshan: Both fixed in 817a69289.

On the comment: you are right, and `main`'s deleted fixture is the evidence — it sends `"createdByUser": null` alongside `createdByFullName`, so the key is present, just unresolved on this path. Reworded on both the service struct and the domain type to say exactly that, with an explicit "do not align this decode with the search path's UserReference shape" note.

On the test: restored as `TestSNCaseService_GetAttachmentByID_HappyPath`, adapted to `AttachmentDetails` (bare `CreatedBy` string, `Content`, `CreatedOn` as `time.Time`), plus `..._RejectsUnparseableCreatedOn` — since an unparseable `createdOn` fails the whole call rather than degrading one field, that seemed worth pinning too.

apps/csm-portal/webapp/src/hooks/useAuthTokens.ts
SasmithaDilshan: Fixed in 817a69289 — `useAuthTokens` now routes its silent re-auth through the shared `trySilentSignInOnce` and its module-local `silentSignInInFlight` is gone, so the SSE stream and any REST fetch share one flight. The stale "useAuthApiClient.ts is this hook's primary consumer" doc comment is corrected too.

I deliberately did **not** unify the redirect path in this commit. `useAuthTokens` and `useAuthApiClient` still hold separate `signInInFlight` booleans, and `recentlyForcedSignIn`/`markForcedSignIn` are private to `useAuthApiClient`, so your point about the loop guard being bypassed still stands. Fixing it properly means extracting the redirect single-flight plus the loop guard into a shared module and pointing both hooks at it — a larger change to auth code I could not run tests for. Happy to do it if you would rather it land here; otherwise it wants your eyes.

Same caveat as the microapp change: no node toolchain available, so this is unverified beyond reading. `useAuthTokens.test.ts` mocks Asgardeo's `signInSilently` rather than the single-flight, and `inFlight` clears in `.finally()` exactly as the old guard did, so the existing assertions should still hold — but CI should confirm.

rksk (approved):

PR #1641, opened 2026-09-01
[CSM Portal] Add product vulnerability full-replace sync endpoint

rksk:
The Go entity-service exposes read-only product vulnerability endpoints (`search`, `get`, `meta`) but has no way to write vulnerability data. A separate scheduled sync job needs a write path to refresh the full set of product vulnerabilities from an upstream scanner feed.

Add `POST /products/vulnerabilities/sync` to the Go entity-service that accepts the complete current set of product vulnerabilities and applies it as a full replace.

Mirrors the existing three product-vulnerability methods:
- `interfaces.go`: adds `SyncProductVulnerabilities` to `ProductVulnerabilityService`, documented as a full-replace operation — the downstream data source deletes any existing record whose join key is not present in the submitted items and upserts everything that is, so callers must submit the complete current set, never a partial delta.
- `sn_product_vulnerability_service.go`: implements the method on `snProductVulnerabilityService`, POSTing to a corresponding new resource on the same backing service the existing read methods already call (a corresponding entity-service change is tracked separately). This call is machine-to-machine with no end-user identity, so it forwards an empty token — the client already omits the identity header when the token is empty, same as the existing read paths do when no token is present.
- `entity.go`: adds `ProductVulnerabilitySyncItem` (request item) and `ProductVulnerabilitySyncResult` (response, pointer fields since counts may be partially reported) domain types, following this file's existing conventions for the entity.
- `product_vulnerability_handler.go`: adds `SyncProductVulnerabilities`, decoding a JSON array body and responding `201 Created` (a write, unlike the existing `200` read responses).
- `routes.go`: wires `POST /products/vulnerabilities/sync` into the existing `productVulnerabilityHandler != nil` block, gated the same way as the other three routes (ServiceNow data source only).
- `openapi.yaml`: documents the new path and its two new schemas, including an explicit warning that this is a full-replace sync and a partial submission causes downstream data loss.

Purely additive — the existing search/get/meta routes, handlers, and service methods are untouched.

As a scheduled sync job, I can push the full current set of product vulnerabilities to the entity-service so downstream consumers see up-to-date vulnerability data without the entity-service needing direct access to the scanner feed.

Added a write endpoint to the Go entity-service for syncing the full set of product vulnerabilities.

N/A — this is a machine-to-machine sync endpoint for a backend job, not a user-facing feature; no product/user documentation is affected.

## Training
N/A — no training content impact for this internal sync endpoint.

## Certification
N/A — no certification exam impact for this internal sync endpoint.

## Marketing
N/A — no user-facing feature to promote.

## Automation tests
 - Unit tests
   > Sibling methods (`SearchProductVulnerabilities`, `GetProductVulnerability`, `GetVulnerabilityMeta`) have no dedicated unit tests in this package, so no new test file was added for the new method — matching the existing coverage convention for this entity rather than introducing a new one.
 - Integration tests
   > None available in this worktree: the downstream resource this endpoint calls has not been deployed to any reachable environment yet. Verified via `go build ./...`, `go vet ./...`, and the full existing test suite (`go test ./...`), all passing.

## Security checks
 - Ran FindSecurityBugs plugin and verified report? no — not part of this Go module's toolchain; relied on `go vet` instead

## Samples
N/A — no new samples for this internal endpoint.

A corresponding entity-service change (separate repo, tracked separately) adds the resource this endpoint proxies to. Not linked here per repo policy on cross-repo references.

## Migrations (if applicable)
N/A — no schema or data migration; this is a stateless proxy endpoint.

Go toolchain per `go.mod` in this module, macOS (Darwin), no browser/DB involved — built and tested locally in a git worktree; no live call against the downstream service was possible (not yet deployed anywhere reachable).

Followed the existing pattern in this file set for the three read methods on this entity (request/response struct shape, error handling via `client.Post` + `writeServiceError`, data-source gating in `routes.go`).

### DIFF
--- a/entity-service/internal/domain/entity.go
+++ b/entity-service/internal/domain/entity.go
@@ -2998,6 +2998,37 @@ type SearchProductVulnerabilitiesResponse struct {
 	Offset                 int                        `json:"offset"`
 }
 
+// ProductVulnerabilitySyncItem is one entry submitted to a full-replace sync of product
+// vulnerabilities. WSO2ID is the join key the downstream data source uses to match this item
+// against an existing record.
+type ProductVulnerabilitySyncItem struct {
+	WSO2ID                 string `json:"wso2Id"`
+	ComponentName          string `json:"componentName"`
+	ApprovedBySecurityTeam bool   `json:"approvedBySecurityTeam"`
+	ComponentType          string `json:"componentType"`
+	ComponentVersion       string `json:"componentVersion"`
+	CVE                    string `json:"cve"`
+	Justification          string `json:"justification"`
+	ProductName            string `json:"productName"`
+	ProductVersion         string `json:"productVersion"`
+	Resolution             string `json:"resolution"`
+	ScannerName            string `json:"scannerName"`
+	Severity               any    `json:"severity"`
+	UpdateLevel            string `json:"updateLevel"`
+	UseCase                string `json:"useCase"`
+	VulnerabilityID        string `json:"vulnerabilityId"`
+	WSO2Resolution         string `json:"wso2Resolution"`
+}
+
+// ProductVulnerabilitySyncResult reports the outcome of a full-replace product vulnerability
+// sync. Fields are pointers because the downstream data source may omit or partially populate
+// them; a nil field means the count was not reported, not that it was zero.
+type ProductVulnerabilitySyncResult struct {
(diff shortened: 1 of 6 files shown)


rksk:
@coderabbitai review

Rashmika998 (approved):

PR #1648, opened 2026-09-02
[CSM Portal] Record a work note when an auto-closure hold is set

rksk:
A CS engineer reported that setting/extending an auto-closure hold on a case has no visible trail in the CSM portal — the legacy ticketing UI's equivalent action used to add a work note with the hold date, but the CSM portal's "Hold auto-closure…" flow only updates the underlying fields silently.

When a case's `autocloseHoldUntil` field is set/extended via `PATCH /cases/{id}`, the backend now also records an internal work note documenting the hold and its date, matching the wording the legacy ticketing UI already produces for the same action — so CS engineers get parity, not a new UX pattern.

Backend-only (BFF) change, no API contract or entity-service changes:
- `internal/handler/cases.go`: `PatchCase` now checks for `autocloseHoldUntil` in the request body, and on a successful PATCH calls a new helper, `recordAutocloseHoldWorkNote`, which reuses the existing `CreateCaseComment` entity-service call (already used for case commenting) with `{"type":"work_note", "content": "..."}`.
- The comment write is best-effort: if it fails, it's logged (`slog.WarnContext`) and swallowed — the primary hold PATCH, which already succeeded, is never rolled back or failed because of a secondary write.
- No UI changes; the note surfaces through the existing case Activities/comments feed.

Deliberately implemented at the BFF layer rather than in the Go entity-service/ServiceNow adapter, since entity-service's SN-specific case logic won't carry forward to the Postgres-native rewrite — this keeps the fix portable.

As a CS engineer, when I put a case on hold from auto-closure, I want a record of that action (and the date) in the case's activity so I and others can tell whether/when a hold was extended, without having to check the raw field value.

Setting or extending a case's auto-closure hold now records an internal note documenting the hold and its date.

N/A — no dedicated help-doc topic exists for the "Hold auto-closure" action today; the change is additive behavior on an already-undocumented flow.

## Training
N/A — no training content references this internal ticketing action.

## Certification
N/A — no certification exam content covers this internal case-management action.

## Marketing
N/A — internal case-management behavior, not customer-facing.

## Automation tests
 - Unit tests
   > Added 3 subtests to `TestPatchCase` in `internal/handler/cases_test.go`: (1) a work note is recorded with the expected content when `autocloseHoldUntil` is set, (2) the hold PATCH still succeeds if the work-note comment call fails, (3) no note is recorded for PATCH requests that don't touch `autocloseHoldUntil`.
 - Integration tests
   > None added — no local stack / entity-service integration test harness exercised for this change; covered at the unit level via a mocked entity client.

## Security checks
 - Ran FindSecurityBugs plugin and verified report? N/A — Go codebase, not a JVM target FindSecurityBugs applies to

## Samples
N/A

## Related PRs
None

## Migrations (if applicable)

## Test environment
Verified with `go build ./...` and `go test ./...` in `apps/csm-portal/backend` (Go, macOS/Darwin dev environment). No browser/DB-specific testing needed — backend-only change with mocked entity client in tests.

## Learning
N/A

## Summary by CodeRabbit

* **New Features**
  * Case auto-closure holds now automatically generate an internal work note showing the hold date.
  * Updating a hold remains successful even if recording the work note fails.
* **Bug Fixes**
  * Work notes are no longer created when a case update does not include an auto-closure hold.

### DIFF
--- a/apps/csm-portal/backend/internal/handler/cases.go
+++ b/apps/csm-portal/backend/internal/handler/cases.go
@@ -1107,6 +1107,25 @@ func (h *CaseHandler) PatchCase(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
+	// The auto-closure hold work note (below) must only fire for an actual change in
+	// the hold date, not for a retry or a no-op PATCH that resends the existing value —
+	// otherwise every duplicate request pollutes the case's activity feed with an
+	// identical note. Read the prior value before the PATCH; after it, the case
+	// already reflects the new value and there'd be nothing to diff against.
+	var priorHoldDate string
+	if patchErr == nil && patch.AutocloseHoldUntil != nil {
+		if current, err := h.entity.GetCase(r.Context(), caseID); err != nil {
+			slog.WarnContext(r.Context(), "entity GetCase failed reading prior autoclose hold date; proceeding without dedup", "userID", user.UserID, "caseID", caseID, "err", err)
+		} else {
+			var currentCase struct {
+				AutoclosureStateTime *string `json:"autoclosureStateTime"`
+			}
+			if err := json.Unmarshal(current, &currentCase); err == nil && currentCase.AutoclosureStateTime != nil {
+				priorHoldDate = formatHoldDate(*currentCase.AutoclosureStateTime)
+			}
+		}
+	}
+
 	result, err := h.entity.PatchCase(r.Context(), caseID, body)
 	if err != nil {
 		slog.ErrorContext(r.Context(), "entity PatchCase failed", "userID", user.UserID, "caseID", caseID, "err", err)
@@ -1117,13 +1136,32 @@ func (h *CaseHandler) PatchCase(w http.ResponseWriter, r *http.Request) {
 	// Setting/extending the auto-closure hold has no visible trail of its own on the
 	// case (unlike the legacy ticketing UI's equivalent action, which records a work
 	// note). Record one here so CS engineers can see when a hold was set/extended and
-	// until when. Best-effort: the hold PATCH above already succeeded and must not be
-	// rolled back or fail the request just because this secondary write-note fails.
-	if patchErr == nil && patch.AutocloseHoldUntil != nil {
-		h.recordAutocloseHoldWorkNote(r.Context(), user, caseID, *patch.AutocloseHoldUntil)
+	// until when. Best-effort and fire-and-forget: the hold PATCH above already
+	// succeeded, so this secondary write must not delay the response or fail/roll back
+	// the request if it errors. Detached from the request context (which is canceled
+	// once the handler returns) and bounded by its own timeout instead.
+	if patchErr == nil && patch.AutocloseHoldUntil != nil && formatHoldDate(*patch.AutocloseHoldUntil) != priorHoldDate {
+		holdUntil := *patch.AutocloseHoldUntil
+		go func() {
+			ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
+			defer cancel()
+			h.recordAutocloseHoldWorkNote(ctx, user, caseID, holdUntil)
+		}()
 	}
 
 	writeJSON(w, http.StatusOK, result)
+}
+
+// formatHoldDate renders an auto-closure hold timestamp (RFC3339, as sent by the
+// FE or read back from the entity service) as the date-only form CS engineers see
+// in the UI and in the work note, since the hold is date-granularity. Falls back
+// to the raw input when it doesn't parse, so an already-invalid value is not
+// silently dropped from the comparison/note.
+func formatHoldDate(raw string) string {
+	if t, err := time.Parse(time.RFC3339, raw); err == nil {
+		return t.Format("2006-01-02")
+	}
+	return raw
 }
 
 // recordAutocloseHoldWorkNote adds an internal work note documenting an
@@ -1132,12 +1170,7 @@ func (h *CaseHandler) PatchCase(w http.ResponseWriter, r *http.Request) {
 // logged, never surfaced to the caller, since the primary hold PATCH already
 // succeeded by the time this runs.
 func (h *CaseHandler) recordAutocloseHoldWorkNote(ctx context.Context, user *middleware.UserInfo, caseID, holdUntil string) {
-	displayDate := holdUntil
-	if t, err := time.Parse(time.RFC3339, holdUntil); err == nil {
-		displayDate = t.Format("2006-01-02")
-	}
-
-	note := "Please note that this case is on-hold until " + displayDate +
+	note := "Please note that this case is on-hold until " + formatHoldDate(holdUntil) +
 		", hence it will not go through the auto closure process. It will be eligible " +
 		"for auto-closure again after this date passes, or if the case state is changed " +
 		"to 'Waiting on WSO2'."
--- a/apps/csm-portal/backend/internal/handler/cases.go
+++ b/apps/csm-portal/backend/internal/handler/cases.go
@@ -27,6 +27,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/wso2-open-operations/cs-tools/apps/csm-portal/backend/internal/middleware"
 )
@@ -1076,10 +1077,12 @@ func (h *CaseHandler) PatchCase(w http.ResponseWriter, r *http.Request) {
 
 	// Validate state transition and workState guard before forwarding to the entity service.
 	var patch struct {
-		State     *string `json:"state"`
-		WorkState *string `json:"workState"`
+		State              *string `json:"state"`
+		WorkState          *string `json:"workState"`
+		AutocloseHoldUntil *string `json:"autocloseHoldUntil"`
 	}
-	if err := json.Unmarshal(body, &patch); err == nil && (patch.State != nil || patch.WorkState != nil) {
+	patchErr := json.Unmarshal(body, &patch)
+	if patchErr == nil && (patch.State != nil || patch.WorkState != nil) {
 		current, err := h.entity.GetCase(r.Context(), caseID)
 		if err != nil {
 			slog.ErrorContext(r.Context(), "entity GetCase failed during state validation", "userID", user.UserID, "caseID", caseID, "err", err)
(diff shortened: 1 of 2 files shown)


apps/csm-portal/backend/internal/handler/cases.go
```diff
@@ -1076,10 +1077,12 @@ func (h *CaseHandler) PatchCase(w http.ResponseWriter, r *http.Request) {
...
 	// Validate state transition and workState guard before forwarding to the entity service.
 	var patch struct {
-		State     *string `json:"state"`
-		WorkState *string `json:"workState"`
+		State              *string `json:"state"`
+		WorkState          *string `json:"workState"`
+		AutocloseHoldUntil *string `json:"autocloseHoldUntil"`
 	}
-	if err := json.Unmarshal(body, &patch); err == nil && (patch.State != nil || patch.WorkState != nil) {
+	patchErr := json.Unmarshal(body, &patch)
```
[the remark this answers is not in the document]
rksk: Reasoned no-op: this can't actually happen with the current data flow. The BFF forwards the PATCH body verbatim to the Go entity-service, whose request struct types `autocloseHoldUntil` as `*time.Time` (entity-service's domain type), so a malformed value like "not-a-date" fails JSON unmarshaling upstream and `h.entity.PatchCase` returns an error before this handler ever reaches the note-recording code (which only runs after the PATCH already succeeded, i.e. after the date already passed strict typed validation). Leaving as-is.

apps/csm-portal/backend/internal/handler/cases.go
```diff
@@ -1111,9 +1114,48 @@ func (h *CaseHandler) PatchCase(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// Setting/extending the auto-closure hold has no visible trail of its own on the
+	// case (unlike the legacy ticketing UI's equivalent action, which records a work
+	// note). Record one here so CS engineers can see when a hold was set/extended and
+	// until when. Best-effort: the hold PATCH above already succeeded and must not be
+	// rolled back or fail the request just because this secondary write-note fails.
+	if patchErr == nil && patch.AutocloseHoldUntil != nil {
```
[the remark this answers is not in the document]
rksk: Fixed in c0bd1b445: the note is now only recorded when the requested hold date actually differs from the case's current `autoclosureStateTime` (fetched via GetCase before the PATCH), so a retry or no-op PATCH resending the same value no longer creates a duplicate note.

apps/csm-portal/backend/internal/handler/cases.go
```diff
@@ -1111,9 +1114,48 @@ func (h *CaseHandler) PatchCase(w http.ResponseWriter, r *http.Request) {
...
+		h.recordAutocloseHoldWorkNote(r.Context(), user, caseID, *patch.AutocloseHoldUntil)
```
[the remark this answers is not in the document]
rksk: Fixed in c0bd1b445: the work-note write now runs in a background goroutine detached from the request context (its own bounded 15s timeout) after the PATCH response has already been prepared, so a slow/stalled CreateCaseComment call no longer holds up the caller.

shayanmalinda:
🤖 **Claude review complete** — re-reviewed at `02ac3d5`; all four earlier findings verified fixed, no new findings. [Original review](https://github.com/wso2-open-operations/cs-tools/pull/1648#pullrequestreview-5093048240)

<sub>Reviewed on behalf of @shayanmalinda. AI-generated — please verify before applying.</sub>

rksk:
@coderabbitai resume

apps/csm-portal/backend/internal/handler/cases.go
```diff
@@ -1104,16 +1107,88 @@ func (h *CaseHandler) PatchCase(w http.ResponseWriter, r *http.Request) {
...
+	// case (unlike the legacy ticketing UI's equivalent action, which records a work
+	// note). Record one here so CS engineers can see when a hold was set/extended and
+	// until when. Best-effort and fire-and-forget: the hold PATCH above already
+	// succeeded, so this secondary write must not delay the response or fail/roll back
+	// the request if it errors. Detached from the request context (which is canceled
+	// once the handler returns) and bounded by its own timeout instead.
+	if patchErr == nil && patch.AutocloseHoldUntil != nil && formatHoldDate(*patch.AutocloseHoldUntil) != priorHoldDate {
+		holdUntil := *patch.AutocloseHoldUntil
+		go func() {
+			ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
```
shayanmalinda: **🔴 High: The work note is written without the caller's identity token**

The entity client pulls `x-user-id-token` and `X-CSM-Correlation-ID` off the **context**, not the request (`internal/entity/customer_client.go:123-128`), and `CreateCaseComment`'s own doc comment says "createdBy is resolved by the entity service from the forwarded x-user-id-token" (`cases.go:234`). A fresh `context.Background()` carries neither, so this note reaches the entity service unattributed — either rejected outright or filed with no author — and the only signal is a `slog.Warn`, which silently degrades the very audit trail this PR exists to create. `context.WithoutCancel` keeps the values while still dropping the request's cancellation (go.mod is on 1.26, so it's available):

```go
ctx, cancel := context.WithTimeout(context.WithoutCancel(r.Context()), 15*time.Second)
```

Same site, smaller point: the goroutine isn't tied to `srv.Shutdown` (`cmd/server/main.go:287`), which waits on in-flight requests only — a note in flight during a rolling deploy is dropped.

---

apps/csm-portal/backend/internal/handler/cases.go
```diff
@@ -1104,16 +1107,88 @@ func (h *CaseHandler) PatchCase(w http.ResponseWriter, r *http.Request) {
...
+	// otherwise every duplicate request pollutes the case's activity feed with an
+	// identical note. Read the prior value before the PATCH; after it, the case
+	// already reflects the new value and there'd be nothing to diff against.
+	var priorHoldDate string
+	if patchErr == nil && patch.AutocloseHoldUntil != nil {
+		if current, err := h.entity.GetCase(r.Context(), caseID); err != nil {
+			slog.WarnContext(r.Context(), "entity GetCase failed reading prior autoclose hold date; proceeding without dedup", "userID", user.UserID, "caseID", caseID, "err", err)
+		} else {
+			var currentCase struct {
+				AutoclosureStateTime *string `json:"autoclosureStateTime"`
```
shayanmalinda, line 1128: **🟠 Medium: The dedup key isn't the prior hold date, so the first note is often skipped**

`autoclosureStateTime` isn't "the hold date" — per `openapi.yaml:7586` it is "when the auto-closure sequence next advances" for whichever `autoclosureStep` the case is currently in, so on a `FIRST_COMMENT`/`SECOND_COMMENT` case it holds a date that has nothing to do with a hold. `SetAutocloseHoldDialog` pre-fills its picker with exactly that value (`webapp/src/features/csm-cases/components/SetAutocloseHoldDialog.tsx:62`, fed from `CsmCaseDetailPage.tsx:2682`), so the default path — open the dialog, accept the pre-filled date, click Hold — sets a genuinely new hold and then discards the note as a duplicate. Gate the read on the step:

```go
var currentCase struct {
    AutoclosureStep      *string `json:"autoclosureStep"`
    AutoclosureStateTime *string `json:"autoclosureStateTime"`
}
if err := json.Unmarshal(current, &currentCase); err == nil &&
    currentCase.AutoclosureStep != nil && *currentCase.AutoclosureStep == "ON_HOLD" &&
    currentCase.AutoclosureStateTime != nil {
    priorHoldDate = formatHoldDate(*currentCase.AutoclosureStateTime)
}
```

---

apps/csm-portal/backend/internal/handler/cases.go
```diff
@@ -1104,16 +1107,88 @@ func (h *CaseHandler) PatchCase(w http.ResponseWriter, r *http.Request) {
...
 }
 
+// formatHoldDate renders an auto-closure hold timestamp (RFC3339, as sent by the
+// FE or read back from the entity service) as the date-only form CS engineers see
+// in the UI and in the work note, since the hold is date-granularity. Falls back
+// to the raw input when it doesn't parse, so an already-invalid value is not
+// silently dropped from the comparison/note.
+func formatHoldDate(raw string) string {
+	if t, err := time.Parse(time.RFC3339, raw); err == nil {
+		return t.Format("2006-01-02")
```
shayanmalinda, line 1175: **🟡 Low: The note's date is a day late for anyone on a negative UTC offset**

The doc comment above claims this renders "the date-only form CS engineers see in the UI", but the FE sends end-of-day *local* time collapsed to a `Z`-suffixed instant (`zonedInputToUtcIso` ends in `toISOString()`, called from `SetAutocloseHoldDialog.handleSubmit`), so `time.Parse` lands in UTC and `Format` prints the UTC date rather than the picked one. Picking 2026-08-01 from a UTC-05:00 browser puts `2026-08-02T04:59:00.000Z` on the wire, and the note then reads "on-hold until 2026-08-02" while the case chip renders "On hold until Aug 1" from the same field via `formatAbsoluteForUser` — Colombo (UTC+05:30) is unaffected, which is why nothing here catches it. Either have the dialog send the offset form instead of `Z` (Go then formats in that offset and the two agree), or render the note date in the same display zone the UI uses.

---

apps/csm-portal/backend/internal/handler/cases_test.go
```diff
@@ -1364,6 +1364,130 @@ func TestPatchCase(t *testing.T) {
...
+			t.Errorf("comment type = %q, want %q", note.Type, "work_note")
+		}
+		wantContent := "Please note that this case is on-hold until 2026-08-01, hence it will not go through the auto closure process. It will be eligible for auto-closure again after this date passes, or if the case state is changed to 'Waiting on WSO2'."
+		if note.Content != wantContent {
+			t.Errorf("comment content = %q, want %q", note.Content, wantContent)
+		}
+	})
+
+	t.Run("autocloseHoldUntil PATCH does not record a duplicate work note when the hold date is unchanged", func(t *testing.T) {
+		commentCalled := false
```
shayanmalinda: **🟡 Low: These negative tests read a flag the handler writes from a goroutine**

`commentCalled` here (and again at line 1469) is a plain `bool` set inside `createCaseCommentFn`, which the handler now calls from a detached goroutine, while the test reads it from the test goroutine with no synchronisation. It's clean today only because the guard under test means the goroutine never fires — so the day that guard regresses you get a data race under `-race` rather than the clean "expected no work note" failure the test was written to give. Mirror the positive subtest's channel, or use an atomic:

```go
var commentCalled atomic.Bool
// ... commentCalled.Store(true) in the mock
if commentCalled.Load() { t.Error("expected no work note ...") }
```

---

shayanmalinda (commented):
Build and `go test -race ./internal/handler/...` both pass clean at `c0bd1b4`. The fire-and-forget shape reads deliberate and the comments explaining *why* are genuinely good, but the detached context drops the `x-user-id-token` the entity client needs to attribute the note, and the dedup key turns out not to be the field it's assumed to be. A gosec pass over the changed lines found nothing citable — flagging that `G118`, which a local scanner reports on this goroutine, is **not** a real gosec rule (absent from upstream `rules/rulelist.go`).

apps/csm-portal/backend/internal/handler/cases.go
rksk: Fixed in 02ac3d5c8: switched to `context.WithoutCancel(r.Context())` before the timeout wrap, so the caller's x-user-id-token/correlation-id survive detachment instead of being dropped. Added a regression test that captures the context's error state synchronously inside the mocked entity call (to avoid racing the WithTimeout's own deferred cleanup) and asserts it isn't Done even after the request context is canceled post-return. On the smaller shutdown-tracking point: left as-is — wiring this one fire-and-forget write into srv.Shutdown's WaitGroup felt like more surface area than a best-effort, already-logged-on-failure note warrants; a note lost in the rare window of a hold set during a rolling deploy is an acceptable trade here.

apps/csm-portal/backend/internal/handler/cases.go
rksk: Good catch, fixed in 02ac3d5c8: the dedup read now also requires autoclosureStep == "ON_HOLD" (matching the exact gate suggested) before treating autoclosureStateTime as a prior hold date to compare against. Added a test covering exactly the scenario described — a case whose autoclosureStateTime already matches the FE's pre-filled default under a non-hold step (FIRST_COMMENT) — asserting the note still fires on the first real hold.

apps/csm-portal/backend/internal/handler/cases.go
rksk: Valid, but leaving as a known limitation rather than fixing in this PR: the root cause is that `zonedInputToUtcIso` collapses the picker's local end-of-day instant to a bare `Z`-suffixed instant before it ever reaches the BFF, so this handler has no way to recover the original local calendar day — the fix has to be on the FE side (send the offset instead of `Z`, or send a date-only value), which is out of scope for this PR (deliberately BFF-only, see PR description). Filing this as a follow-up rather than blocking here, since it's Low severity and only manifests for negative-UTC-offset viewers picking a hold date near midnight.

apps/csm-portal/backend/internal/handler/cases_test.go
rksk: Fixed in 02ac3d5c8: both negative-test flags switched from a plain bool to atomic.Bool.

rksk:
@coderabbitai resume

shayanmalinda (approved):

PR #1649, opened 2026-09-02
[CSM Portal] Fix unresponsive Back button in Case view

rksk:
Resolves a corresponding CSM Portal tracking issue for: the Back button in the Case detail view was unresponsive — clicking it did nothing.

Make the Back button (and any other navigation leaving the current case, e.g. a "create new" shortcut) actually navigate away from the case, back to the dashboard or the filtered list the user came from.

Each open case tab renders behind an isolated `navigate()` override (`CaseTabIsolatedRouter`) so in-tab redirects/repairs don't leak into real browser history. That override only handled two cases: staying on the same case, or switching to a different case tab. A navigation target that isn't a case route at all (Back to dashboard/list) fell into the "same case" branch by default and only updated the tab's own local state — the real router's `location.pathname` never moved, so `CaseTabsContentHost` kept showing the case tab's content. From the user's perspective, Back did nothing.

Fix: `navigate()` now checks whether the target resolves to a case route first; if not, it forwards to the real router (`useNavTransition`) instead of touching only local override state. The tab itself stays open in the tab strip (not closed), consistent with how leaving a case via other paths already behaves.

As a CS engineer viewing a case, clicking Back returns me to the dashboard or the filtered case list I came from.

## Release note
Fix: the Back button in the Case detail view is now responsive and correctly returns to the previous view.

## Documentation
N/A — internal navigation bug fix, no user-facing behavior change beyond restoring intended behavior.

## Training
N/A

## Certification
N/A — no exam-relevant behavior change.

## Marketing
N/A

## Automation tests
 - Unit tests
   > Added a regression test (`CaseTabsIntegration.test.tsx`) that renders the real `CaseTabsProvider`/`CaseTabIsolatedRouter`/`CaseTabsContentHost` wiring and asserts clicking Back actually changes `window.location.pathname` and renders the list route. Verified the test fails without the fix and passes with it.
 - Integration tests
   > Full existing suite re-run: 251 test files / 2587 tests, all passing.

## Security checks
 - Ran FindSecurityBugs plugin and verified report? N/A — TypeScript/React frontend, not a JVM component

## Samples
N/A

## Related PRs
None

## Migrations (if applicable)
N/A

## Test environment
Verified via `tsc -b`, `eslint`, and `vitest run` in `apps/csm-portal/webapp`. Live browser click-through against a running local stack was not performed (no local topology stood up for this fix); the added integration test exercises the production component wiring as a proxy.

## Learning
Traced the bug by following the Back button's `onClick` down through `CsmCaseDetailPage` into the tab-scoped `navigate` override in `CaseTabIsolatedRouter`, and found the visibility gate in `CaseTabsContentHost` compares against the real router location, not the tab's local override state.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed navigation from case tabs so Back buttons and “create new” actions update the actual application route.
  * Ensured returning to the case list hides retained case-tab content while keeping the case tab available for later access.
  * Preserved tab-specific navigation when moving between pages within the same case.

### DIFF
--- a/apps/csm-portal/webapp/src/features/case-tabs/CaseTabsIntegration.test.tsx
+++ b/apps/csm-portal/webapp/src/features/case-tabs/CaseTabsIntegration.test.tsx
@@ -53,8 +53,10 @@ type StubSection = (typeof STUB_SECTIONS)[number];
 
 function StubCaseDetailPage() {
   const override = useCaseRouteOverride();
+  const routedNavigate = useNavigate();
   const { caseId: routedCaseId } = useParams();
   const caseId = override?.caseId ?? routedCaseId;
+  const navigate = override?.navigate ?? routedNavigate;
   // A fake "still loading" -> "loaded" transition, matching a real
   // `useGetCsmCaseDetail` query — the label should appear once this
   // resolves, without needing the user to switch tabs away and back.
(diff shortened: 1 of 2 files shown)


rksk:
@coderabbitai resume

rksk:
@coderabbitai resume

rksk:
@coderabbitai review

rksk:
@coderabbitai resume

cloby99 (approved):

PR #1655, opened 2026-09-03
[CSM Portal] Add standalone /cs-monitor-dashboard route, revert CS Overview to standard grid

Ananth-Abi:
## What

- CS Overview goes back to rendering through the standard `AgentsLandingPagePilot` grid, matching every other dashboard's styling — the Wallboard-styled version moves to its own dedicated route instead of living on `cs-overview` itself.
- New route `/cs-monitor-dashboard`: requires real Asgardeo sign-in (same flow as the rest of the portal), but renders *only* the Wallboard-styled dashboard full-screen — no header, sidebar, banners, or dashboard switcher.

## How

- `AuthGuard` gets an optional `bare` mode that reuses all its existing sign-in/redirect/token-refresh logic, just skips rendering `AppLayout` once authenticated. New `BareAuthLoader` for its own loading state, with its own `Suspense` boundary (`AppLayout` is otherwise the only place in the tree that provides one, and every route page is `React.lazy`-loaded).
- `CsmDashboardPage.tsx` no longer special-cases `cs-overview` by id.
- `WallboardDashboard`'s outer container now fills the viewport (flex column, CRE/SRE growing to fill leftover space, Security Report/FDE staying at natural height) instead of leaving dead space below a fixed-height panel. Its three render states (loading/error/loaded) share one `WallboardPageFrame` wrapper instead of each repeating the same dark full-viewport styling.
- Text sizing unified across every `WallboardStatTile` variant, CRE's secondary tier text bumped to match, and CRE's own primary 2×2 tiles get extra internal padding — visual polish from iterative feedback.
- `CLAUDE.md` documents the new route as an explicit exception to the `AuthGuard` → `FeatureRouteGuard` routing rule.

## Testing

352/352 tests passing, `tsc`/`eslint` clean. Includes a regression test proving a real `React.lazy`-loaded page under `bare` mode suspends correctly instead of crashing (caught by review before this was pushed).

## Summary by CodeRabbit

- **New Features**
  - Added a standalone `/cs-monitor-dashboard` full-screen view for the CS Overview dashboard.
  - Added organized wallboard panels for CRE, SRE, Security, and FDE metrics.
  - Added responsive metric tiles with loading, error, alert, and clickable detail states.
  - Added optional automatic dashboard and metric refresh, including a 60-second cadence for CS Overview tiles.
  - Added a simplified loading experience for full-screen authenticated views.

- **Bug Fixes**
  - Preserved the last known metric values when background refreshes fail.

- **Documentation**
  - Documented the standalone dashboard route and its full-screen behavior.


apps/csm-portal/webapp/src/features/csm-dashboard/components/WallboardDashboard.tsx
```diff
@@ -0,0 +1,309 @@
...
+          // 2x2 grid until it looked emptier, not fill the page more
+          // sensibly).
+          gridTemplateRows: "1fr auto",
+          gap: 2,
+          flex: 1,
+          minHeight: 0,
+        }}
+      >
+        {(["cre", "sre", "security", "fde"] as const).map((family) => {
+          const group = groups.find((g) => familyFor(g.section) === family);
```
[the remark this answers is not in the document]
Ananth-Abi: Good catch — fixed. `familyFor` classification now uses `groups.filter(...)` + `flatMap` to merge widgets from every section string that maps to the same family, instead of `.find()` taking only the first match. Added a regression test with two distinct "CRE ..." section names.

apps/csm-portal/webapp/src/features/csm-dashboard/components/WallboardSecondaryStat.tsx
```diff
@@ -0,0 +1,153 @@
...
+  return (
+    <Box
+      component={RouterLink}
+      to={href}
+      sx={{ display: "flex", flex: 1, textDecoration: "none", color: "inherit", height: "100%" }}
+    >
+      {tileBody}
+    </Box>
+  );
+}
```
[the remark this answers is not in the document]
Ananth-Abi: Fixed — added `awaitingCurrentUser` to the link-wrapping guard (`if (!href || isLoading || isError || awaitingCurrentUser) return tileBody;`), matching the skeleton condition above it. Added a regression test with an unresolved `__current_user__` placeholder confirming no `<a>` renders.

apps/csm-portal/webapp/src/features/csm-dashboard/components/WallboardStatTile.tsx
```diff
@@ -0,0 +1,214 @@
...
+          lineHeight: 1.2,
+          color: hasAlert ? colors.label : PLAIN_LABEL,
+        }}
+      >
+        {resolvedDisplayName}
+      </Typography>
+    </Box>
+  );
+
+  if (!href || isLoading || isError) return tileBody;
```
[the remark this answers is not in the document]
Ananth-Abi: Fixed here too, same change as the sibling `WallboardSecondaryStat.tsx` comment — added `awaitingCurrentUser` to the link-wrapping guard, with its own regression test.

apps/csm-portal/webapp/src/features/csm-dashboard/utils/wallboardMetricStyle.ts
```diff
@@ -0,0 +1,354 @@
...
+  "open engagement": "Open Engagements",
+  "open engagements": "Open Engagements",
+  "in-progress engagement": "In-Progress Engagements",
+  "in-progress engagements": "In-Progress Engagements",
+  "active engagement": "Active Engagements",
+  "active engagements": "Active Engagements",
+};
+
+function resolveFdeAlias(name: string): string {
+  const bareName = name.replace(FDE_PREFIX, "").replace(TRAILING_CASES_SUFFIX, "").trim();
```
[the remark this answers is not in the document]
Ananth-Abi: Fixed, using the exact diff you proposed — `resolveFdeAlias` now only falls back to the "Cases"-stripped name when an FDE prefix was actually present; an unrecognized, non-FDE-prefixed name (e.g. "Some Future Cases") now returns unchanged. Added a regression test for that case.

rksk (commented):
Automated review pass — correctness/policy findings below.

apps/csm-portal/webapp/src/features/csm-dashboard/components/WallboardDashboard.tsx
```diff
@@ -0,0 +1,317 @@
...
+  // structural sharing keeps the SAME `data` reference across a refetch
+  // that returns identical values, so a `data !== prevData` check would
+  // silently skip the timestamp on a real refetch that happened to change
+  // nothing. `dataUpdatedAt` updates on every successful fetch regardless,
+  // and — living in the shared query cache rather than this component's
+  // own state — survives switching away to another dashboard and back
+  // (which fully unmounts this component) without going blank in between.
+  const lastUpdated = dataUpdatedAt ? new Date(dataUpdatedAt).toLocaleTimeString() : undefined;
+
+  if (isError) {
```
rksk: `if (isError)` blanks the whole wallboard on any failed background refetch, even though React Query keeps the previously-fetched `data` around on a refetch error (it only clears `data` on the very first fetch). A single transient 5xx during the 60s auto-refresh will replace a working display with "Could not load the dashboard." on an unattended kiosk. Suggest checking `isError && !data` (or just `!data` first, matching the `isLoading || !data` branch right below) so a background error only shows once there's nothing cached to fall back on.

apps/csm-portal/webapp/src/features/csm-dashboard/components/WallboardStatTile.tsx
```diff
@@ -0,0 +1,220 @@
...
+      }}
+    >
+      {isLoading || awaitingCurrentUser ? (
+        <Skeleton
+          variant="rounded"
+          height={parseFloat(valueFontSize) * 16}
+          width="60%"
+          sx={{ bgcolor: "rgba(255,255,255,0.08)" }}
+        />
+      ) : isError ? (
```
rksk: Same pattern as WallboardDashboard.tsx: `isError` is checked ahead of any cached `data`, so a single failed poll on this tile's own `useWidgetData` call flickers a valid, already-fetched count to a dash — independent of whether the other tiles on screen are still fine. Same issue repeats in `WallboardSecondaryStat.tsx` (~line 117). Worth deciding deliberately: is a stale-but-real number preferable to a dash during a transient failure on a display meant to be glanced at from across a room?

apps/csm-portal/webapp/src/features/csm-dashboard/components/WallboardCreSection.tsx
```diff
@@ -0,0 +1,89 @@
...
+ * own fixed order — matching the original's `StatCard` 2x2 grid +
+ * `SecondaryStat` 3-column grid.
+ */
+export default function WallboardCreSection({
+  widgets,
+  selectedTeamCreGroupId,
+  selectedTeamSreGroupId,
+  selectedTeamLabel,
+}: WallboardCreSectionProps): JSX.Element {
+  const byName = new Map(widgets.map((w) => [w.displayName, w] as const));
```
rksk: `byName` keys widgets by post-alias `displayName` into a `Map`, so if two backend CRE widgets ever alias to the same display name (`resolveDisplayNameAlias` is explicitly many-to-one per its own doc comment, built because live backend naming keeps drifting), the earlier widget silently disappears from the primary grid — no warning, no test covers the collision. Given the alias table is already fragile against live-config drift, might be worth a dev-time console warning (or at least a test) when two widgets collapse to the same primary slot.

apps/csm-portal/webapp/src/layouts/AuthGuard.tsx
```diff
@@ -127,17 +142,30 @@ export default function AuthGuard(): JSX.Element {
...
+  // Computed once, used in both return branches below. In `bare` mode this
+  // is a lazy-loaded page's route element with NO ancestor `Suspense`
+  // anywhere else in the tree — `AppLayout` is the only place that
+  // provides one (see its own `<Suspense fallback={<RouteSuspenseFallback
+  // />}>`), and `bare` mode's whole point is skipping `AppLayout`. Without
+  // this, a `React.lazy` page rendered here has nothing to suspend
+  // against. Reusing `BareAuthLoader` as the fallback here too — the same
+  // "nothing but a centered progress bar" moment either way, whether the
+  // wait is for auth or for the page's own chunk to download.
+  const content = bare ? (
```
rksk: In `bare` mode, `content` skips `AppLayout` entirely, which also drops `GlobalNotificationBanner`/`TopBanner` with no substitute. `/cs-monitor-dashboard` is meant to run unattended for hours as a kiosk display — if ops raises an incident banner ("data may be stale, backend degraded") through `notificationBannerConfig`, every other authenticated route shows it, but this one never will. Worth confirming that's intentional, or wiring the banner into the bare-mode tree too.

apps/csm-portal/webapp/src/features/csm-dashboard/components/WallboardDashboard.tsx
```diff
@@ -0,0 +1,317 @@
...
+                family,
+                widgets,
+                selectedTeamCreGroupId,
+                selectedTeamSreGroupId,
+                selectedTeamLabel,
+              )}
+            </WallboardPanel>
+          );
+        })}
+        {groups
```
rksk, line 336: The fallback render for any section not matched by `familyFor` (cre/sre/security/fde) is an unbounded, natural-height row appended below the fixed 2x2 grid inside a `height: 100dvh` frame with no scroll container. If the live `cs-overview` config ever adds or renames a section that doesn't match one of the four regexes, this row has no height budget and risks overflowing/clipping on the kiosk display. `WallboardDashboard.test.tsx` only covers the four known families — might be worth a test asserting the frame stays within viewport bounds (or an explicit `overflow` policy) once a 5th section shows up.

apps/csm-portal/webapp/src/features/csm-dashboard/api/useDashboard.ts
```diff
@@ -28,9 +28,17 @@ import type { BeDashboard } from "@api/backend/types";
  *
  * Disabled while `dashboardId` is `undefined`, e.g. before the dashboard
  * list has loaded and an initial selection has been made.
+ *
+ * @param refetchIntervalMs Background auto-refetch interval in
+ * milliseconds — `undefined` (the default) means no auto-refetch at all,
```
rksk, line 33: This comment names `digiops-cs` directly ("match `digiops-cs`'s own Wallboard.tsx 60s refresh"). cs-tools is public OSS and digiops-cs is a private repo — same class of leak as the digiops-cs#2244 reference that had to be scrubbed from cs-tools PRs #1043-1045. There are several more of these across the diff (e.g. JSDoc in `CsMonitorDashboardPage.tsx`, a comment in `WallboardStatGrid.tsx`). Suggest a repo-wide sweep replacing `digiops-cs`'s `Wallboard.tsx` with something functional like "the existing wallboard reference implementation" — same information, no private-repo name.

apps/csm-portal/webapp/src/layouts/BareAuthLoader.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * `AuthGuard`'s own `loader` for a `bare` route (see `AuthGuardProps`) while
+ * the Asgardeo sign-in flow is still resolving — a full-viewport, chrome-free
+ * equivalent of `AppLayout`'s own loading state (no header/sidebar/banners,
+ * matching what the route itself renders once signed in). Kept as its own
+ * small component, not inlined in `AuthGuard`, so a `bare` route's loading
+ * state can evolve independently of the normal `AppLayout` one.
+ */
+export default function BareAuthLoader(): JSX.Element {
```
rksk, line 34: This is nearly identical to the existing `RouteSuspenseFallback` (`src/components/route-fallback/RouteSuspenseFallback.tsx`) — same `LinearProgress` markup and props, differing only in the outer `Box` sizing (`height: 100dvh; width: 100%` here vs. `flex: 1; minHeight: 0` there, since this one has no app-shell parent to fill). Could factor the shared `LinearProgress` bit into one small internal component (or have `BareAuthLoader` render `RouteSuspenseFallback` inside its own full-viewport `Box`) so a future spinner tweak doesn't need to be applied twice.

apps/csm-portal/webapp/src/features/csm-dashboard/components/WallboardDashboard.tsx
Ananth-Abi: Fixed. The error screen now only renders when there's no cached data at all — once a fetch has succeeded, a later failed background refetch keeps the last-known dashboard on screen, and the "Last updated" stamp conveys the staleness. Added a regression test (keeps rendering the last-known dashboard when a background refetch errors...).

apps/csm-portal/webapp/src/features/csm-dashboard/components/WallboardStatTile.tsx
Ananth-Abi: Fixed here and in WallboardSecondaryStat.tsx: the dash now only shows on isError && !data, so a transient poll failure keeps the last good, glanceable number on the tile. Regression test added in each (keeps showing the last good count when a background refetch fails...). Went with stale-but-real over a dash, per your framing — this is a display meant to be read from across a room.

apps/csm-portal/webapp/src/features/csm-dashboard/components/WallboardCreSection.tsx
Ananth-Abi: Added a dev-build console.warn when two CRE widgets resolve to the same display name (guarded by import.meta.env.DEV, matching featureFlags.ts's pattern), so the collision surfaces in testing rather than as a mystery missing tile in prod. Kept last-write-wins as the runtime behaviour. Test added (warns in dev when two CRE widgets collapse to the same display name).

apps/csm-portal/webapp/src/layouts/AuthGuard.tsx
Ananth-Abi: Confirmed intentional — the kiosk route is deliberately chrome-free (that's its whole purpose), and staleness is already surfaced on-screen by the "Last updated: HH:MM:SS" indicator plus the live pulse dot. If ops specifically wants the incident banner on the kiosk we can wire just GlobalNotificationBanner into the bare tree as a follow-up, but I'd rather keep this PR's bare mode strictly bare unless Chamara wants that.

apps/csm-portal/webapp/src/features/csm-dashboard/components/WallboardDashboard.tsx
Ananth-Abi: Fixed — added overflowY: "auto" to the panel grid container as an explicit overflow policy, so an unrecognized 5th+ section scrolls rather than clipping off the kiosk screen. Test asserts the policy is present (gives the panel grid an explicit overflow policy so an extra unrecognized section can't clip off the fixed-height frame).

apps/csm-portal/webapp/src/features/csm-dashboard/api/useDashboard.ts
Ananth-Abi: Swept. Replaced all 9 occurrences of "digiops-cs's Wallboard.tsx" across 8 files (useDashboard.ts, useWidgetData.ts, wallboardMetricStyle.ts, WallboardPanel.tsx, WallboardDashboard.tsx, WallboardSecondaryStat.tsx, CsmDashboardPage.tsx, CsMonitorDashboardPage.tsx) with "the reference wallboard implementation" — same information, no private-repo name. Note: LogTimeCardDialog.tsx:255 also has a digiops-cs#2821 reference, but it's pre-existing and outside this PR's diff — happy to clean it in a separate sweep.

apps/csm-portal/webapp/src/layouts/BareAuthLoader.tsx
Ananth-Abi: Refactored — BareAuthLoader now renders <RouteSuspenseFallback /> inside its own full-viewport Box, so the centered LinearProgress markup lives in exactly one place and a future spinner tweak only needs applying once.

apps/csm-portal/webapp/src/features/csm-dashboard/components/WallboardStatTile.tsx
```diff
@@ -0,0 +1,225 @@
...
+    filters,
+    "count",
+    undefined,
+    0,
+    !awaitingCurrentUser,
+    selectedTeamCreGroupId,
+    selectedTeamSreGroupId,
+    undefined,
+    currentUserId,
+    CS_OVERVIEW_REFETCH_INTERVAL_MS,
```
rksk: All wallboard tiles poll every 60s (`CS_OVERVIEW_REFETCH_INTERVAL_MS`) but funnel through the app-wide `WIDGET_FETCH_CONCURRENCY_LIMIT = 1` FIFO queue, which was tuned for a one-time mount burst, not steady-state polling. If a refresh wave (each request up to a 10s timeout) takes longer than 60s under any backend slowdown, the next wave queues behind the previous one — an unattended kiosk display degrades into a permanently-behind, thrashing poll loop instead of settling into idle between refreshes.

apps/csm-portal/webapp/src/features/csm-dashboard/components/WallboardDashboard.tsx
```diff
@@ -0,0 +1,330 @@
...
+    }
+    return <LoadingSkeleton />;
+  }
+
+  // Alias resolution runs once here, centrally, before grouping — every
+  // downstream component (WallboardCreSection, WallboardSreSection,
+  // WallboardStatGrid, the emphasis lookup) then only ever sees the
+  // already-canonical displayName, never the raw backend one.
+  const aliasedWidgets = data.widgets
+    .filter((w) => w.shape === "count")
```
rksk, line 248: This filters `data.widgets` to `shape === "count"` before grouping/rendering, unlike `CsmDashboardPage`/`AgentsLandingPagePilot`, which render every widget shape for the same `cs-overview` dashboard id. If the `cs-overview` backend config ever gains a non-count widget (list, etc.), it'll render fine on `/dashboard` but silently disappear from `/cs-monitor-dashboard` with no error or placeholder — breaking parity between the two views of the same data.

apps/csm-portal/webapp/src/features/csm-dashboard/components/WallboardStatTile.tsx
```diff
@@ -0,0 +1,225 @@
...
+  paddingY,
+  selectedTeamCreGroupId,
+  selectedTeamSreGroupId,
+  selectedTeamLabel,
+}: WallboardStatTileProps): JSX.Element {
+  const { user } = useCurrentUser();
+  const currentUserId = user?.id;
+  const awaitingCurrentUser = currentUserId === undefined && hasCurrentUserPlaceholder(filters);
+
+  const { data, isLoading, isError } = useWidgetData(
```
rksk: `WallboardStatTile` and `WallboardSecondaryStat` are near-duplicate components (~140 overlapping lines): same `useCurrentUser`/`useWidgetData` wiring, same `awaitingCurrentUser` guard, same href/link-wrapping, same skeleton/error/value branching, differing mainly in glow/emphasis styling. A future fix to the shared fetch/skeleton/error logic (e.g. last-known-value-on-error, which both already implement per their mutual comments) has to land twice, and missing one leaves the two tile types inconsistent under the same failure mode. Worth extracting the shared logic into a hook or base component.

apps/csm-portal/webapp/src/features/csm-dashboard/api/useWidgetData.ts
```diff
@@ -95,6 +95,14 @@ export function useWidgetData(
...
    * carrying that placeholder is dropped rather than sent literally. */
   currentUserId?: string,
+  /** Background auto-refetch interval in milliseconds — `undefined` (the
+   * default) means no auto-refetch at all, the behavior every existing
+   * caller had before this parameter existed. Only the CS Overview
+   * dashboard's own tiles (`WallboardStatTile`/`WallboardSecondaryStat`)
+   * pass a value here, to match the reference wallboard implementation's
+   * 60s refresh; `DashboardWidgetTile` and every other caller leaves it
+   * unset. */
+  refetchIntervalMs?: number,
```
rksk: `refetchIntervalMs` is added as a new trailing positional param on an already long, same-typed-heavy parameter list (listLimit, offset, enabled, two team-group ids, sortBy, currentUserId, now this). Callers must pass positional placeholders just to reach it, and a future same-typed param could get transposed with no type error. Consider an options object instead.

apps/csm-portal/webapp/src/layouts/AuthGuard.tsx
```diff
@@ -286,19 +306,32 @@ export default function AuthGuard(): JSX.Element {
...
   // expiring while the user does nothing at all needs no proactive fix.
+  //
+  // In `bare` mode the authenticated content is the matched route's own
+  // element with no portal chrome — no header/sidebar/banners, and none of
+  // `AuthorizedAppShell`'s `AppLayout`-wrapped loading / not-authorized
+  // states. `AppLayout` is otherwise the only place in the tree that
+  // provides a `Suspense` boundary, so `bare` mode brings its own — falling
+  // back to `BareAuthLoader`, the same centered progress bar the pending
+  // and sign-in-redirect states show.
+  const authenticatedContent = bare ? (
```
rksk, line 329: Pre-existing behavior worth flagging in this new context: the post-login deep-link redirect-restore effect (`POST_LOGIN_REDIRECT_KEY` from sessionStorage, ~L271-280) runs unchanged for this bare-mode instance too, with no exemption for a route meant to stay put indefinitely. If that key is left stale from an earlier abandoned sign-in elsewhere in the same browser (e.g. a shared kiosk machine), a fresh sign-in on `/cs-monitor-dashboard` will navigate away to that stale target instead of staying on the wallboard.

apps/csm-portal/webapp/src/features/csm-dashboard/components/WallboardStatTile.tsx
Ananth-Abi: Fixed. useWidgetData's refetchInterval is now computed per tick via resolveWidgetRefetchInterval(refetchIntervalMs, widgetFetchQueueDepth()) — while a wave is still draining it returns false, so a tile's 60s tick can't stack another wave behind the one in flight. Healthy backend → queue empty seconds into each minute, no-op; slow backend → kiosk degrades to "refreshes less often" instead of an unbounded backlog. Unit tests for the resolver and widgetFetchQueueDepth.

apps/csm-portal/webapp/src/features/csm-dashboard/components/WallboardDashboard.tsx
Ananth-Abi: The count-only filter is intentional — the monitor dashboard is a fixed count-tile design ported from the reference wallboard. But it shouldn't be silent: added a dev-build console.warn listing any non-count widgets dropped from this view (name + shape), so a cs-overview config change that adds a list/chart widget leaves a trace. Test added.

apps/csm-portal/webapp/src/features/csm-dashboard/components/WallboardStatTile.tsx
Ananth-Abi: Fixed. New useWallboardTileData hook owns the shared wiring — useCurrentUser + useWidgetData, the __current_user__ deferral, the last-known-value-on-error state derivation, and the href/link guard. Both components drop ~60 lines each and are now styling + the hook call. Direct hook test covers the loading / value / stale-on-error / deferred-user states.

apps/csm-portal/webapp/src/features/csm-dashboard/api/useWidgetData.ts
Ananth-Abi: Done — converted to a single UseWidgetDataOptions object. All 7 call sites updated; the undefined positional placeholders (DashboardWidgetPreviewPage, DashboardWidgetTile, etc.) are gone.

apps/csm-portal/webapp/src/layouts/AuthGuard.tsx
Ananth-Abi: Good catch. Fixed — the redirect-restore effect now short-circuits in bare mode: it never navigates away from the kiosk route, and it clears any key it finds (stale or its own). SignInRedirect only ever stores /cs-monitor-dashboard itself for a bare route, so nothing legitimate is lost. Regression test added (stale /cases/999 key → stays on the wallboard, key cleared).

rksk (approved):

PR #1657, opened 2026-09-03
acp-closure-service: real email sending, branded templates, security hardening (+ inert Phase 2 scaffolding)

Ananth-Abi:
## What

Sends real branded HTML emails for ACP closure notices via WSO2's
internal email service, replacing the previous log-only notifier:

- Customer-facing notices (15/7/0-day) get a branded shell — logo
  header, orange accent border, footer disclaimer.
- Internal notices (Account Owner/Renewal Manager/Technical Owner) get
  their own distinct branded shell — light-bordered white card on an
  ash-gray page background, blue greeting/intro, a bordered
  project/account detail box, footer disclaimer with the logo
  right-aligned via a table (not flexbox — Gmail doesn't reliably
  honor `display:flex` in received mail).
- The no-business-contact notice gets its own dedicated template too
  (bold red heading, bold blue intro, bolded project name and red
  italic warning clauses, a real hyperlink to the business-contact
  doc) — was previously falling through to a generic plain-text
  fallback.
- Per-notice delivery tracking: `EmailNotifier.Send` reports whether a
  specific notice was actually delivered, so a notice silently
  filtered out (e.g. a non-WSO2 recipient in staging) isn't recorded
  as sent.
- HTTPS enforcement and open-redirect protection on the email
  service client's token/API requests.
- `IS_EMAIL_SEND_ENABLED` (renamed from `SEND_REAL_EMAILS`, per review)
  gates real sending; defaults to log-only.

## Why

Requested directly, with real received-email screenshots as the
reference for both the internal and customer-facing designs. Also
addresses CodeRabbit's security findings and Rashmika's review
comments on this PR.

## Verification

- TDD throughout — extensive coverage in `email_notifier_test.go`,
  `client_test.go`, and the `sweep` package's tests.
- Real end-to-end testing against the dedicated test project
  (`e3e87599-1bc7-6650-182c-0dc5604bcb68`) via the Choreo POC: both
  customer and internal emails sending correctly, logo loading (after
  tracking down a Gmail image-transparency-flattening quirk and a
  flexbox layout quirk), idempotency re-confirmed.
- `gofmt` / `go vet` / `go test -race ./...` all clean.

## Note on this PR's diff

This PR also includes early Phase 2 (invoice-based closure) work —
`closure.DecideInvoice`, `suspensionstate`'s invoice read/write path,
the invoice notice templates, and a `dueInvoice` placeholder struct.
This is **inert code, not wired into `sweep.Run`** — it doesn't touch
anything reviewed so far and has zero effect on the currently-running
service. It rode along here because it was committed on this same
branch; nothing to review differently because of it.

## Summary by CodeRabbit

* **New Features**
  * Added real email delivery through WSO2’s authenticated email service, with production safety controls and recipient filtering.
  * Added branded HTML notifications for customer, internal, invoice, and missing-contact scenarios.
  * Added invoice-based closure notices, grace-period handling, and independent tracking of invoice-related notification state.

* **Bug Fixes**
  * Notification records now reflect whether each message was actually delivered, including filtered recipients.

* **Documentation**
  * Updated setup and operational guidance for email delivery, configuration, templates, and testing.


Ananth-Abi:
@coderabbitai review

integrations/acp-closure-service/internal/emailservice/client.go
```diff
@@ -0,0 +1,189 @@
...
+
+// NewClient constructs a Client that authenticates against the email
+// notification service using the OAuth2 client credentials grant type,
+// mirroring internal/entity.NewClient's identical setup for
+// csm-integration-service.
+func NewClient(cfg Config) *Client {
+	cc := clientcredentials.Config{
+		ClientID:     cfg.ClientID,
+		ClientSecret: cfg.ClientSecret,
+		TokenURL:     cfg.TokenURL,
```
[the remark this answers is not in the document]
Ananth-Abi: Fixed — NewClient now returns an error if TokenURL/BaseURL don't use https://, checked before the client is ever constructed. Added a loopback exception (127.0.0.1/localhost) since that traffic never leaves the machine — it's also what every test in this package already uses via httptest.NewServer. Also found the same gap exists in internal/entity.Client (the older, already-merged client) — left that one unfixed here to keep this PR scoped, but flagged it explicitly in CLAUDE.md as a tracked follow-up rather than letting it go unnoticed.

integrations/acp-closure-service/internal/emailservice/client.go
```diff
@@ -0,0 +1,189 @@
...
+		// No Scopes: confirmed with Rashmika that this token endpoint,
+		// unlike csm-integration-service's, requires none.
+	}
+
+	tokenCtx := context.WithValue(context.Background(), oauth2.HTTPClient,
+		&http.Client{Timeout: tokenFetchTimeout})
```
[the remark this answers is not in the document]
Ananth-Abi: ▎ Fixed — moved CheckRedirect onto the http.Client actually embedded in tokenCtx (the one that performs the token-fetch request carrying the real client secret), not just the one returned for regular API calls after login. Added TestTokenFetchRejectsRedirects, confirmed red against the old code first — the token endpoint's redirect was being followed and the real request would reach the redirect target.

integrations/acp-closure-service/internal/notify/email_notifier.go
```diff
@@ -0,0 +1,194 @@
...
+// already established throughout this codebase for absent recipients.
+func (n *EmailNotifier) Send(ctx context.Context, notice Notice) error {
+	to, cc := recipientsToToCC(notice.Recipients)
+	to = n.filterRecipients(to)
+	cc = n.filterRecipients(cc)
+
+	if len(to) == 0 {
+		n.Logger.InfoContext(ctx, "email skipped: no valid recipients",
+			"projectID", notice.ProjectID, "window", notice.Window, "subject", notice.Subject)
+		return nil
```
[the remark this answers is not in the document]
Ananth-Abi: Fixed — reworked notifier.Send to report delivery per call ((delivered bool, err error)) instead of relying on a blanket per-notifier Delivers() signal. EmailNotifier.Send now returns false (not an error) when every recipient gets filtered out, and sweep.notifyForWindow ANDs the delivered result across every notice it sends for a window before deciding whether to record "SUCCESSFUL". Added TestProcessProject_RecordsIgnoredWhenOnlyCustomerNoticeWasntDelivered reproducing the exact scenario you flagged — internal notice delivers, customer notice gets filtered — confirming the window is now correctly recorded as "IGNORED", not "SUCCESSFUL".

integrations/acp-closure-service/internal/notify/email_notifier.go
```diff
@@ -0,0 +1,194 @@
...
+		// branded look to both was wrong.
+		htmlBody = renderEmailHTML(notice.Body)
+	}
+
+	if err := n.Sender.SendEmail(ctx, to, cc, notice.Subject, htmlBody); err != nil {
+		return fmt.Errorf("send email: %w", err)
+	}
+
+	n.Logger.InfoContext(ctx, "email sent",
+		"projectID", notice.ProjectID, "window", notice.Window, "subject", notice.Subject, "to", to, "cc", cc)
```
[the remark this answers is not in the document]
Ananth-Abi: ▎ Fixed — the "email sent" log line now records toCount/ccCount instead of the actual to/cc address lists. Added a test asserting no log attribute on that line contains a real recipient address.

Ananth-Abi:
@coderabbitai review

Ananth-Abi:
@coderabbitai full review

Ananth-Abi:
@coderabbitai full review

integrations/acp-closure-service/cmd/acp-closure/main.go
```diff
@@ -67,11 +82,29 @@ func main() {
...
+		if err != nil {
+			slog.Error("invalid email service configuration", "err", err)
+			os.Exit(1)
+		}
+		ntf = &notify.EmailNotifier{
+			Sender:                 emailClient,
+			Logger:                 slog.Default(),
+			AllowNonWSO2Recipients: envBool("EMAIL_SERVICE_ALLOW_NON_WSO2_RECIPIENTS", false),
+		}
+	}
```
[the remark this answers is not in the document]
Ananth-Abi: Leaving this one for now — in practice SEND_REAL_EMAILS=true is only ever set alongside DRY_RUN=false when actually running this against real recipients, so the conflicting-flag state you're describing doesn't come up in how this gets run today. Worth enforcing in code eventually so a future misconfiguration can't silently hit this, but treating it as a deliberate scope call for this PR rather than blocking on it now.

integrations/acp-closure-service/.env.example
```diff
@@ -32,3 +32,33 @@ TEST_PROJECT_ID=
...
 # to hold real entries in dev/staging.
 EXCLUDED_PROJECT_IDS=
+
+# Safety toggle, same convention as DRY_RUN: defaults to false (log-only,
+# via notify.LoggingNotifier) on anything except an explicit "true". When
+# true, notify.EmailNotifier is used instead, which genuinely calls the
+# real email notification service — see CLAUDE.md's "Real email sending"
+# section before ever setting this true anywhere but a deliberate
+# production rollout.
+SEND_REAL_EMAILS=false
```
Rashmika998: Let's change this to `IS_EMAIL_SEND_ENABLED`

integrations/acp-closure-service/CLAUDE.md
```diff
@@ -286,8 +297,95 @@ wrong answer:
   testing against real data shows this role is rarely configured in
   practice regardless — most real resolutions land on `primary_contact` or
   `am_nudge`, not `business_contact`.
-- **Real email-sending mechanism** — deferred pending message-queue design
-  on the entity-service side, not blocked on this component's own work.
+- **`internal/entity.Client` doesn't validate its configured URLs use
+  `https`** — the same gap `internal/emailservice.Client.NewClient` was
+  given a fix for (CodeRabbit, PR #1657; see `requireHTTPS` there).
```
Rashmika998: We don't need to mention PRs in this file. Can you please check throughout the code?

integrations/acp-closure-service/CLAUDE.md
```diff
@@ -286,8 +297,95 @@ wrong answer:
...
+  the same way the real service expects — sending it as a plain string
+  would not match what the server decodes. `bcc`/`replyTo`/`attachments`
+  exist on the real API but have no ACP use case, so they're left out of
+  this component's client entirely rather than plumbed through unused.
+- **No OAuth2 scope is required** for this token endpoint — confirmed
+  explicitly with Rashmika, unlike `csm-integration-service`'s
+  `CSM_INTEGRATION_SCOPES`. Don't add a scopes config value here without
+  re-confirming that's changed.
+- **`FromAddress` is fixed at config level**, not a per-`Notice` value —
+  confirmed via a real received email to be `no-reply@wso2.com`.
```
Rashmika998: Let's remove this as well

integrations/acp-closure-service/.env.example
Ananth-Abi: Done — renamed to IS_EMAIL_SEND_ENABLED everywhere: .env.example, CLAUDE.md, and main.go (including the local Go variable and its log field, so the code stays consistent with the new name, not just the env var string itself).

integrations/acp-closure-service/CLAUDE.md
Ananth-Abi: Removed, and checked throughout the code as you asked — found 8 more PR #<number> mentions scattered across main.go, internal/emailservice, internal/sweep, and internal/closure (mostly in comments citing which review a fix came from). Dropped the PR number from all of them, kept the reviewer/CodeRabbit attribution where it was already there since you only flagged the PR mentions specifically.

integrations/acp-closure-service/CLAUDE.md
Ananth-Abi: Trimmed the "confirmed via a real received email" verification note, kept the actual fact (no-reply@wso2.com) — let me know if you meant the whole bullet point instead, happy to drop that too.

Ananth-Abi:
@coderabbitai review

integrations/acp-closure-service/internal/closure/decide_invoice_test.go
```diff
@@ -0,0 +1,190 @@
...
+
+	w0 := NoticeWindow0
+
+	got := DecideInvoice(now, invoiceDate, invoiceDueDate, 3.4, false, &w0)
+
+	if got.ShouldNotify {
+		t.Error("ShouldNotify = true, want false — day-0 notice already recorded, should not refire")
+	}
+	if !got.ShouldSuspend {
+		t.Error("ShouldSuspend = true, want false — suspend has no idempotency signal here, must still report true")
```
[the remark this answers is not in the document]
Ananth-Abi: Fixed — the message was backwards (said "= true, want false" while inside the !got.ShouldSuspend branch, where it's actually false). Corrected to "ShouldSuspend = false, want true".

integrations/acp-closure-service/internal/suspensionstate/suspensionstate.go
```diff
@@ -20,10 +20,12 @@
...
 // e3e87599-1bc7-6650-182c-0dc5604bcb68) and this component's own
-// closure.NoticeWindow. Phase 1 only ever reads or writes the
-// based_on_subscription_end_date key; based_on_due_invoices and
-// based_on_compliance belong to Phase 2 and must survive every write
-// untouched.
+// closure.NoticeWindow. Phase 1 reads/writes based_on_subscription_end_date;
+// Phase 2 reads/writes based_on_due_invoices the same way, via its own
+// parallel functions below — each track has its own independent
+// idempotency signal, and a write to one key always preserves every other
+// key byte-for-byte. based_on_compliance is never read or written by this
```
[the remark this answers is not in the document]
Ananth-Abi: Confirmed this is real — verified independently with a standalone script before trusting it, then reproduced with a failing test: encoding/json's default marshaling HTML-escapes <, >, & in any output, including bytes from an untouched json.RawMessage. Fixed with the lighter option rather than the full byte-splicing rewrite you suggested: withState now marshals through a json.Encoder with SetEscapeHTML(false) instead of json.Marshal. Closes the escaping gap with no behavior change; key ordering (alphabetical, from Go's map marshaling) is left as-is since JSON key order was never semantically meaningful to begin with. Affects both WithSubscriptionEndDateState and WithDueInvoicesState, since both share this helper.

integrations/acp-closure-service/CLAUDE.md
Rashmika998: I believe we don't need to mention from where we send the emails in the code itself

Rashmika998: @rksk WDYT?

rksk: +1, no need to mention the email addresses (config values) in the repo

rksk (approved):

PR #1659, opened 2026-09-03
[CSM] Sync main into dev-app-csm-portal

SasmithaDilshan:
Syncs `main` back into `dev-app-csm-portal` after #1636 merged.

`main` is 397 commits ahead of `dev`; `dev` is 14 ahead. Without this, every new
branch cut from `dev` starts 397 commits behind and re-resolves the same drift.

## Conflicts — 4 files

**The entity-service three are one decision, not three.** `main` and `dev` answered
the same review question in opposite directions:

- `main` — `487226347` *"Document that Postgres is required for every DATA_SOURCE"*
- `dev` — `ac8787941` *"make the database optional when `DATA_SOURCE=servicenow"* (#1647)

`dev`'s is the newer and more complete answer: it makes the database genuinely
optional (`Config.HasDatabase`, no pool opened, the two Postgres-only feature sets
unregistered) rather than documenting the requirement as permanent. Taking `dev`'s
side keeps ServiceNow-mode deployments from needing DB credentials they never use.

- `entity-service/internal/config/config.go`
- `entity-service/.env.example`
- `entity-service/internal/domain/entity.go`

**The fourth is additive:**

- `apps/csm-portal/backend/internal/handler/response.go` — both sides added error
  message constants. Union them; a duplicate const is a compile error, so the
  compiler will catch a mistake here, while dropping one is silent.

## Note on merge direction

`main` already contains everything from #1636, so this brings that work back to
`dev` along with the 397 commits `main` accumulated alongside it. `dev`'s own 14
commits (the `#1647` optional-database fix, the SFTPGo attachment authz work, the
scheduled-tasks open-cases report) are untouched by this merge and stay on `dev`.


SasmithaDilshan:
Superseded — replaced by a sync branch with the four conflicts resolved. Resolving them on this PR would have meant merging dev into main and pushing a merge commit to the trunk, since GitHub resolves conflicts on the head branch (main here). Follows the existing sync/main-into-dev-app-csm-portal pattern from #1335.

PR #1668, opened 2026-09-07
[Customer Portal] Implement Role-Based Access Control (RBAC) across backend and webapp

Hesara2003:
## Purpose
Implement full-stack Role-Based Access Control (RBAC) across the Customer Portal Go backend (`backend-v2`) and React webapp (`webapp`). Enforces an 8-module by 5-persona CRUD permission matrix, secures REST endpoints with thread-safe in-memory caching, filters navigation layout items, and gates frontend routes with standardized 403 Forbidden handling.

## Goals
- Enforce backend role gating across Cases, Time Cards, Change Requests, Deployments, Products, Resources, and Security Admin endpoints based on canonical persona permissions.
- Provide thread-safe in-memory role resolution caching (5-minute TTL) in the Go BFF to eliminate upstream latency and avoid redundant identity provider lookups.
- Restrict contact management operations (`POST`, `DELETE`, `PATCH roles`) to `admin`, `customer_admin`, `partner_admin`, and `super_admin`.
- Provide `useCustomerPermissions` hook in the React webapp to normalize user roles and evaluate permissions against the identical matrix.
- Guard frontend routes with `CustomerRoleGuard`, rendering `Error403Page` wrapped in `ErrorLayout` for unauthorized access attempts.
- Dynamically filter `SideBar` navigation items, hiding Security Center for non-super admins and Operations for stakeholders.

## Approach
- **Go Backend (`apps/customer-portal/backend-v2`)**:
  - Implemented `NormalizeRoles` in `internal/middleware/rbac.go` to map raw ServiceNow wire roles to canonical roles (`super_admin`, `admin`, `agent`, `customer_admin`, `customer_user`, `partner_admin`, `partner_user`, `stakeholder`).
  - Implemented `CachedRoleResolver` with in-memory TTL caching (5 minutes per user ID) using `sync.RWMutex`.
  - Implemented `RequirePermission(resolver, module, action)` and `RequireRoles(resolver, roles...)` middleware.
  - Gated 21 REST API endpoints in `cmd/server/main.go` covering Cases, Time Cards, Change Requests, Deployments, Products, Contacts, and Security Vulnerabilities.
  - Added unit tests in `internal/middleware/rbac_test.go` and integration tests in `cmd/server/main_rbac_test.go`.
- **React Webapp (`apps/customer-portal/webapp`)**:
  - Implemented `useCustomerPermissions` hook (`src/hooks/useCustomerPermissions.ts`) evaluating `can(module, action)` and role convenience flags (`canAccessSecurityAdmin`, `canCreateCase`, `isStakeholder`, etc.).
  - Implemented `CustomerRoleGuard` (`src/layouts/CustomerRoleGuard.tsx`) supporting both layout routes (`<Outlet />`) and element wrappers (`children`), with standard `Error403Page` fallback.
  - Updated `SideBar.tsx` to conditionally hide Security Center when `!canAccessSecurityAdmin` and Operations when `isStakeholder`.
  - Wired `CustomerRoleGuard` in `App.tsx` across `security-center`, `operations/change-requests`, `support/change-requests`, and case creation flows (`describe-issue`, `create-case`, `create-related-case`, `security-report/create`).
  - Added unit tests for `useCustomerPermissions`, `CustomerRoleGuard`, and `SideBar`.

## User stories
- As a **Super Admin**, I have full CRUD access across all 8 modules, including the Security Admin module.
- As an **Admin**, I have full CRUD access across 7 operational modules, with zero access to Security Admin.
- As an **Agent**, I have CRU access on Cases, Time Cards, and Change Requests, full CRUD on Deployments, and read/update on Projects.
- As an **External User** (*Customer User, Customer Admin, Partner User, Partner Admin*), I have full CRUD access to Deployments, CRU access to Cases, and read-only access to Time Cards, Projects, and Change Requests.
- As an **Administrator** (*Admin, Customer Admin, Partner Admin, Super Admin*), I can manage project contacts.
- As a **Stakeholder**, I have strictly read-only access to Cases, Projects, and Deployments, with no access to Time Cards, Change Requests, or Security Admin.

## Release note
Implement full-stack Role-Based Access Control (RBAC) across Customer Portal backend endpoints and webapp routes.

## Documentation
N/A — internal Customer Portal feature; no public doc surface affected.

## Automation tests
- Unit tests:
  - Backend: `internal/middleware/rbac_test.go` (100% pass across all personas, actions, TTL caching, and error mappings).
  - Backend: `cmd/server/main_rbac_test.go` (100% pass across route gating and contact management).
  - Frontend: `src/hooks/__tests__/useCustomerPermissions.test.ts` (9/9 pass).
  - Frontend: `src/layouts/__tests__/CustomerRoleGuard.test.tsx` (4/4 pass).
  - Frontend: `src/components/side-nav-bar/__tests__/SideBar.test.tsx` (12/12 pass).
- Build verification:
  - Backend: `go test -v -race ./...` (100% pass).
  - Frontend: `tsc -b && vite build` (clean pass in 28.77s).

## Security checks
- Ran FindSecurityBugs plugin and verified report? N/A (Go backend and React frontend; not a Java/FindSecurityBugs target)

## Test environment
- macOS (ARM64), Node.js v20+, Go 1.22 toolchain
- `go test -v -race ./...` passing locally
- `tsc -b && vite build` passing locally

## Summary by CodeRabbit

* **New Features**
  * Added role-based access controls across customer portal pages and backend APIs.
  * Restricted access to cases, change requests, security features, time cards, deployments, contacts, and related operations based on user roles.
  * Added permission-aware navigation, including visibility rules for Operations and Security Center.
  * Added loading, error, and forbidden-access states for protected pages.
* **Tests**
  * Added coverage for permission checks, role normalization, route protection, caching, and navigation visibility.


apps/customer-portal/backend-v2/internal/middleware/rbac.go
```diff
@@ -0,0 +1,343 @@
...
+	}
+
+	roles := NormalizeRoles(resp.Roles)
+
+	r.mu.Lock()
+	r.cache[user.UserID] = cachedRoleEntry{
+		roles:     roles,
+		expiresAt: time.Now().Add(r.ttl),
+	}
+	r.mu.Unlock()
```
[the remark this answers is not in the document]
Hesara2003: Addressed in commit bbde5da7c: added write-time sweep to evict expired cache entries from `CachedRoleResolver.cache` and added a unit test verifying the eviction behavior.

apps/customer-portal/backend-v2/cmd/server/main.go
```diff
@@ -208,59 +209,59 @@ func main() {
...
+	mux.Handle("POST /cases", middleware.RequirePermission(roleResolver, middleware.ModuleCases, middleware.ActionCreate)(http.HandlerFunc(caseHandler.CreateCase)))
+	mux.Handle("PATCH /cases/{id}", middleware.RequirePermission(roleResolver, middleware.ModuleCases, middleware.ActionUpdate)(http.HandlerFunc(caseHandler.PatchCase)))
+	mux.Handle("POST /cases/{id}/comments", middleware.RequirePermission(roleResolver, middleware.ModuleCases, middleware.ActionCreate)(http.HandlerFunc(caseHandler.CreateCaseComment)))
+	mux.Handle("POST /cases/{id}/activities/search", middleware.RequirePermission(roleResolver, middleware.ModuleCases, middleware.ActionRead)(http.HandlerFunc(caseHandler.SearchCaseActivities)))
+	mux.Handle("GET /cases/{id}/attachments", middleware.RequirePermission(roleResolver, middleware.ModuleCases, middleware.ActionRead)(http.HandlerFunc(caseHandler.SearchCaseAttachments)))
+	mux.Handle("POST /cases/{id}/attachments", middleware.RequirePermission(roleResolver, middleware.ModuleCases, middleware.ActionCreate)(http.HandlerFunc(caseHandler.CreateCaseAttachment)))
```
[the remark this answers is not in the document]
Hesara2003: Addressed in commit 193471caf: wrapped `GET /cases/{id}/feedback` with `ModuleCases, ActionRead` and `POST /cases/{id}/feedback` with `ModuleCases, ActionCreate` using `RequirePermission`.

apps/customer-portal/backend-v2/cmd/server/main_rbac_test.go
```diff
@@ -0,0 +1,267 @@
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			resolver := middleware.NewCachedRoleResolver(&mockEntityClient{roles: tt.roles}, time.Minute)
+			handler := middleware.RequirePermission(resolver, tt.module, tt.action)(dummyHandler)
+
+			req := httptest.NewRequest(http.MethodGet, "/test", nil)
+			req = req.WithContext(middleware.WithUserInfo(req.Context(), &middleware.UserInfo{UserID: "usr-test"}))
```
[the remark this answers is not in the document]
Hesara2003: Addressed in commit 820a899b3: replaced `httptest.NewRequest` with `httptest.NewRequestWithContext` passing the user info context directly.

Hesara2003:
All review comments addressed:
1. **Role cache eviction on write**: Commits `31a81902e` & `bbde5da7c` added write-time eviction of expired role cache entries and corresponding unit tests.
2. **Case feedback route authorization**: Commit `193471caf` gated `GET /cases/{id}/feedback` with `ModuleCases, ActionRead` and `POST /cases/{id}/feedback` with `ModuleCases, ActionCreate`.
3. **`httptest.NewRequestWithContext`**: Commit `820a899b3` replaced `httptest.NewRequest` calls with `httptest.NewRequestWithContext`.
4. **Operations landing page route guard**: Commit `bd0ba0ada` wrapped only the Operations index route (`/projects/:projectId/operations` -> `<OperationsPage />`) with `<CustomerRoleGuard module="change_requests" action="read">` while keeping `service-requests` outside the guard. Commit `b7a131b57` added unit tests in `CustomerRoleGuard.test.tsx` verifying blocked index access and permitted child route access.

apps/customer-portal/backend-v2/cmd/server/main.go
```diff
@@ -200,66 +201,66 @@ func main() {
...
 	mux.HandleFunc("POST /projects/{id}/instances/stats/usages/search", instanceHandler.SearchProjectInstanceUsageStats)
-	mux.HandleFunc("POST /projects/{id}/registry-tokens", registryHandler.CreateRegistryToken)
-	mux.HandleFunc("POST /projects/{id}/registry-tokens/search", registryHandler.SearchRegistryTokens)
-	mux.HandleFunc("GET /projects/{id}/integration-users", registryHandler.GetProjectIntegrationUsers)
+	mux.Handle("POST /projects/{id}/registry-tokens", middleware.RequirePermission(roleResolver, middleware.ModuleDeploymentResources, middleware.ActionCreate)(http.HandlerFunc(registryHandler.CreateRegistryToken)))
+	mux.Handle("POST /projects/{id}/registry-tokens/search", middleware.RequirePermission(roleResolver, middleware.ModuleDeploymentResources, middleware.ActionRead)(http.HandlerFunc(registryHandler.SearchRegistryTokens)))
+	mux.Handle("GET /projects/{id}/integration-users", middleware.RequirePermission(roleResolver, middleware.ModuleDeploymentResources, middleware.ActionRead)(http.HandlerFunc(registryHandler.GetProjectIntegrationUsers)))
 	mux.HandleFunc("GET /projects/{id}/contacts", contactHandler.GetProjectContacts)
```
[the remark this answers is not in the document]
Hesara2003: Skipping this one. Applying the admin guard to the read would break Settings > User Management for every non admin user.

The tabs list is not role gated (SettingsPage.tsx line 69) and the Users tab renders for everyone (line 117). Only the controls are gated, via canAddOrRemoveUsers={isCustomerAdmin && !isRestricted} on line 120. So a regular user sees the contact list today and would get a 403 instead.

The asymmetry is deliberate rather than an oversight: reading who is on your own project is a normal user action, and the list is already scoped to projects the caller can access. Creating, removing and rerolling a contact are the admin actions, and those four routes are guarded.

If the team does want the read restricted, it needs the webapp change in the same PR, hiding the tab for non admins, otherwise the page just fails for them.

Rashmika998:
@coderabbitai review

Rashmika998 (commented):
## Summary

Solid, well-structured RBAC implementation with good test coverage, clean separation between backend middleware and frontend hook, and thorough documentation in commit messages. A few issues to address: one duplicate test case that masks a missing scenario, and a semantic mismatch between a token-regenerate endpoint and `ActionDelete`. Minor suggestions on guard defensiveness, navigation UX consistency, and cache-eviction overhead.

---

apps/customer-portal/backend-v2/cmd/server/main_rbac_test.go
```diff
@@ -0,0 +1,287 @@
...
+			name:         "PartnerAdmin can manage contacts",
+			roles:        []string{"sn_customerservice.partner_admin"},
+			wantHTTPCode: http.StatusOK,
+		},
+		{
+			// A role outside the allow-list no longer slips through: the
+			// blanket super_admin grant in RequireRoles is gone.
+			name:         "CustomerUser cannot manage contacts",
+			roles:        []string{"sn_customerservice.customer"},
+			wantHTTPCode: http.StatusForbidden,
```
Rashmika998: **[Required]** Duplicate test case name and configuration. This entry is byte-for-byte identical to the test at line 237 (`"CustomerUser cannot manage contacts"`, roles `sn_customerservice.customer`, want 403). The comment just above it (`// A role outside the allow-list no longer slips through...`) suggests it was supposed to cover a *different* scenario. Based on the surrounding cases, the missing scenario is likely **PartnerUser cannot manage contacts** — change `name` to `"PartnerUser cannot manage contacts"` and `roles` to `[]string{"sn_customerservice.partner"}`.

---

apps/customer-portal/backend-v2/cmd/server/main.go
```diff
@@ -200,66 +201,66 @@ func main() {
...
+	mux.Handle("DELETE /projects/{id}/contacts/{email}", middleware.RequireRoles(roleResolver, middleware.RoleAdmin, middleware.RoleCustomerAdmin, middleware.RolePartnerAdmin)(http.HandlerFunc(contactHandler.RemoveProjectContact)))
+	mux.Handle("PATCH /projects/{id}/contacts/{email}", middleware.RequireRoles(roleResolver, middleware.RoleAdmin, middleware.RoleCustomerAdmin, middleware.RolePartnerAdmin)(http.HandlerFunc(contactHandler.UpdateProjectContactRole)))
+	mux.Handle("POST /projects/{id}/contacts/validate", middleware.RequireRoles(roleResolver, middleware.RoleAdmin, middleware.RoleCustomerAdmin, middleware.RolePartnerAdmin)(http.HandlerFunc(contactHandler.ValidateProjectContact)))
+	mux.Handle("DELETE /registry-tokens/{id}", middleware.RequirePermission(roleResolver, middleware.ModuleDeploymentResources, middleware.ActionDelete)(http.HandlerFunc(registryHandler.DeleteRegistryToken)))
+	mux.Handle("POST /registry-tokens/{id}/regenerate", middleware.RequirePermission(roleResolver, middleware.ModuleDeploymentResources, middleware.ActionDelete)(http.HandlerFunc(registryHandler.RegenerateRegistryToken)))
```
Rashmika998: **[Required]** `POST /registry-tokens/{id}/regenerate` is gated with `ActionDelete`, but regenerating a token is a create/update operation — it produces a new credential, it does not remove one. Using `ActionDelete` accidentally ties regeneration access to deletion access; if those are ever split in the matrix (e.g. an agent who can regenerate but not delete), this will silently enforce the wrong check. Change to `ActionCreate` (or `ActionUpdate`) to match the actual semantic of the endpoint.

---

apps/customer-portal/webapp/src/layouts/CustomerRoleGuard.tsx
```diff
@@ -0,0 +1,101 @@
...
+        }}
+      >
+        <Typography color="text.secondary" variant="body2">
+          Unable to verify access permissions. Please refresh and try again.
+        </Typography>
+      </Box>
+    );
+  }
+
+  let isAuthorized = true;
```
Rashmika998, line 102: **[Suggestion]** When `CustomerRoleGuard` is mounted without a `module`/`action` pair **and** without `requiredRoles`, `isAuthorized` stays `true` and every child/outlet passes through unchecked. The same silent passthrough happens when only one of `module` or `action` is supplied (e.g. `<CustomerRoleGuard module="cases">`). A misconfigured callsite will look correct but enforce nothing. Consider adding a `DEV`-mode invariant:
```tsx
if (process.env.NODE_ENV !== 'production') {
  if (!((module && action) || (requiredRoles && requiredRoles.length > 0))) {
    console.warn('CustomerRoleGuard: no constraints specified — all access is granted');
  }
}
```

---

apps/customer-portal/webapp/src/App.tsx
```diff
@@ -145,7 +146,17 @@ export default function App(): JSX.Element {
                   />
                   {/* Operations */}
                   <Route path="operations">
-                    <Route index element={<OperationsPage />} />
+                    <Route
+                      index
+                      element={
+                        <CustomerRoleGuard
```
Rashmika998: **[Suggestion]** The Operations index (`OperationsPage`) is gated on `change_requests:read`, but the `service-requests` child route has no guard. A user whose roles grant SR access but not CR access will be blocked at the landing page, yet can still reach `operations/service-requests` by direct URL or deep-link. If the intent is to hide the Operations section entirely from CR-restricted users, the guard should sit on the `<Route path="operations">` wrapper rather than only on the index element. If SR-only users should see Operations, the index guard should use a less restrictive check (e.g. `can('change_requests','read') || canAccessSR`).

---

apps/customer-portal/backend-v2/internal/middleware/rbac.go
```diff
@@ -0,0 +1,355 @@
...
+
+	resp, err := r.client.GetMe(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("rbac: failed to resolve user roles: %w", err)
+	}
+
+	roles := NormalizeRoles(resp.Roles)
+
+	r.mu.Lock()
+	now := time.Now()
```
Rashmika998, line 266: **[Suggestion]** The eviction loop scans **every** cache entry on every write, while holding the write lock. Under concurrent load with many active users this O(n) sweep (under `mu.Lock`) can cause lock contention and latency spikes for all in-flight requests. A few lighter alternatives:
- Accept the memory growth (expired entries are harmlessly skipped on read; the set is bounded by active user count).
- Run a background goroutine with a ticker to sweep periodically.
- Evict lazily: delete the stale entry discovered on the *read* path (requires an upgrade from `RLock` → `Lock`, but avoids the write-path sweep).

If cache size is expected to be small (< a few hundred concurrent users), this is fine as-is — worth a comment explaining the tradeoff.

---

apps/customer-portal/webapp/src/hooks/useCustomerPermissions.ts
```diff
@@ -0,0 +1,374 @@
...
+      can,
+      hasRole,
+      hasAnyRole,
+      canCreateCase: can("cases", "create"),
+      canUpdateCase: can("cases", "update"),
+      canDeleteCase: can("cases", "delete"),
+      canAccessTimeCards: can("time_cards", "read"),
+      canAccessChangeRequests: can("change_requests", "read"),
+      canAccessSecurityAdmin: can("security_admin", "read"),
+      canManageContacts: hasAnyRole([
```
Rashmika998, line 363: **[Nitpick]** Indentation inconsistency inside the `hasAnyRole([...])` call: `"admin"` is indented 10 spaces while `"customer_admin"` and `"partner_admin"` use 8 spaces. Align all three to the same indent level.

---

apps/customer-portal/webapp/src/hooks/useCustomerPermissions.ts
```diff
@@ -0,0 +1,374 @@
...
+    case "partner_user":
+      return "partner_user";
+    // snc_internal is the ServiceNow wire form of internal; both resolve to the
+    // same persona so access no longer depends on which data source
+    // entity-service runs. "internal" carries the grants snc_internal already
+    // had via agent.
+    case "snc_internal":
+    case "internal":
+      return "internal";
+    default:
```
Rashmika998, line 87: **[Nitpick]** `return trimmed as CanonicalRole` casts an arbitrary string to the canonical role union type. The cast is functionally harmless (the string won't match any matrix entry and will grant nothing), but it defeats TypeScript's exhaustiveness checking — the compiler won't warn if a new `CanonicalRole` literal is added to the union but not to the switch. Consider either returning a dedicated sentinel (e.g. `return "unrecognized" as CanonicalRole`) or annotating the default branch with a comment that explains the deliberate passthrough, so future readers don't silently add a new role and wonder why it doesn't work.

---

apps/customer-portal/backend-v2/cmd/server/main_rbac_test.go
Hesara2003: Deleted it instead. PartnerUser cannot manage contacts already exists as the second case, so renaming would have been a duplicate too. Both scenarios were already covered.

apps/customer-portal/backend-v2/cmd/server/main.go
Hesara2003: Changed to ActionUpdate. The token survives regeneration, only its secret changes, so it is an update rather than a create.

apps/customer-portal/backend-v2/internal/middleware/rbac.go
Hesara2003: Added a comment rather than changing it. The sweep runs only on a cache miss, not every request, so it is roughly one sweep per user per TTL. Noted a background ticker as the move if the user count ever grows.

apps/customer-portal/webapp/src/layouts/CustomerRoleGuard.tsx
Hesara2003: Added, as a dev only warning. Covers both shapes: no constraints at all, and only one half of a module/action pair. Vite strips it from production.

apps/customer-portal/webapp/src/App.tsx
Hesara2003: Moved the guard to the operations wrapper so it covers every child. There is no service_requests module in the matrix, so change_requests:read governs the whole section.

apps/customer-portal/webapp/src/hooks/useCustomerPermissions.ts
Hesara2003: Fixed. That was left over from removing super_admin from the list.

apps/customer-portal/webapp/src/hooks/useCustomerPermissions.ts
Hesara2003: Added a comment on the default branch explaining the deliberate passthrough, and noting that adding a role to the union will not make the compiler flag this switch.

apps/customer-portal/webapp/src/App.tsx
```diff
@@ -229,7 +278,15 @@ export default function App(): JSX.Element {
...
-                  <Route path="security-center">
+                  <Route
+                    path="security-center"
+                    element={
+                      <CustomerRoleGuard
+                        module="security_admin"
+                        action="read"
+                      />
+                    }
+                  >
```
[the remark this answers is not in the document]
Hesara2003: Good catch, that was a real regression. Removed the security_admin gating entirely: the route guard, the three vulnerability endpoints and the SideBar check. Before this PR none of them existed and access came from the project feature flags, so that is restored.

Also removed the module itself rather than leaving it empty. Who should hold Security Admin needs a product decision, and the user story describing a Super Admin persona needs reconciling since no data source emits that role.

apps/customer-portal/backend-v2/cmd/server/main_rbac_test.go
```diff
@@ -0,0 +1,287 @@
...
+		},
+		{
+			name:         "PartnerAdmin can manage contacts",
+			roles:        []string{"sn_customerservice.partner_admin"},
+			wantHTTPCode: http.StatusOK,
+		},
+		{
+			// A role outside the allow-list no longer slips through: the
+			// blanket super_admin grant in RequireRoles is gone.
+			name:         "CustomerUser cannot manage contacts",
```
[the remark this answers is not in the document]
Hesara2003: Already removed in the previous commit. PartnerUser cannot manage contacts also existed, so both scenarios were covered and the case was redundant rather than misnamed.

apps/customer-portal/backend-v2/cmd/server/main.go
```diff
@@ -200,66 +201,66 @@ func main() {
 
 	mux.HandleFunc("POST /projects/search", projectHandler.SearchProjects)
 	mux.HandleFunc("GET /projects/{id}", projectHandler.GetProject)
-	mux.HandleFunc("PATCH /projects/{id}", projectHandler.PatchProject)
+	mux.Handle("PATCH /projects/{id}", middleware.RequirePermission(roleResolver, middleware.ModuleProjects, middleware.ActionUpdate)(http.HandlerFunc(projectHandler.PatchProject)))
```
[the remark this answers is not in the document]
Hesara2003: Fixed by making the UI follow the same permission. canEdit now comes from can("projects", "update") instead of the customer_admin role, so the page and the endpoint cannot disagree.

Worth flagging: this means a customer admin no longer gets the AI Assistant toggle, since the spec puts Projects/Update on the WSO2 side only. If they are meant to keep it, the fix is to grant customer_admin projects:update in the matrix rather than special case this page.

Hesara2003:
@coderabbitai resume

Rashmika998 (approved):

PR #1674, opened 2026-09-07
[Entity-Service] Add a native watch list for Postgres-backed cases

SasmithaDilshan:
## Purpose

The watch list existed **only on the ServiceNow path**. On the Postgres data source `watchList` was rejected outright, and no table stored watchers. ServiceNow is being decommissioned, so the watch list has to become native.

This is the storage + API layer. It is deliberately self-contained and does not depend on native event publishing (which does not exist yet — see *Follow-ups*).

## Changes

**Migration `000017_create_case_watchers`**

```sql
case_watchers (case_id UUID FK cases ON DELETE CASCADE,
               user_id TEXT NULL FK users,
               email   TEXT NOT NULL,
               added_at TIMESTAMPTZ,
               PRIMARY KEY (case_id, email))
```

ServiceNow modelled the watch list as **four separate glide_lists** collapsed into one by Ballerina, and those upstream lists could name groups as well as users — which is why `domain.WatchListUser` already carries a nullable user reference. The table preserves that looseness on purpose: `user_id` is a real FK when the watcher is a known platform user, `NULL` otherwise. `email` is therefore the always-present column, and the one the notification path actually consumes.

Keyed on `(case_id, email)`, normalised to lower case, so watching a case is idempotent even for an address with no user record. Partial index on `user_id` serves the other direction (one user's watched cases) — `ON DELETE CASCADE` already covers removal with the case.

**Service (`case_service.go`)**
- `watchList` removed from the ServiceNow-only rejection list and given its own branch, mutually exclusive with every other field.
- A non-nil pointer to an **empty slice clears** the list rather than reading as absent.

**Repository (`case_watcher_repo.go`)**
- `ListCaseWatchers` — LEFT JOIN to users, so non-user entries survive the read.
- `ReplaceCaseWatchers` — transactional wholesale replace. Delete + insert in one transaction, so a concurrent read sees one complete list or the other, never a half-cleared one.

**Native read** — `GetCaseByID` populates the watch list with a second query rather than a join, which would multiply the case row by its watcher count.

## Two decisions worth reviewing

1. **Unresolvable ids reject the whole request** rather than applying the resolvable subset. A watcher silently dropped from the list is a person who stops being notified, and nobody observes the absence — a rejected update is the better failure.
2. **The user reference id is emitted here, unlike on the ServiceNow path.** SN nulls it because a collapsed glide_list entry might not be a user at all; a non-NULL `user_id` here is a foreign key, so the id is known to be a user's. The reason for the null no longer applies.

## Testing

- `TestCaseService_UpdateCase_SetWatchList` — replace, clear-via-empty-list, mutual exclusion with state, non-UUID rejection, unresolvable-watcher error.
- `TestCaseService_UpdateCase_WatchListNoLongerServiceNowOnly` — regression guard for the field's removal from the ServiceNow-only list.

`go build`, `go vet` and `go test ./internal/...` are clean, with one exception: `TestSNCaseService_CreateCase_PublishesCaseCreated` fails, and it **fails identically on `dev-app-csm-portal`** — verified in a clean worktree of the base branch. Unrelated to this change.

## Follow-ups (not in this PR)

- **The native case service publishes no domain events at all.** All 8 publish call sites live in `sn_case_service.go` / `sn_incident_service.go`. Until `case_service.go` publishes `case.created` / `status_changed` / `assigned` / `comment_added`, native watchers are stored and editable but nothing notifies them — and `csm-flow-service` has no native stream to consume.
- The ServiceNow `Watch List` flow (auto-add a commenter to the list) can only be ported once the above lands.

### DIFF
--- a/entity-service/internal/repository/case_repo.go
+++ b/entity-service/internal/repository/case_repo.go
@@ -56,6 +56,19 @@ type CaseRepository interface {
 	// closed_at is set to NOW() when transitioning to closed.
 	// Returns a NotFoundError if no matching row exists.
 	UpdateCase(ctx context.Context, req domain.UpdateCaseRequest) (domain.Case, error)
+	// ListCaseWatchers returns the case's watch list, in the order watchers were
+	// added. A watcher need not be a platform user, so an entry's user reference
+	// may carry a null id -- see migration 000017 and domain.WatchListUser.
+	// Returns an empty (non-nil) slice for a case with no watchers, and for a
+	// case that does not exist: the read is a projection, not an existence check.
+	ListCaseWatchers(ctx context.Context, caseID string) ([]domain.WatchListUser, error)
+	// ReplaceCaseWatchers replaces the case's watch list wholesale with the users
+	// named by userIDs, transactionally, and returns the resulting list. An empty
+	// userIDs clears the list. Returns a NotFoundError if the case does not
+	// exist, and a ValidationError if any id names no user or a user with no
+	// email address -- the request is rejected whole rather than partially
+	// applied, so a caller is never silently left with fewer watchers than asked.
+	ReplaceCaseWatchers(ctx context.Context, caseID string, userIDs []string) ([]domain.WatchListUser, error)
 	// CreateCaseAttachment inserts a new attachment metadata row for the case
 	// identified by req.ReferenceID. req.StorageKey must be non-nil: this data
 	// source stores file bytes externally in SFTPGo, never inline in Postgres.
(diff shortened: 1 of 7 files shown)


SasmithaDilshan:
Closing: not needed.

These tables already exist in `wso2-enterprise/digiops-cs` → `operations/csm-sync-service`, which is where existing ServiceNow tables get migrated to Postgres:

- `time_card` + `time_card_approver` — migration `0041`, mapping `time_card.yaml`
- `work_item_watcher` — migration `0042`, mapping `task_watch_list.yaml`

The division is by provenance: **new** Postgres schemas belong in entity-service; **existing SN tables** being migrated belong in csm-sync-service. Both of these are the latter, so building them here duplicated work already done.

No code from this PR is needed elsewhere — the sync-service equivalents are field-for-field complete.

PR #1675, opened 2026-09-07
[Entity-Service] Native (Postgres) time-cards entity

SasmithaDilshan:
## Summary

Adds a native time-cards implementation so time cards no longer require the ServiceNow passthrough:

- **migration 000016**: `time_cards` + `time_card_approvers`, state enum, indexes, checks
- **`TimeCardRepository`** (pgx): create/get/search/update/transition/delete
- **native `TimeCardService`**: validation, submitted-on-create, approve/reject state machine, submitter/approver authorization; `SearchCaseTimeCards` stays ServiceNow-only
- **routes**: use the native service under the Postgres data source
- **openapi**: drop `(ServiceNow data source only)` from the four endpoints this makes native
- table-driven service tests

Prerequisite for the billable/non-billable case-type rule (see Follow-ups).

## No new API — a second backend for the existing one

The routes, handler and OpenAPI contract already existed; they were served by `sn_time_card_service.go`, which forwarded every call to ServiceNow. This adds a second implementation behind the same `TimeCardService` interface, selected once at startup by `cfg.DataSource`:

```
POST /time-cards → TimeCardHandler → TimeCardService (interface)
                                       ├── ServiceNowTimeCardService → HTTP → ServiceNow
                                       └── timeCardService (NEW)     → SQL  → Postgres
```

Callers send identical requests either way. Neither the repository nor the service imports anything ServiceNow-related — no outbound HTTP, no integration-service client. `x-user-id-token` is the platform's own caller-identity header (decoded locally, user resolved from the Postgres `users` table), not a ServiceNow dependency.

## Spec changes

Four endpoint summaries claimed `(ServiceNow data source only)` and were made false by this change — corrected:

| Endpoint | Now |
|---|---|
| `POST /time-cards` | native |
| `POST /time-cards/search` | native |
| `PATCH /time-cards/{id}` | native |
| `DELETE /time-cards/{id}` | native |
| `POST /cases/time-cards/search` | **still ServiceNow-only** — summary unchanged, and it already documents the `503` |

## Schema note

`users.id` and `projects.id` are `TEXT` (migrations 000001 / 000003), so the user and project references here are `TEXT`; `case_id` is `UUID` because `cases.id` is.

This matches the convention in `000009_create_case_comments` and `000012_create_case_attachments`, which also reference `users(id)` as `TEXT`. It deliberately does **not** match `000006` / `000008`, which declare `UUID` columns against those `TEXT` primary keys — a pre-existing inconsistency, out of scope here.

## Testing

`go build`, `go vet` and `go test ./internal/service/ -run TestTimeCard` are clean — 4 test functions, 14 subtests covering create validation, the approve/reject state machine, submitter/approver authorization, and delete authorization.

Full `go test ./internal/...` is clean with one exception: `TestSNCaseService_CreateCase_PublishesCaseCreated` fails, and it **fails identically on `dev-app-csm-portal`** — verified in a clean worktree of the base branch. Unrelated to this change.

The migration was **not** applied against a live database — no Postgres or Docker in this environment. Verified statically only (FK type compatibility against the referenced primary keys; `down` drops in reverse dependency order). Worth a real up/down run before merge.

## Follow-ups

1. **Billable/non-billable case-type cascade.** Needs native `u_case_type` first, then a cascade in `case_service.UpdateCase`: billable ← Query; non-billable ← Incident, Security Report Analysis, Service Request, Engagement.
2. ~~Confirm a `DELETE /time-cards/{id}` route exists.~~ **Confirmed — already exists** at `internal/server/routes.go:476`, now resolving to the native service.
3. **The `DECISION`-marked authorization rules need checking against the ServiceNow definition** — particularly who may edit a card after submission and who may approve.
4. **Check whether anything actually calls `POST /cases/time-cards/search`** (portal, microapp). It returns 503 on Postgres, so a live caller would break at cutover.

### DIFF
--- a/entity-service/internal/repository/time_card_repo.go
+++ b/entity-service/internal/repository/time_card_repo.go
@@ -1,0 +1,463 @@
+// Copyright (c) 2026 WSO2 LLC. (https://www.wso2.com).
+//
+// WSO2 LLC. licenses this file to you under the Apache License,
+// Version 2.0 (the "License"); you may not use this file except
+// in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package repository
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/jackc/pgx/v5"
+	"github.com/jackc/pgx/v5/pgconn"
+	"github.com/jackc/pgx/v5/pgxpool"
+	"github.com/wso2-open-operations/cs-tools/entity-service/internal/apierror"
+	"github.com/wso2-open-operations/cs-tools/entity-service/internal/domain"
+)
+
+// TimeCardRepository is the native (Postgres) persistence for time cards.
+// SearchCaseTimeCards (the by-case rollup) is intentionally absent: it is a
+// ServiceNow-only capability, so the service layer rejects it before reaching
+// the repository.
+type TimeCardRepository interface {
+	// CreateTimeCard inserts a time card (in the submitted state) and its
+	// eligible-approver rows atomically, returning the enriched view.
+	CreateTimeCard(ctx context.Context, req domain.CreateTimeCardRequest, submitter string) (domain.TimeCardView, error)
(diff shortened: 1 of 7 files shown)


SasmithaDilshan:
Closing: not needed.

These tables already exist in `wso2-enterprise/digiops-cs` → `operations/csm-sync-service`, which is where existing ServiceNow tables get migrated to Postgres:

- `time_card` + `time_card_approver` — migration `0041`, mapping `time_card.yaml`
- `work_item_watcher` — migration `0042`, mapping `task_watch_list.yaml`

The division is by provenance: **new** Postgres schemas belong in entity-service; **existing SN tables** being migrated belong in csm-sync-service. Both of these are the latter, so building them here duplicated work already done.

No code from this PR is needed elsewhere — the sync-service equivalents are field-for-field complete.

PR #1686, opened 2026-09-08
Add case PATCH and comment POST proxy endpoints to csm-integration-service

Rashmika998:
## Summary
- Adds `PATCH /cases/{id}` and `POST /cases/{id}/comments` to csm-integration-service, proxying to the equivalent entity-service endpoints (`internal/entity/entity.go`, new `internal/handler/cases.go`), following the same pattern as the existing `PATCH /projects/{id}` passthrough.
- `PATCH /cases/{id}` succeeds for a state/severity/workState-only update when entity-service runs on a Postgres data source; every other field it accepts is ServiceNow-only and requires a forwarded end-user identity this M2M-only service can't supply, so those calls get a mapped 401 — same as the existing `PATCH /projects/{id}`.
- `POST /cases/{id}/comments` always 401s on both data sources (entity-service resolves the comment author from the forwarded token even on its Postgres path) — kept for API-shape completeness, documented accordingly.
- Updates `openapi.yaml`, `README.md`, and `CLAUDE.md` to reflect the new endpoints and their auth caveats.

## Test plan
- [x] `go vet ./...`
- [x] `go test -race ./...`
- [x] `gosec -fmt=text ./...` — 0 issues (12 files scanned)
- [x] `govulncheck ./...` — only pre-existing Go-stdlib CVEs tied to the pinned toolchain version in `go.mod` (unrelated to this change; `go.mod` untouched)
- [x] New handler tests for both endpoints (`internal/handler/cases_test.go`) covering path validation, body-size/JSON validation, verbatim passthrough, and the shared `upstreamErrors` mapping table

## Summary by CodeRabbit

- **New Features**
  - Added case updates through `PATCH /cases/{id}` for state, severity, and work-state changes when backed by Postgres.
  - Added case comment submission endpoint: `POST /cases/{id}/comments`.
  - Added validation for case IDs and request bodies, with clear responses for invalid or oversized requests.

- **Documentation**
  - Updated service documentation and API specifications with case operations, request formats, responses, and current authorization limitations.

### DIFF
--- a/integrations/csm-integration-service/internal/entity/entity.go
+++ b/integrations/csm-integration-service/internal/entity/entity.go
@@ -74,16 +74,25 @@ func (c *Client) UpdateProject(ctx context.Context, id string, body []byte) ([]b
 	return c.do(ctx, http.MethodPatch, fmt.Sprintf("/projects/%s", url.PathEscape(id)), body)
 }
 
-// PatchCase calls PATCH /cases/{id} on the entity service to update a case's
-// state, severity, or workState. Unlike UpdateProject, this entity-service
-// operation does NOT unconditionally require a forwarded end-user identity
-// token: on a Postgres data source, a state/severity/workState-only update
-// (the only fields this operation accepts there) succeeds for a pure M2M
-// caller. Every other field this request shape can carry (watchList,
-// assigneeEmail, subject, etc.) is ServiceNow-data-source-only and does
-// require a forwarded token this service cannot supply, so those calls are
-// expected to 401 the same way UpdateProject always does. Response is
-// returned as raw JSON; typed response structs are deferred.
+// PatchCase calls PATCH /cases/{id} on the entity service to update a case.
+// Unlike UpdateProject, this entity-service operation does NOT unconditionally
+// require a forwarded end-user identity token — whether it does depends on
+// entity-service's own DATA_SOURCE:
+//   - On a Postgres data source, a state/severity/workState update (optionally
+//     combined with resolutionCode/cause/closeNotes when state is closed or
+//     solution_proposed) succeeds for a pure M2M caller — no token check on
+//     this path. Every other field this request shape can carry (watchList,
+//     assigneeEmail, type and its companions, parentId, relatedCaseId,
+//     autocloseHoldUntil, subject, description, deploymentId,
+//     deployedProductId, the fix-ETA group, acknowledge, workaroundProvided)
+//     is ServiceNow-only and is rejected with 400 on this data source, not
+//     proxied through to any token check.
+//   - On a ServiceNow data source, every field this operation accepts —
+//     including a bare state/severity/workState update — requires a forwarded
+//     end-user identity token, so every call here is expected to 401 the same
+//     way UpdateProject always does, with no field combination that succeeds.
+//
+// Response is returned as raw JSON; typed response structs are deferred.
 func (c *Client) PatchCase(ctx context.Context, id string, body []byte) ([]byte, error) {
 	return c.do(ctx, http.MethodPatch, fmt.Sprintf("/cases/%s", url.PathEscape(id)), body)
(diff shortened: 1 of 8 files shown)


shayanmalinda:
🤖 **Claude review complete** — 1 comment(s) posted ([view review](https://github.com/wso2-open-operations/cs-tools/pull/1686#pullrequestreview-5140356252))

🟡 Minor: 1

<sub>⚠️ AI-generated review. Suggestions may be incorrect or incomplete, please verify before applying.</sub>

integrations/csm-integration-service/internal/entity/entity.go
```diff
@@ -74,6 +74,31 @@ func (c *Client) UpdateProject(ctx context.Context, id string, body []byte) ([]b
 	return c.do(ctx, http.MethodPatch, fmt.Sprintf("/projects/%s", url.PathEscape(id)), body)
 }
 
+// PatchCase calls PATCH /cases/{id} on the entity service to update a case's
+// state, severity, or workState. Unlike UpdateProject, this entity-service
+// operation does NOT unconditionally require a forwarded end-user identity
+// token: on a Postgres data source, a state/severity/workState-only update
+// (the only fields this operation accepts there) succeeds for a pure M2M
+// caller. Every other field this request shape can carry (watchList,
+// assigneeEmail, subject, etc.) is ServiceNow-data-source-only and does
```
shayanmalinda: **🟡 Minor: `subject` is not a field this operation accepts**

The entity service's `UpdateCaseRequest` (`entity-service/openapi.yaml:7001`) accepts `state`, `severity`, `workState`, `watchList`, `assigneeEmail`, `acknowledge`, `type` and the combinable ETA group, with no `subject` anywhere, so a consumer following this hint gets a 400 rather than the documented 401. The same wording is repeated in `openapi.yaml:355` and `CLAUDE.md:65`. Upstream also marks only `watchList`, `assigneeEmail`, `type` and that ETA group as ServiceNow-only, so `acknowledge` (and the `resolutionCode`/`cause`/`closeNotes` companions to `state`) should work on Postgres too, making the "only fields this operation accepts there" note on line 81 narrower than reality.

---
<sub>⚠️ AI-generated review. Suggestions may be incorrect or incomplete, please verify before applying.</sub>

shayanmalinda (commented):
Read the two handlers against `UpdateProject`, and checked the documented upstream behaviour against `entity-service`'s own routes and OpenAPI spec in this repo. The passthrough itself matches the existing pattern; one comment inline about the field list in the docs.

integrations/csm-integration-service/internal/entity/entity.go
Rashmika998: Good catch — thanks. `subject` is a real, functional field though: it's wired into `sn_case_service.go`'s ServiceNow path (used to update the case title), it's just missing from entity-service's own `openapi.yaml` documentation, which is a separate pre-existing gap in that repo, not something this PR introduced.

The more useful fix was the second point: the doc conflated two different entity-service deployment states. Corrected it to be explicit —

- On `DATA_SOURCE=postgres`: state/severity/workState (optionally with resolutionCode/cause/closeNotes) succeeds; every other field (watchList, assigneeEmail, type+companions, parentId, relatedCaseId, autocloseHoldUntil, subject, description, deploymentId, deployedProductId, the fix-ETA group, acknowledge, workaroundProvided) gets a **400** there, not 401 — entity-service's Postgres path rejects them explicitly before ever reaching a token check.
- On `DATA_SOURCE=servicenow`: every field, including a bare state/severity/workState update, requires the forwarded identity token and 401s, with no field combination that succeeds.

Updated `internal/entity/entity.go`, `openapi.yaml`, `CLAUDE.md`, and `README.md` accordingly in the latest commit.

shayanmalinda (approved):

PR #1709, opened 2026-09-10
[Customer Portal v2] Accept both id shapes on every path parameter

Hesara2003:
## Purpose
Every id path parameter on `backend-v2` required the dashed UUID form and rejected the dashless one outright, so a URL carrying a bare 32-hex ServiceNow sysid answered `{"message":"Invalid UUID format."}` — reported from staging against `GET /projects/{id}`. The 400 is written by this backend's own handler guard, before entity-service is ever called; entity-service itself would have served the request fine (its ServiceNow path runs `uuidToSysid`, which passes a non-canonical value straight through).

> **Scope note:** path (and the WebSocket `sessionId` query) parameters only — the ids that arrive from a URL a person can bookmark, share, or copy out of ServiceNow. Request-body id fields (`CommentCreateRequest.referenceId`, `CreateCaseRequest.conversationId`, the WebSocket message's own `conversationId`) stay strict-UUID: the app builds those from an entity-service response, which is always dashed, so there is no dashless source to accommodate. `openapi.yaml` keeps `format: uuid` on exactly those, so the spec and the code agree on both sides.

## Goals
- A detail-route URL carrying either id shape resolves instead of 400ing, across all 97 paths — not just the one reported.
- One shared helper for it, rather than the per-endpoint normalization added recently for call requests and deployed products.
- Keep refusing anything that is neither shape, so a malformed value can still never reach an upstream URL path.

## Approach
- `internal/handler/response.go` — two helpers: `isEntityID` (dashed UUID **or** bare 32-hex sysid) and `toDashedID` (inserts the hyphens for a sysid; no-op otherwise). `isAttachmentID` now delegates to `isEntityID` — it was already the identical predicate.
- Every id path parameter is read as `toDashedID(r.PathValue("…"))` and validated with `isEntityID` instead of `uuidRe`, across 18 handler files. The `sessionId` query parameter on `GET /ws` gets the same treatment, since it also carries a project id from a browser-built URL.
- **Normalization is to the dashed form, deliberately, because it is the only shape both entity-service data sources accept:** the ServiceNow path strips the hyphens itself, but the Postgres path runs `validateUUIDs` and rejects a dashless id. Normalizing to a sysid instead — as the ServiceNow-only call-request and deployed-product routes do — would work on a ServiceNow deployment and 400 on a Postgres one. Noted in `CLAUDE.md` so the next endpoint follows the same direction.
- `openapi.yaml` — the 93 id path parameters now declare a two-shape `pattern` in place of `format: uuid`.

## User stories
As a customer-portal user, I can open a project/case/deployment link that carries the dashless form of an id and land on the record, instead of getting a validation error.

## Release note
- [Customer Portal] Accept both the dashed-UUID and ServiceNow sysid form of an id in backend-v2 URLs.

## Documentation
N/A — internal API behaviour; the convention is recorded in `apps/customer-portal/backend-v2/CLAUDE.md`.

## Automation tests
- Unit tests (`internal/handler/entity_id_test.go`): `toDashedID` across both shapes, casing, and every near-miss (31/33 hex, non-hex, path traversal); `isEntityID` accept/refuse sets; and a route-level regression test through a real `http.ServeMux` on `GET /projects/{id}` asserting a dashless id now returns 200 **and** that entity-service receives the dashed form — plus that a malformed id still 400s without the upstream ever being called.
- `gofmt -l` clean, `go vet ./...` clean, `make test` (vet + `go test -race ./...`) green, `gosec -fmt=text ./...` 0 issues across 110 files.

## Security checks
 - Ran FindSecurityBugs plugin and verified report? N/A (Go backend change)

## Test environment
- macOS, Go 1.26, `make test` + `gosec` locally. Needs a staging deploy to confirm the originally reported URL against the live ServiceNow data source.

### DIFF
--- a/apps/customer-portal/backend-v2/CLAUDE.md
+++ b/apps/customer-portal/backend-v2/CLAUDE.md
@@ -703,8 +703,29 @@ struct actually carries it), and a stray extra check on `PATCH` would just be de
 - **Auth**: always check `middleware.UserInfoFromContext(r.Context()) == nil` first → 401.
 - **Body size**: use the shared `readJSONBody(w, r)` helper (`internal/handler/response.go`) — caps
   at `maxRequestBodyBytes` (1 MiB) and validates the body is well-formed JSON.
-- **Path params**: guard against empty string after `r.PathValue("id")`; validate UUID-shaped IDs
-  with the package-level `uuidRe` and return 400 on mismatch before calling entity-service.
+- **Path params**: normalize an id-shaped path param with `toDashedID(r.PathValue("id"))`, then
+  validate it with `isEntityID` and return 400 on mismatch before calling entity-service. Do **not**
+  validate with `uuidRe` directly — an id reaches this API in two shapes (the dashed UUID
+  entity-service returns, and the bare 32-hex ServiceNow sysid the same value has upstream), and
+  `uuidRe` alone rejected every dashless one with `Invalid UUID format.` before it ever left this
+  backend. That was reported from staging as `GET /projects/<32-hex sysid>` → 400, from a URL
+  carrying the dashless form; the customer-portal webapp, unlike the CSM portal's
+  `useNormalizedIdParam`, does not repair such a URL itself.
+  - **Normalize to the dashed form, never to a sysid.** It is the only shape both entity-service
+    data sources accept: the ServiceNow path strips the hyphens itself (`uuidToSysid`, which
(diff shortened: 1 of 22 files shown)


Hesara2003:
Closing in favor of the frontend normalization fix in #1710, following the same architectural pattern used in the CSM portal (#1509) where the webapp normalizes URLs and routes before dispatching backend requests.

PR #1710, opened 2026-09-10
[Customer Portal] Repair dashless ids in portal URLs

Hesara2003:
## Purpose
A customer-portal URL carrying the dashless form of an id (the bare 32-hex ServiceNow sysid, e.g. from a bookmarked or shared link) reached the backend as-is, which answered `Invalid UUID format.`. The CSM portal has repaired this in its own routes since #1509 (`useNormalizedIdParam`); the customer-portal webapp had no equivalent, so the id went out exactly as it arrived.

> **Scope note:** frontend only, and a companion to — not a replacement for — the backend accepting both shapes (that half is #1709). This fixes ids arriving through this app's own routes; a direct API call still needs the backend fix.

## Goals
- A project or detail-page URL carrying a dashless id loads the record instead of erroring.
- The address bar self-corrects to the dashed form, so what gets copied out of it is the canonical id.
- The data fetch uses the corrected id on the **first** render, not only after the redirect lands.

## Approach
- `src/hooks/useNormalizedIdParam.ts` — ported from `apps/csm-portal/webapp`'s hook of the same name: returns the param in dashed form immediately, and `replace`-navigates to the dashed URL as a side effect (preserving search, hash and router state). Uses `useNavigate` directly, since this app has no `useNavTransition`.
- **One deliberate difference from the CSM version:** the redirect rewrites *every* dashless-id segment of the pathname, not just the value of the param it was asked about. These routes nest two ids (`/projects/:projectId/support/cases/:caseId`), so two instances of the hook are live on every detail page — a per-param rewrite would have each instance navigate to a pathname built from the pre-fix original, so the last one to run would undo the others. Whole-path means every instance computes the identical target and the duplicate navigations are no-ops.
- Wired into `ProjectGuard` (which covers the whole `projects/:projectId` subtree, including the project fetch the guard itself makes) and the pages that read their own id param: case details, service-request details, change-request details, conversation details, announcement details, vulnerability details, and the Novera chat page (whose `conversationId` is optional on the index route).

## User stories
As a customer-portal user, I can open a link that carries the dashless form of a project or case id and land on the record, with the URL corrected to the canonical form.

## Release note
- [Customer Portal] Repair dashless ids in portal URLs instead of failing the request.

## Documentation
N/A — internal routing behaviour.

## Automation tests
- Unit tests (`src/hooks/__tests__/useNormalizedIdParam.test.tsx`, 11 cases) against a real `MemoryRouter`/`Routes` rather than a mocked `useNavigate`, so the redirect itself is observable: first-render value, URL rewrite, already-dashed no-op, two nested ids in one pass, one-of-two dashless, query/hash preserved, absent optional param, and four malformed shapes left alone.
- `VulnerabilityDetailsPage.test.tsx` needed `useLocation` added to its existing wholesale `react-router` mock — that page reads its ids through the new hook, which inspects the current location. Caught by comparing the full suite against a baseline run on `main` rather than by eyeballing the failure count.
- **Suite compared against `main`, since this app's suite is not currently green on `main`:** `main` = 58 failed / 347 passed files (140 failed / 1217 passed tests); this branch = 58 failed / **348** passed files (140 failed / **1228** passed tests). The failing set is identical file-for-file; the deltas are exactly this PR's own new tests. `tsc -b` and `vite build` clean; `eslint` reports no problems in any file this PR touches (the ~287 pre-existing errors elsewhere in `src` are untouched).

## Security checks

## Test environment
- `./node_modules/.bin/tsc -b`, `eslint`, `vitest run`, `vite build` locally.

## Summary by CodeRabbit

* **New Features**
  * Added automatic normalization of dashless UUIDs in route URLs to standard dashed format.
  * Preserved query strings and URL fragments during normalization.
  * Applied normalized identifiers across project, case, conversation, request, announcement, and vulnerability pages.

* **Bug Fixes**
  * Improved handling of malformed, optional, and already-normalized route identifiers without unnecessary navigation.

* **Tests**
  * Added coverage for identifier normalization, nested routes, URL preservation, and invalid values.


madubashinii:
@coderabbitai review

rksk (approved):

PR #1738, opened 2026-09-13
[CSM Portal] Fix CR clone-gap comment and add Implementation Plan visibility toggle

rksk:
Two related, small CR fixes bundled together:
1. A code comment in the CR clone gap analysis claimed `implementationPlan` was write-only (accepted on create, never returned by GET). That's no longer true — the entity-service's read path already returns it (a prior field-parity fix).
2. `isPlanningVisibleToCustomers` on a Change Request was read-only everywhere: CSM could see its value on the CR detail page, but had no way to set it on create or change it later.

1. Correct the stale comment to reflect current reality.
2. Give CSM engineers a real write path for `isPlanningVisibleToCustomers`: a toggle on the Change Request create page and the edit dialog, labeled "Implementation Plan visible to customers" for clarity.

**Comment fix:** two comment-only edits, no behavior change, in `changeRequests.ts` and its test.

**Implementation Plan visibility toggle** — additive, bypass-role-only change (CSM/internal callers; the customer-portal-facing path is untouched and cannot set this field, by design):

- `entity-service`: added `IsPlanningVisibleToCustomers *bool` to `CreateChangeRequestRequest` and `PatchChangeRequestRequest`, forwarded through the outgoing payload to the downstream API, added to both request schemas in `openapi.yaml`. A `*bool` (not a bare `bool`) matches the sibling boolean fields' existing convention and correctly distinguishes "not provided" from an explicit `false`.
- `apps/csm-portal/webapp`: added the field to the create/patch request types; added a `Switch` toggle to `CreateChangeRequestPage` and to `EditChangeRequestDialog` (wired into the dialog's existing "only send changed fields" diff logic); relabeled the existing read-only display from "Planning visible to customers" to "Implementation Plan visible to customers" for consistency; updated the in-app help guide topic that previously said this field had no edit control.

A corresponding change on the downstream data-source side (adding the write path there) is tracked separately.

As a CS engineer, I can mark whether a change request's implementation plan is visible to the customer, both when creating the CR and later when editing it.

Change Requests: "Implementation Plan visible to customers" can now be set from the CSM portal's create and edit forms, not just viewed.

Updated the in-app `/help` guide topic for Operations (Change Requests section) to reflect the new edit control.

## Training
N/A — no training content affected.

## Certification
N/A — no certification content affected.

## Marketing
N/A — internal workflow improvement, not a marketed feature.

## Automation tests
 - Unit tests
[...]

PR #1740, opened 2026-09-14
[CSM Portal] Add write control for Implementation Plan visible to customers on change requests

rksk:
`isPlanningVisibleToCustomers` on a Change Request was read-only everywhere: CSM could see its value on the CR detail page, but had no way to set it on create or change it later — the only way to change it was a direct ServiceNow edit, which isn't a real workflow for CS engineers.

Give CSM engineers a real write path for this field: a toggle on the Change Request create page and the edit dialog, labeled "Implementation Plan visible to customers" for clarity.

Additive, bypass-role-only change (CSM/internal callers; the customer-portal-facing path is untouched and cannot set this field, by design):

- `entity-service`: added `IsPlanningVisibleToCustomers *bool` to `CreateChangeRequestRequest` and `PatchChangeRequestRequest`, forwarded through the outgoing payload to the downstream API, added to both request schemas in `openapi.yaml`. A `*bool` (not a bare `bool`) matches the sibling boolean fields' existing convention and correctly distinguishes "not provided" from an explicit `false`.
- `apps/csm-portal/webapp`: added the field to the create/patch request types; added a `Switch` toggle to `CreateChangeRequestPage` and to `EditChangeRequestDialog` (wired into the dialog's existing "only send changed fields" diff logic); relabeled the existing read-only display from "Planning visible to customers" to "Implementation Plan visible to customers" for consistency; updated the in-app help guide topic that previously said this field had no edit control.

A corresponding change on the downstream data-source side (adding the write path there) is tracked separately.

As a CS engineer, I can mark whether a change request's implementation plan is visible to the customer, both when creating the CR and later when editing it.

Change Requests: "Implementation Plan visible to customers" can now be set from the CSM portal's create and edit forms, not just viewed.

Updated the in-app `/help` guide topic for Operations (Change Requests section) to reflect the new edit control.

## Training
N/A — no training content affected.

## Certification
N/A — no certification content affected.

## Marketing
N/A — internal workflow improvement, not a marketed feature.

## Automation tests
 - Unit tests
   > entity-service: extended `sn_change_request_service_test.go` to cover explicit-`false`-not-dropped and explicit-`true`-forwarded for the new field. `go test ./...` passes in full.
 - Integration tests
   > webapp: existing `CreateChangeRequestPage`, `EditChangeRequestDialog`, and `CsmChangeRequestDetailPage` test suites pass unchanged (85/85). No new dedicated test was added for the toggle itself.

## Security checks
 - Ran FindSecurityBugs plugin and verified report? N/A — no new external dependency or security-sensitive code path

## Samples
N/A

## Related PRs
A corresponding entity-service-side change is tracked separately (not in this repo).

## Migrations (if applicable)
N/A

## Test environment
`go build`, `go vet`, `go test ./...` (entity-service); `pnpm build`, `pnpm lint`, `pnpm vitest run` (webapp) — all passing locally. No live/staging verification attempted in this PR.

## Learning
Found while scoping a related support request about Implementation Plan visibility on Change Requests; this PR only covers giving CSM a write control for the existing flag, not any customer-portal-facing enforcement of it (that was investigated and deliberately deferred — see the linked task notes — because the flag's current data distribution would make naive enforcement a large, unintended regression for customers).


rksk:
Superseded — folded into #1738, no need for a separate PR.

PR #1747, opened 2026-09-14
[Customer Portal v2] Fix product class filtering and deployed-product identifier handling

Hesara2003:
## Purpose
Two unrelated-looking failures in the customer portal turned out to be the same class of problem — this backend rewriting an identifier or a filter value into a shape the upstream does not accept.

1. **`GET /products?class=product_model` returned `"products": []`** against a page that was entirely product models, with `totalRecords: 41` alongside it.
2. **`POST /deployments/{deploymentId}/products` failed with `projectId contains invalid UUID`**, so a deployed product could not be created at all.

## Goals
- The product list loads when the frontend filters by class.
- Creating and updating a deployed product works, along with its metrics and usage-count searches.
- The identifier direction is consistent across this backend, so the same bug cannot be reintroduced by reaching for the wrong helper.

## Approach

**1. Product class filter** (`internal/dto/product.go`)

entity-service has no class filter, so the class is applied in memory after the fact. The two sides spell the same value differently: the frontend sends the portal's own enum `product_model` (matching the Ballerina backend's `entity:ProductClass`), while entity-service passes ServiceNow's display label `Product Model` straight through. The comparison used `strings.EqualFold`, which folds case but not the space against the underscore — so every item on the page was discarded.

Added `normalizeProductClass` (trim, lowercase, spaces to underscores) and compared through it. A genuinely different class still matches nothing, so this does not widen the filter.

**2. Deployed-product identifiers** (`internal/dto/deployed_product.go`, `internal/handler/deployed_products.go`)

These routes converted identifiers to a hyphen-stripped sysid before calling entity-service. entity-service does the opposite: it validates `projectId`/`deploymentId`/`productId`/`versionId`/`id` with `validateUUIDs` and performs the sysid conversion itself. A stripped value therefore failed validation before any work happened — on create, on patch, and on both the metrics and usage-count searches.

Switched all of them to the canonical dashed form via the existing `toDashedID` helper, which is the direction the call-request builders already use ("expected by entity-service") and the direction the deployed-product *search* builder already used. Removed the now-unused `dto.ToSysID` and `handler.toSysID` so the wrong direction is no longer reachable.

> **Scope note:** three tests asserted the old sysid output. They only ever ran against a fake, so they encoded the assumption rather than verifying it; their expectations are flipped here.

## User stories
As a customer-portal user, I can filter the product list by class and see results, and I can add a product to a deployment instead of hitting a validation error.

## Release note
- [Customer Portal] Fix the empty product list when filtering by class, and fix deployed-product create/update failing with an identifier validation error.

## Documentation
N/A — internal API behaviour.

## Automation tests
- `internal/dto/product_test.go` (new): the portal value against the ServiceNow label, both spellings either way, mixed case and spacing, surrounding whitespace, a different class, a substring near-miss, a nil class, and an absent filter. Also asserts the pagination envelope is left untouched.
- `internal/dto/deployed_product_test.go`: create and update builders must emit dashed UUIDs whichever shape arrives.
- `internal/handler/deployed_products_test.go`: create and patch, across dashed, bare sysid, mixed and uppercase inputs.
- `gofmt` clean on every file touched, `make test` (vet + `-race`) green, `gosec -fmt=text ./...` 0 issues across 110 files.
- Verified against a locally running backend and the real upstream:
  - `GET /products?class=product_model&offset=0&limit=10` → `products: 0` before, `products: 10` after; a different class value still returns `0`.
  - `POST /deployments/{deploymentId}/products/{productId}/metrics/search` → `200` after. This is a read-only route that exercises the same identifier path that was returning the validation error; create and patch were not exercised live, since those write to the upstream system of record.

## Security checks
 - Ran FindSecurityBugs plugin and verified report? N/A (Go backend change)

## Test environment
- macOS, Go 1.26, local backend against the shared upstream services.

## Summary by CodeRabbit

- **Bug Fixes**
  - Product class filters now correctly match values such as `product_model` with labels such as “Product Model.”
  - Product and deployment requests now use consistent canonical dashed UUIDs.
  - Identifier handling remains compatible with dashed, undashed, mixed-case, and uppercase UUID formats.
  - Updated product, deployment, usage, and metrics operations to forward identifiers consistently.


madubashinii:
@coderabbitai review

Rashmika998 (approved):

PR #1763, opened 2026-09-15
[CSM Flow Service] Port the CR approval and plan-start-date notification flows

SasmithaDilshan:
Introduces the flow engine component and its first real ported flow.

## The component

An event consumer: `cmd/consumer` joins the bus and, for each record, runs every `Flow` whose `Match` returns true. A flow reacts by patching a native entity or by publishing a request that `csm-notification-service` sends — **it never sends email/chat/twilio itself**.

| Package | |
|---|---|
| `internal/flows` | the `Flow` contract and `Registry` |
| `internal/eval`, `internal/spec` | pure condition evaluation |
| `internal/porting/snquery` | ServiceNow encoded query → `spec.Condition`; a migration aid, deleted after cutover |
| `internal/eventbus`, `events`, `entity`, `apierror`, `middleware` | copies of the canonical versions in csm-notification-service / entity-service, kept in sync by hand |

`docs/` is excluded — it holds the ServiceNow discovery scripts and analysis notes, not component code.

## The ported flow

The original triggers on `change_request` `record_update` with `stateCHANGESTO-4^OR-3^OR5^OR0^OR1` and branches five ways:

| state | audience | recipients | subject suffix |
|---|---|---|---|
| `-4` ASSESS | internal | **Devops Approval** members | Request for Approval - Implementation |
| `-3` AUTHORIZE | internal | **CAB Approval** members | Request for CAB approval - Authorize |
| `0` REVIEW | internal | **Devops Review** members | Request for approval - Review |
| `5` CUSTOMER_APPROVAL | customer | project contacts | Request for Approval - Implementation |
| `1` CUSTOMER_REVIEW | customer | project contacts | Request for approval - Customer Review |

Subjects are **verbatim** from ServiceNow's `sys_element_mapping` rows, including the internal branches' `[team]` segment. The team comes from the change request's git reference (Choreo / Asgardeo / MS) — an unrecognised or absent reference is `MS`, which is the original's `Else` branch, not a fallback invented here.

**"CHANGES TO", not "is".** A change request updated for some other reason while already sitting in an approval state must not re-notify. `entity.changed` carries before/after per field, which is what makes that expressible — and a test pins the `ASSESS → ASSESS` case as a non-match.

**One deliberate difference:** ServiceNow sent one email *per recipient*, because its subflows looped (`For Each`) with the address as the To. This publishes **one** event carrying the whole resolved recipient list. Nothing about the notice is per-person — same subject, same body — so the fan-out was an artefact of Flow Designer's loop, not intent. Addresses are de-duplicated, so one person in two approval groups is notified once.

Group **names**, not sys_ids: the original hardcoded three group sys_ids, which are meaningless outside that instance. The synced `team` table carries the name, and a name is also what a reader of the ported flow can recognise.

## ⚠️ Not registered, and a test enforces it

`cr_approval_notice` is **not** in `flows.All()`. `TestCRApprovalNotice_NotRegistered` fails if it is. Four things must land first:

1. **The double-fire guard** — ServiceNow still sends these. Registering is a paired change with disabling `CR Approval notifications` there, in the same commit.
2. **A consumer** for `change_request.approval_requested` in csm-notification-service. That's a change in that repo; this service is forbidden from sending email itself.
3. **`change_request.state` in Postgres** — added by wso2-enterprise/digiops-cs **PR #3058**.
4. **An `entity.changed` publisher for `change_request`** — nothing emits one today.

So this PR is the logic, complete and tested, with its activation deliberately gated.

## Testing

`go build`, `go vet`, `gofmt` clean; `go test ./...` green. 8 test functions on the new flow:

- `Match` across all five states plus the negatives — a non-approval state, no state change, already-in-state, a different entity, other event types, a malformed payload
- Both subject shapes, verbatim
- The git-reference → team chain including its `MS` fallback
- A guard that the branch table and the trigger condition stay in agreement — a sixth state added without updating the trigger, or one dropped, is a silent behaviour change

## Provenance

Reconstructed from the live instance via read-only background scripts: the flow and subflow definitions, `sys_choice` for the state labels, `sys_user_group` for the three approval groups, and `sys_element_mapping` for the email subjects — which is where Flow Designer actually stores them, not `sys_variable_value`, where every field reads empty.

## Summary by CodeRabbit

* **New Features**
  * Added the CSM Flow Service for processing domain events and database-triggered changes.
  * Added retry handling with optional dead-letter routing for failed records.
  * Added a health endpoint reporting service status and registered flows.
  * Added development tools for replaying, publishing, and dry-running events.
  * Added ServiceNow-style condition translation and validated flow specifications.
  * Added change-request approval notifications for customer and internal audiences.

* **Documentation**
  * Added setup, configuration, local execution, testing, event replay, and flow-porting guidance.


SasmithaDilshan:
The port read every field of its notice off the outbox snapshot, which cannot carry them — fixed here.

The trigger writes `to_jsonb(NEW)` of the table that changed, so a `change_request` row change carries exactly that table's own columns — `id`, `state`, `git_reference` — under their **database** names. The flow looked for `number`, `gitReference`, `projectId` and `requesterName` and would have found none of them.

The failure mode was not a crash:

- every branch would have published a subject reading `(...)` with no number
- every branch would have said team **MS**, since an absent git reference is the fallback the original's `Else` defined
- both customer branches resolve their audience from the project id, so that would have been `""` → `ProjectContactEmails` returns nothing → **`CUSTOMER_APPROVAL` and `CUSTOMER_REVIEW` publish nothing at all**

Two of five states silently notifying nobody, which looks like success in any test written against a hand-made fixture — and the fixtures carried the invented camelCase keys too, which is exactly how it passed. They now mirror what the trigger really writes, with a README saying so.

A flow reads the record through `flows.ChangeRequests` now, one join across `change_request`, `work_item`, `"user"` and `project`. That is also the more faithful port: the ServiceNow subflows read fields off the triggering record, not off a diff.

## Testing without a broker

`cmd/dryrun` runs a catalogued flow against the real database and prints the event it would publish — nothing published, no mail sent. Recipient resolution is real, which is the point: a flow that resolves nobody is the most common way a port looks fine and notifies no one.

```
DATABASE_URL=… go run ./cmd/dryrun -flow cr_approval_notice -out /tmp/notice.json
```

It needs only `DATABASE_URL` — no consumer group, no mail credential — and reports the outbox ids it claimed plus the SQL to replay them. Feed its output to csm-notification-service's `cmd/preview` (#1767) to see the email body itself.

`flows.Catalogue()` is what makes that reachable: a flow gated behind the double-fire guard was previously invisible outside its own package, so the one flow most in need of a dry run was the one flow that could not have one. `All()` is still the subset that actually runs, and a test keeps it a subset.

SasmithaDilshan:
Closing — superseded by #1811.

The two flows this PR ported have moved into entity-service, which already owns the `change_request` rows, the database the audiences are resolved from, and an event producer. Delivery stays in csm-notification-service (#1767). That leaves nothing for a separate flow service to do.

**Why close rather than strip the flows out.** `csm-flow-service` doesn't exist on `dev-app-csm-portal` — this PR introduces the whole service, +7,517 lines across 54 files, every line an addition. Removing the change-request flows wouldn't make it smaller in any useful sense; it would leave a PR that adds a ~4,600-line generic flow engine (`eval`, `spec`, `porting/snquery`, the registry and four commands) with **zero flows in it** and nothing using it.

**What carried across to #1811**, as a move rather than a rewrite — logic and tests intact:

- the five-state approval table and its subject wording
- the plan-start-date precedence rule (a date change is read *before* a cleared confirmation)
- the WSO2-actor gate on customer-proposed dates
- audience resolution over `team_member` / `project_contact`
- the outbox drainer

Plus one thing this PR had only referenced: `event_outbox`, its trigger and diff function had been applied to staging **by hand** and never committed. `CLAUDE.md` here points readers at `0045_event_outbox.sql`, a file that exists in neither repo. #1811 commits it as `000055_event_outbox`, reproduced from the live definition.

**If a future flow needs a generic engine**, this branch stays on the fork and the work is recoverable. The bet is that it won't: the change-request case suggests flow logic belongs with the data it reads, and the remaining CSM operations flows look the same. Better to re-introduce an engine when a flow genuinely spans services than to merge a framework with no users.

The `docs/` research — the migration tracker, flow classification and ServiceNow discovery output — was never committed here and is unaffected.

PR #1769, opened 2026-09-15
[Entity Service] Add Postgres-backed product-consumption provisioning state

Hesara2003:
## Purpose

First step of moving the product-consumption license-download path off ServiceNow and onto Postgres.

That path is six upstream calls, and only three of them are ServiceNow: read the project's provisioning status, record each step's result, and generate the license. The other three (create the Choreo application, subscribe it, mint its credentials) talk to the Choreo Dev Portal and are not a data source at all, so they are unaffected by a data-source change.

This PR implements the **first two** of those three — the provisioning state — as `GET` and `PATCH /projects/{id}/consumption`.

**Nothing changes behaviour yet.** The routes are new and nothing calls them; the Choreo subscription operation still reads and writes ServiceNow. Deliberately so — see "What this PR does not do".

## Goals

- A project's provisioning state (status 1–5, Choreo application ID, OAuth2 credentials, subscription secret keys) can live in Postgres.
- Advancing that state is safe under concurrency, because the failure mode is expensive: a lost race mints a **second Choreo application** for a customer who already has one.
- The credentials it stores are not readable from the database.

## Approach

**Postgres-only, by design.** This is the mirror image of the many ServiceNow-only routes already in `routes.go`. On the ServiceNow path this state lives on the `customer_project` record and is reached through the product-consumption scripted REST API, which the Choreo operation calls directly — neither this service nor the ServiceNow integration service is in that path, and neither has an endpoint for it. Adding a ServiceNow implementation here would have meant inventing a second route to data that already has one.

Three decisions worth review attention:

**1. The forward-only guard is in SQL, not read-then-write.**

```sql
ON CONFLICT (project_id) DO UPDATE SET ...
WHERE project_consumption.status < EXCLUDED.status
```

Two concurrent license downloads both read status 1 and both try to write 2. The loser's update is filtered out by that `WHERE` rather than overwriting the winner's application ID. The service turns the filtered-out case into a no-op that returns the stored state with an explanatory message — not an error — so a retrying caller converges on reality instead of failing and retrying the same step again.

A `COALESCE` of the supplied value against the stored one, on every other column, is what makes a partial write safe: a caller records only the artefacts of the step it just completed, so recording step 4's credentials cannot erase step 2's application ID. Note **where** that merge happens — in the `SELECT` that builds the proposed row, not in the `DO UPDATE SET`. See "Automation tests" for why; it is the bug the live-database run caught.

**2. Credentials are encrypted at rest, behind an interface.**

No secret has ever been stored in this schema, so this sets the precedent. `internal/crypto` is AES-256-GCM with a version byte on every value, reached through a `SecretCodec` interface — an external KMS can replace it later without a schema change or a backfill. The columns are `BYTEA` and hold ciphertext only.

Matching that, `ProjectConsumptionView` returns **no secret material**: the consumer secret and both subscription secret keys are reported only as the booleans `hasConsumerSecret`/`hasSecretKeys`. A status read happens on every license download, so returning them would put three long-lived credentials on the wire routinely for no reason.

**3. A missing key disables the feature rather than weakening it.**

`CONSUMPTION_SECRET_KEY` is optional, so this does not break existing Postgres deployments with no interest in product consumption. When it is absent the two routes are simply not registered — a missing key must never silently degrade into storing these values in the clear. Every path that leaves them unregistered says so at startup, because a disabled route and a mistyped URL are both a bare 404, and the deliberate choice is the half a person cannot discover from outside.

## What this PR does not do

**License generation is not migrated, and the Choreo operation is not repointed.** Those two facts are connected, and the second depends on the first.

License generation is not CRUD. It resolves the deployment's subscription key, runs the 12-position secret shuffle, persists it so a re-download reproduces the same file byte for byte, and HMAC-signs the payload. That logic exists only inside the ServiceNow instance as script includes (`LicenseFileContentSigner`, `ProductConsumptionUtils`) — it is vendored in no repository — and it is a live contract with deployed customer products, which unshuffle the secrets from their license file and key their `X-Hub-Signature-256` off the primary key. A reimplementation differing by one byte in canonicalisation breaks either existing license files or ingestion for every customer who re-downloads. It needs the script-include source and a differential test against ServiceNow, not a port from the design docs.

Until that exists, repointing the Choreo operation at these routes would split the state in half: provisioning would write status and secret keys to Postgres, while the license request still goes to ServiceNow, which would read its own `customer_project` record, find status below 5 and no secret keys, and fail to issue a license. So the operation stays on ServiceNow until either license generation moves too, or a deliberate bridge is designed for ServiceNow to reach the keys.

## Heads-up: the existing migration chain does not apply end to end

Not introduced here and not fixed here, but **verified against a real PostgreSQL 16 instance**, and worth knowing before anyone points an automated migration runner at staging. Applying `migrations/*.up.sql` in order to an empty database, **6 of 14 fail**:

| Migration | Result |
|---|---|
| `000001`–`000005` | apply |
| `000006_create_deployments` | **fails** — `foreign key constraint "deployments_project_id_fkey" cannot be implemented` |
| `000007_create_deployed_products` | fails — cascades from `000006` (`relation "deployments" does not exist`) |
| `000008_create_cases` | **fails** — `foreign key constraint "cases_created_by_fkey" cannot be implemented` |
| `000009`, `000012`, `000013_add_case_attachments_status` | fail — cascade from `000008` |
| `000010`, `000011`, `000013_create_scheduled_task_run`, `000014` | apply |

Two independent causes:

- **A type mismatch on two foreign keys.** `deployments.project_id` is `UUID` referencing `projects.id`, which `000003` declares `TEXT`; `cases.created_by` is `UUID` referencing `users.id`, also `TEXT`. PostgreSQL cannot create either.
- **Two migrations share version `000013`** — `000013_add_case_attachments_status` and `000013_create_scheduled_task_run`. A standard runner (golang-migrate and friends) rejects a duplicate version outright.

Both have stayed latent because `internal/db/migrate.go` is an empty stub, so nothing applies these at startup. This PR's table follows `000003` and uses `TEXT` for `project_id`; if the live schema turns out to use `UUID`, that one column changes with it. `000014` up **and** down both verified against a clean database.

## User stories

N/A — no user-facing change in this PR. The endpoints are new and unreferenced.

## Release note

- [Entity Service] Add Postgres-backed product-consumption provisioning state at `GET`/`PATCH /projects/{id}/consumption`, gated on `CONSUMPTION_SECRET_KEY`.

## Documentation

- `README.md` — new "Product-consumption provisioning state" section, including that license generation is explicitly still ServiceNow-side.
- `.env.example` — `CONSUMPTION_SECRET_KEY`, how to generate one, and that rotating it is a migration rather than a config tweak.
- `openapi.yaml` — both paths and three schemas.

## Automation tests

**A live PostgreSQL run caught a real bug in this PR**, which the unit tests could not: PostgreSQL evaluates CHECK constraints against the *proposed* insert tuple, before `ON CONFLICT` resolves anything. Merging the stored row in the `DO UPDATE SET` was therefore too late — recording a step that carries no artefacts of its own (status 3, subscribed) proposed a row full of NULLs and was rejected by `chk_project_consumption_application_id`, even though the update it would have performed preserved the stored value. The provisioning walk was broken at step 3 and no fake-repository test could have shown it. Fixed by moving the COALESCEs into the SELECT that builds the proposed row, so the constraints see the final row and `EXCLUDED` already carries the merged values.

- `internal/repository/project_consumption_repo_integration_test.go` (new, live database, skipped unless `ENTITY_TEST_DATABASE_URL` is set): the full 1→5 provisioning walk with each step supplying only its own artefacts; a partial write preserving earlier steps' output; a backwards write and a repeated same-status write both rejected without altering stored state; 8 goroutines racing the first write, asserting exactly one winner and 7 stale; the CHECK constraint rejecting an incomplete step even when the service layer is bypassed; unknown project not found on both read and write; and that `consumer_secret` read straight out of the column is neither empty nor the plaintext, while still round-tripping through the repository.
- `internal/crypto/secret_test.go` — round trip across empty/long/Unicode/embedded-NUL input; ciphertext does not contain the plaintext; two encryptions of the same value differ; tamper detection across a flipped body byte, a flipped nonce byte, a relabelled version byte, truncation and an empty value; a different key fails rather than returning damaged plaintext; and neither the decrypt error nor the key-parsing error leaks its input, since both reach logs.
- `internal/service/project_consumption_service_test.go` — the view never carries a secret (asserted against the **serialised** form, so a future field cannot leak one without failing); unprovisioned project reports pending rather than 404; each step's artefacts required, with an empty string treated as missing; step 3 requires none; out-of-range status rejected; a stale status is a no-op returning stored state; unsupplied fields reach the repository as nil.
- Full suite green both with and without `ENTITY_TEST_DATABASE_URL` — with it unset the live-database tests skip, so CI without a database stays green.
- The existing routes-contract test caught the missing `openapi.yaml` entry before I did.
- `go vet` clean, `gofmt` clean, `govulncheck` no vulnerabilities.

## Security checks

 - Ran FindSecurityBugs plugin and verified report? N/A (Go service)
 - `gosec -fmt=text ./...` — 0 issues across 132 files (file count verified non-zero, per the toolchain caveat in `CLAUDE.md`)

## Test environment

- macOS, Go 1.26, PostgreSQL 16.15.
- The live-database tests need only four migrations — `000001`, `000002`, `000003`, `000014` — since the full chain does not apply (see above). The exact steps are in the comment at the top of the integration test file.


Hesara2003:
Closing: this work is moving to `dev-app-csm-portal`, where product-consumption tracking is being built for both portals rather than the customer portal alone.

The branch and commits stay available — the change is re-targeted, not abandoned. Two things it needs on the way over:

- the migration renumbers from `000014` to `000055` (that branch's chain is at `000054`, which already mirrors the ServiceNow usage tables);
- the route gate drops its `DATA_SOURCE=postgres` condition. `project_consumption` has no ServiceNow equivalent, so it belongs with `sla_clocks`/`scheduled_task_run` as always-Postgres, registered whenever a pool exists. As written it would never register in a `DATA_SOURCE=servicenow` deployment, which is what staging and production run.

PR #1786, opened 2026-09-16
fix(customer-portal): add the missing PATCH /projects/{id} route

SasmithaDilshan:
Enabling the AI chat assistant returns **404**. The webapp saves that setting with `PATCH /projects/{id}`:

```ts
// usePatchProject.ts
// Patches AI Chat Assistant (Novera) settings for a project (PATCH /projects/:projectId).
await authFetch(`${baseUrl}/projects/${projectId}`, { method: "PATCH", ... })
```

The Ballerina backend serves that route. **backend-v2 never registered it**, so the request fell through the mux to its default handler and never reached any code.

## How it was found

By diffing every route the two backends register — 101 Ballerina resources against 106 Go ones. Seven differ; this is the one the frontend actually calls.

Worth recording what the diff ruled out, since "the Go backend is missing AI endpoints" was the obvious theory and it is **wrong**:

- all ten conversation / chat / classification / recommendation routes match one for one
- the agent's HTTP paths match (`/chat`, `/case_classification`, `/recommendations`, `/chat/summary/{p}/{c}`)
- the WebSocket path matches (`/ws?sessionId=`)
- `GET /projects/{id}/features` exists on both

The two backends do differ architecturally — Ballerina reads conversations and history from the AI agent (`/chat/conversations/…`, `/chat/history/…`) while backend-v2 reads them from entity-service — but that is deliberate and complete on both sides. This single route was the gap.

## Exactly one field per request

entity-service enforces it and the Ballerina resource checked before calling:

> `"Only one field can be updated at a time."`

Checked here too, so a mistake returns immediately with that message instead of after a round trip.

**Both fields are `*bool`, and that is load-bearing.** With plain `bool`, a request setting `hasAgent: false` is indistinguishable from one that omits it — so every request would appear to set both fields and none would pass the one-field rule. The tests cover the `false` cases specifically, because that is the shape that breaks.

## Response shape

Returns `response.project`, not the `{message, project}` envelope — matching the Ballerina resource and what the frontend reads.

## Changes

| file | |
|---|---|
| `cmd/server/main.go` | register `PATCH /projects/{id}` |
| `internal/entity/projects.go` | `UpdateProject` client method |
| `internal/entity/types.go` | `UpdateProjectRequest` (+ `FieldCount`), `UpdateProjectResponse`, `UpdatedProjectRef` |
| `internal/handler/projects.go` | `PatchProject` |
| `internal/handler/patch_project_test.go` | route, one-field rule, response shape, UUID validation |
| `openapi.yaml` | the operation and its two schemas — the path had a GET but no PATCH |

## Testing

`go build`, `go vet`, `gofmt` clean. Full suite passes, no pre-existing failures in this module.

The route test goes through a real `http.ServeMux` registered with main.go's exact pattern, so it fails with a 404 if the registration is ever dropped again — which is precisely the bug.

## Summary by CodeRabbit

- **New Features**
  - Added the ability to update a project’s AI assistant settings.
  - Updates can change either agent availability or knowledge-base references, with exactly one setting per request.
  - Successful updates return the refreshed project details.
  - Invalid project IDs, malformed requests, or requests with zero or multiple settings are rejected with clear errors.

- **Documentation**
  - Updated the API specification with the new project settings update endpoint and response formats.


SasmithaDilshan:
Superseded by the same fix targeting `main` — see the newer PR. `main` and `dev-app-csm-portal` have diverged (102/404 commits), so retargeting this one would have dragged the whole branch delta into the diff; rebasing onto `main` was cleaner.

PR #1788, opened 2026-09-16
fix(customer-portal): add the missing PATCH /projects/{id} route

SasmithaDilshan:
Enabling the AI chat assistant returns **404**. The webapp saves that setting with `PATCH /projects/{id}`:

```ts
// usePatchProject.ts
// Patches AI Chat Assistant (Novera) settings for a project (PATCH /projects/:projectId).
await authFetch(`${baseUrl}/projects/${projectId}`, { method: "PATCH", ... })
```

The Ballerina backend serves that route. **backend-v2 never registered it**, so the request fell through the mux to its default handler and never reached any code.

Confirmed still missing on `main` before opening this.

## How it was found

By diffing every route the two backends register — 101 Ballerina resources against 106 Go ones. Seven differ; this is the one the frontend actually calls.

Worth recording what the diff ruled out, because "backend-v2 is missing AI endpoints" is the obvious theory and it is **wrong**:

- all ten conversation / chat / classification / recommendation routes match one for one
- the agent's HTTP paths match — `/chat`, `/case_classification`, `/recommendations`, `/chat/summary/{p}/{c}`
- the WebSocket path matches — `/ws?sessionId=`
- `GET /projects/{id}/features` exists on both

The backends do differ architecturally — Ballerina reads conversations and history from the AI agent, backend-v2 reads them from entity-service — but both are complete. This single route was the whole gap.

## Exactly one field per request

entity-service enforces it and the Ballerina resource checked before calling:

> `"Only one field can be updated at a time."`

Checked here too, so a mistake returns immediately with that message rather than after a round trip.

**Both fields are `*bool`, and that is load-bearing.** With plain `bool`, `hasAgent: false` is indistinguishable from an absent field, so every request would appear to set both and none would pass. The tests cover the `false` cases specifically, since that is the shape that breaks.

## Strict decoding

Ballerina's payload is a **closed** record (`record {| ... |}`), so it already rejects properties it does not declare. `json.Unmarshal` dropped them silently, meaning a caller misspelling `hasKbReferences` got a `200` and no change — the worst failure mode, because it looks like it worked. Decoded with `DisallowUnknownFields`, matching both Ballerina and this PR's own `additionalProperties: false`.

## Changes

| file | |
|---|---|
| `cmd/server/main.go` | register `PATCH /projects/{id}` |
| `internal/entity/projects.go` | `UpdateProject` client method |
| `internal/entity/types.go` | `UpdateProjectRequest` (+ `FieldCount`), `UpdateProjectResponse`, `UpdatedProjectRef` |
| `internal/handler/projects.go` | `PatchProject` |
| `internal/dto/project.go` | `UpdatedProject` + `MapUpdatedProject`, so the frontend contract is not entity-service's type |
| `internal/handler/patch_project_test.go` | route, one-field rule, unknown fields, response shape, UUID validation |
| `openapi.yaml` | the operation and its schemas — the path had a `GET` but no `PATCH` |

The response is `response.project`, not the `{message, project}` envelope, matching the Ballerina resource and what the frontend reads.

## Testing

`go build`, `go vet`, `gofmt` clean on everything this touches. Full module suite passes.

The route test goes through a real `http.ServeMux` registered with main.go's exact pattern, so it fails with a 404 if the registration is ever dropped again — which is precisely the bug.

`internal/handler/cases.go` is unformatted on `main` and left alone; unrelated to this change.


Rashmika998 (commented):
## Summary

This is a well-written, well-tested fix for a real 404 bug. The handler follows existing patterns cleanly, the `*bool` modeling for the false-vs-absent distinction is correct and well-explained, `DisallowUnknownFields` matches the Ballerina closed-record contract, and the test suite directly regresses the route-registration bug. One doc comment placement issue needs fixing; minor suggestions below.

---

apps/customer-portal/backend-v2/internal/entity/projects.go
```diff
@@ -40,6 +40,17 @@ func (c *Client) GetProject(ctx context.Context, id string) (ProjectDetailsView,
 //
 // NOTE: only entity-service's ServiceNow data source supports this route —
 // see cs-tools/entity-service/internal/server/routes.go.
+// UpdateProject patches a project's AI-assistant settings via
```
Rashmika998, line 39: **[Required]** The new `UpdateProject` doc comment is concatenated with `GetProjectMetadata`'s existing doc comment. Because there is no blank line between `// see cs-tools/entity-service/internal/server/routes.go.` (line 42, belonging to `GetProjectMetadata`) and `// UpdateProject patches...` (line 43), `godoc` treats the entire block — starting from `// GetProjectMetadata calls GET /projects/{id}/metadata.` — as the doc for `UpdateProject`. Meanwhile `GetProjectMetadata` (line 54) is left with no doc comment at all.

Fix: insert a blank line before this line so the two comment blocks are separated, and restore `GetProjectMetadata`'s doc comment above its own function. For example:

```go
// GetProjectMetadata calls GET /projects/{id}/metadata.
//
// NOTE: only entity-service's ServiceNow data source supports this route —
// see cs-tools/entity-service/internal/server/routes.go.
func (c *Client) GetProjectMetadata(...) { ... }
```

should remain intact, with `UpdateProject` and its comment placed either before or after (with a blank-line gap).

---

apps/customer-portal/backend-v2/internal/handler/patch_project_test.go
```diff
@@ -0,0 +1,194 @@
...
+		Project: entity.UpdatedProjectRef{ID: id, Name: "Acme", HasAgent: req.HasAgent, HasKbReferences: req.HasKbReferences},
+	}, nil
+}
+
+func boolPtr(b bool) *bool { return &b }
+
+// TestPatchProject_RouteExists is the regression test for the bug this fixes:
+// the Ballerina backend served PATCH /projects/{id} and backend-v2 did not, so
+// toggling the AI assistant returned 404 from the mux rather than reaching any
+// handler. Registered through a real ServeMux with main.go's exact pattern.
```
Rashmika998, line 54: **[Nitpick]** `boolPtr` is declared but never called — all tests use raw JSON string bodies. Go won't error on an unused function (unlike variables), but it's dead code. Either remove it or, if you anticipate needing it for future test cases that assert on `*bool` fields, that's fine to keep.

---

apps/customer-portal/backend-v2/openapi.yaml
```diff
@@ -5962,6 +6013,35 @@ components:
...
+          type: boolean
+          description: Whether the AI assistant is enabled for the project.
+        hasKbReferences:
+          type: boolean
+          description: Whether the assistant cites knowledge-base articles.
+      additionalProperties: false
+      oneOf:
+        - required: [hasAgent]
+        - required: [hasKbReferences]
+    UpdatedProject:
```
Rashmika998, line 6032: **[Suggestion]** The `UpdatedProject` response schema has all properties optional. Consider marking `id` and `name` as `required` — these are always present in entity-service's response and consumers (including generated clients) benefit from knowing they can rely on them. The request schema's `oneOf` constraint is well-done; adding `required` here would bring the response to the same level of specificity.

```yaml
UpdatedProject:
  type: object
  description: The project as it stands after the patch.
  required: [id, name]
  properties:
    ...
```

---

Rashmika998 (approved):

PR #1791, opened 2026-09-16
fix(customer-portal): send bare sysids for the licence download, and log why it fails

SasmithaDilshan:
A licence download returning `500 {"message":"Failed to retrieve license."}` cannot currently be diagnosed from the logs.

`ProcessLicenseDownload` fails three ways that are **not transport errors at all**:

- the application is `PENDING` but the licensing service sent no `name` or `description`
- no application id came back
- the application status is outside the set the flow handles

All three were plain `fmt.Errorf`, so they fell through every case in `summarizeErr` and logged as:

```
upstream request failed
```

Which says the request failed — when in fact it **succeeded** and the *data* was unusable. That sends whoever is debugging to look for an outage that never happened.

It is the same failure `summarizeErr`'s own comment already describes for recommendations, which *"took several rounds of guesswork to place"*.

## The fix

`apierror.DataError` carries a reason for exactly this case: the call succeeded, the response is unusable. `summarizeErr` checks it **first**, ahead of the transport and decode categories, because it is the one kind here that is neither.

```
before   upstream request failed
after    upstream data unusable: application is PENDING but the licensing
         service supplied no name/description for project 6fa0b42d-…
```

## No behaviour change, and Ballerina parity is preserved

The logic still matches the Ballerina implementation **line for line** — same state machine, same guards, same three failure points. I checked head and tail before touching anything; there is no logic bug here, and Ballerina fails identically.

`DataError` is deliberately **not** an `*apierror.Error`, so `mapUpstreamError` still falls through to the handler's own generic message:

- the client sees the same `500` and the same text
- the licensing service's internals stay out of the response

The gain is entirely in the log, which is where diagnosis happens.

## Why not change the status code

`502` was the obvious instinct — nothing in this backend broke. But `mapUpstreamError` maps `502` to `503 + fallbackMsg`, discarding the body, so it would have achieved nothing without also changing shared error mapping for every other caller. Not worth the blast radius for a cosmetic status change on an already-failing path.

## Changes

| file | |
|---|---|
| `internal/apierror/apierror.go` | `DataError` + `NewDataError` |
| `internal/handler/response.go` | `summarizeErr` recognises it, checked first |
| `internal/productconsumption/subscription.go` | the three bare errors become typed, and name the project |
| `internal/handler/data_error_log_test.go` | reason reaches the log · survives a `%w` wrap · does **not** reach the client · existing categories unchanged |

## Testing

`go build`, `go vet`, `gofmt` clean. Full module suite passes, including all five pre-existing `summarizeErr` tests — so the new check does not disturb the existing ordering.

`internal/handler/cases.go` is unformatted on `main` already and left alone; unrelated.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error reporting when upstream services return unusable or unexpected data.
  - Logs now include more specific reasons for data-processing failures instead of generic upstream request messages.
  - Client-facing responses continue to avoid exposing internal error details while returning an appropriate server error.
  - Additional context is captured for subscription and license download failures, including project and application information where applicable.


SasmithaDilshan:
Added the actual fix — the earlier commit only made the failure diagnosable.

## The cause

`projectId` and `deploymentId` reached the product-consumption service as **dashed UUIDs**. That service identifies both by bare ServiceNow sysids — 32 hex characters, no hyphens — so a dashed id matched nothing. The call still succeeded; it just described a project that does not exist, and the flow then failed on data it could not use, producing the `500 "Failed to retrieve license."`

The portal deals in dashed UUIDs everywhere else because **entity-service converts on both sides** — `sysidToUUID` on the way out, `uuidToSysid` on the way in. An id arriving from a portal URL is therefore always dashed, and this is the one boundary where it has to be converted back.

```
before   POST /projects/6fa0b42d-1bfa-a694-a002-c9d3604bcb77/consumption/status
after    POST /projects/6fa0b42d1bfaa694a002c9d3604bcb77/consumption/status
```

## Scope

`uuidToSysID` mirrors entity-service's own helper: hyphens are stripped only from a canonical 8-4-4-4-12 UUID, anything else passes through untouched. An id that is already a bare sysid is unchanged, so the conversion is safe at every call site rather than only where a dashed id happens to arrive today.

Applied to **`projectId` and `deploymentId`** in the three calls that carry them — consumption status (path *and* body), the project-status patch, and the licence call.

**Not** applied to `applicationId`. That is issued *by* this service and goes back in whatever form it came.

## This diverges from Ballerina, deliberately

The Ballerina backend passes the id through unconverted, so **the same bug exists there**. This is a fix on both rather than a port defect — worth knowing if the old portal is still serving anyone.

## Testing

The test asserts on the URLs the client actually builds, through a real `httptest` server: no dashed id reaches any upstream path, the bare sysid does, and the deployment id is converted in the request body too. Plus a table for the conversion itself — uppercase hex, already-bare sysids, wrong hyphen positions, non-hex characters.

Full module suite green.

## The two commits together

1. **diagnosability** — those three data failures logged as `"upstream request failed"`, which is wrong and sends you hunting an outage
2. **this** — the reason they were failing

If the id format turns out not to be the whole story, commit 1 means the next log line says so plainly rather than leaving it ambiguous.

Rashmika998 (approved):

PR #1795, opened 2026-09-16
[CSM Portal] Page through all case attachments instead of capping at 50

rksk:
## Summary
- A case with more than 50 attachments only ever showed the first 50 in the CSM portal — `useGetCsmCaseAttachments` issued a single `POST /attachments/search` with a hardcoded page size and never checked the response's `hasMore`/`total`.
- Converted it to the same bounded paged-loop shape already used by `useGetCsmCaseComments`/`useGetCsmCaseActivities`, so it now fetches every page.
- No backend change: `POST /attachments/search` already supports `pagination.offset/limit` and returns `hasMore`/`total`.
- Kept eager (not lazy/tab-gated): the case-detail activity timeline (`CaseActivitiesFeed`) already renders every attachment inline regardless of which tab is open, so the full list has to be available up front either way. This one hook change also fixes the Attachments tab's count, and "download all" (both were silently capped at 50 too).

## Test plan
- [x] `pnpm run lint` — clean
- [x] `pnpm run test` — new + existing tests for this hook pass (10/10); full-suite failures are a pre-existing, unrelated env-config issue reproduced on a clean `main` checkout
- [x] `pnpm run build` — succeeds, no type errors
- [x] Added unit tests: multi-page accumulation, stop on `hasMore: false`, stop at the safety-bound page count

## Summary by CodeRabbit

* **Bug Fixes**
  * Case attachment searches now retrieve results across multiple pages instead of stopping after the first page.
  * Searches stop safely when all results are fetched, a short page is returned, or the maximum page limit is reached.
* **Tests**
  * Added coverage for multi-page retrieval, single-page results, and the safety limit for continuously paginated responses.


rksk:
Fixed in 968e93431: track a running offset (`offset += rows.length`) instead of deriving it from `page * ATTACHMENTS_PAGE_LIMIT`, and stop only on `!hasMore || rows.length === 0` — a short page that still reports `hasMore: true` no longer gets treated as the end. (Replying at the issue level since this finding had no inline diff anchor — [pullrequestreview-5225552180](https://github.com/wso2-open-operations/cs-tools/pull/1795#pullrequestreview-5225552180).)

Rashmika998 (approved):

PR #1799, opened 2026-09-17
[CSM Portal] Add CS Monitor Dashboard (/cs-monitor-dashboard) — port from dev-app-csm-portal

Ananth-Abi:
## Purpose
The CSM Portal needed a full-screen, kiosk-style "wallboard" view of the
`cs-overview` dashboard for an unattended office display — real numbers, styled
to match the reference wallboard design, with no portal chrome (header,
sidebar, banners, dashboard switcher).

This feature was originally built and reviewed on `dev-app-csm-portal` (PR #1655,
merged there) across two review rounds from `rksk`. Per Rashmika's guidance,
`dev-app-csm-portal` currently holds unrelated migration-specific work and isn't
the right integration point — this PR ports the same feature directly onto
`main` instead, via a clean set of cherry-picks (not a branch merge), so it
carries none of `dev-app-csm-portal`'s unrelated history.

## Goals
- A new `/cs-monitor-dashboard` route: real Asgardeo sign-in required, renders
  only the wallboard view full-screen.
- `cs-overview` itself keeps rendering through the standard dashboard grid
  everywhere else in the portal — unchanged from today.
- Every review finding from both rounds on the original PR is included.

## Approach
- `AuthGuard` gains an optional `bare` prop (default `false`, every other route
  unaffected) that reuses all of its existing sign-in/redirect/token-refresh
  logic but skips the app shell (header/sidebar/banners) once authenticated.
- New `WallboardDashboard` + per-family components (`WallboardCreSection`,
  `WallboardSreSection`, `WallboardStatGrid`, `WallboardStatTile`,
  `WallboardSecondaryStat`, `WallboardPanel`) render the 4-panel layout
  (CRE / SRE / Security Report / FDE), each populated only if the backend's
  `cs-overview` config actually has widgets for that section.
- Shared data logic (`useWallboardTileData`) factors out what the two tile
  components have in common — fetch, current-user deferral, last-known-value-
  on-error, link guarding.
- `useWidgetData` converted from a ~12-arg positional signature to a single
  options object (all existing call sites updated); a wallboard tile's 60s
  auto-refresh self-throttles against the shared fetch-concurrency queue
  (`resolveWidgetRefetchInterval` + `widgetFetchQueueDepth`) so a slow backend
  degrades to "refreshes less often" instead of an unbounded backlog.
- `CS_Monitor_Dashboard.png` (design reference) is the source of truth for
  layout/labels — including the CRE "Escalations" tile's exact primary-grid
  position and amber emphasis.
- Two small additional commits beyond the original feature/review commits,
  needed only because `main` has evolved independently of `dev-app-csm-portal`:
  coalescing a nullable `BeDashboardWidget.query` type, and converting one
  `useWidgetData` call site in `WidgetInlineDrilldownPanel.tsx` (a component
  that only exists on `main`) to the new options-object form.

## User stories
As a CSM team member glancing at the office wallboard, I see live CRE/SRE
(and, once configured, Security Report/FDE) metrics styled consistently with
the reference design, refreshing automatically, without needing to interact
with the portal at all.

## Release note
Adds a standalone, full-screen `/cs-monitor-dashboard` kiosk view of the CS
Overview dashboard for office wallboard displays.

## Documentation
N/A — internal CSM portal, no product documentation impact.

## Training
N/A

## Certification
N/A — internal tooling feature, no exam impact.

## Marketing
N/A

## Automation tests
 - Unit tests
   > 21 new/changed test files under `features/csm-dashboard` and `layouts`
   > covering the wallboard components, `useWallboardTileData`,
   > `useWidgetData`'s options object + self-throttle, `AuthGuard`'s `bare`
   > mode (loading/redirect/Suspense/stale-redirect-key states), and the
   > display-name alias table. 701 tests passing in the affected areas;
   > full-suite run confirms no regressions elsewhere.
 - Integration tests
   > Manually verified against a running dev backend (`localhost:3001/cs-monitor-dashboard`):
   > sign-in flow, live data rendering, auto-refresh behavior.

## Security checks
 - Ran FindSecurityBugs plugin and verified report? N/A — frontend TypeScript, no Java

## Samples
N/A

## Related PRs
- wso2-open-operations/cs-tools#1655 (original feature + first review round, on `dev-app-csm-portal`)
- wso2-open-operations/cs-tools#1685 (the "Escalated Cases" → "Escalations" label fix, on `dev-app-csm-portal`)

## Migrations (if applicable)
N/A — no data/schema changes. The `/cs-monitor-dashboard` route depends on the
`cs-overview` dashboard's backend `DASHBOARDS_CONFIG` having Security Report /
FDE widgets to show those two panels; without them, only CRE/SRE render
(that's config, not a migration).

## Test environment
- Node 24 / pnpm 10, Vitest (jsdom)
- Chrome, `localhost:3001/cs-monitor-dashboard` against the dev backend

## Learning
N/A — ported an already-reviewed feature; the main new technique here was a
targeted, file-by-file cherry-pick port across two independently-diverged
branches rather than a full branch merge, to avoid pulling in unrelated history.

## Summary by CodeRabbit

- **New Features**
  - Added a standalone, sign-in-protected CS Monitor wallboard at `/cs-monitor-dashboard`.
  - Added full-screen CRE, SRE, Security, and FDE sections with ordered metrics, visual status emphasis, and drill-down links.
  - Added automatic updates every 60 seconds without portal navigation or layout chrome.
  - Added loading, error, and last-known-data states for reliable monitoring.

- **Bug Fixes**
  - Prevented dashboard refreshes from overwhelming queued data requests.
  - Preserved previously loaded metrics and drill-down links during background refresh failures.


apps/csm-portal/webapp/src/features/csm-dashboard/api/useWallboardTileData.ts
```diff
@@ -0,0 +1,124 @@
...
+  const state: WallboardTileState =
+    isLoading || awaitingCurrentUser ? "loading" : isError && !data ? "error" : "value";
+
+  // A disabled query (`enabled: !awaitingCurrentUser`) reports `isLoading`
+  // false in TanStack Query v5 — that flag is "actively fetching", not
+  // "deferred, never fetched" — so `awaitingCurrentUser` has to be checked
+  // here too, or a still-loading tile whose filters carry the unresolved
+  // `__current_user__` placeholder could get wrapped in a link built from
+  // it.
+  const linkable = !!href && !isLoading && !isError && !awaitingCurrentUser;
```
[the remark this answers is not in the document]
Ananth-Abi: Fixed. linkable now derives from the already-correct state (!!href && state === "value") instead of re-checking isError directly — a tile showing a real, if stale, cached number during a background refetch failure stays clickable. Regression test added.

apps/csm-portal/webapp/src/features/csm-dashboard/components/WallboardCreSection.tsx
```diff
@@ -0,0 +1,105 @@
...
+      );
+    }
+    byName.set(w.displayName, w);
+  }
+  const primary = CRE_PRIMARY_ORDER.map((name) => byName.get(name)).filter(
+    (w): w is BeDashboardWidget => w !== undefined,
+  );
+  const primaryIds = new Set(primary.map((w) => w.widgetId));
+  const secondary = sortByFixedOrder(
+    widgets.filter((w) => !primaryIds.has(w.widgetId)),
```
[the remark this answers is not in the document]
Ananth-Abi: Fixed. The secondary list is now built from byName's own deduplicated values instead of the raw widgets array, so a widget that lost a primary-slot alias collision doesn't also render a second time as a secondary tile. Regression test added.

apps/csm-portal/webapp/src/layouts/AuthGuard.tsx
```diff
@@ -286,19 +318,32 @@ export default function AuthGuard(): JSX.Element {
...
+  // states. `AppLayout` is otherwise the only place in the tree that
+  // provides a `Suspense` boundary, so `bare` mode brings its own — falling
+  // back to `BareAuthLoader`, the same centered progress bar the pending
+  // and sign-in-redirect states show.
+  const authenticatedContent = bare ? (
+    <Suspense fallback={<BareAuthLoader />}>
+      <Outlet />
+    </Suspense>
+  ) : (
+    <AuthorizedAppShell />
```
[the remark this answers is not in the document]
Ananth-Abi: Good catch — fixed. Added BareAuthorizedContent, bare mode's own counterpart to AuthorizedAppShell: the same /users/me entitlement gate (loading → not-authorized → outlet) every other route already enforces, just rendered without any AppLayout chrome instead of skipped entirely. The routed page (and its widgets' own authenticated queries) now only mount once portal access is confirmed, not the instant Asgardeo sign-in succeeds. Two regression tests added (stays on BareAuthLoader while /users/me is pending; shows the chrome-free not-authorized page on a 401, in both cases never reaching the routed content).

apps/csm-portal/webapp/src/layouts/AuthGuard.tsx
```diff
@@ -184,6 +190,62 @@ function AuthorizedAppShell(): JSX.Element {
...
+
+  if (notAuthorized) {
+    return (
+      <Box sx={{ height: "100dvh", width: "100%", display: "flex", flexDirection: "column" }}>
+        <NoPortalAccessPage />
+      </Box>
+    );
+  }
+
+  return (
```
Rashmika998, line 248: **[Required]** `BareAuthorizedContent` has an access-control gap: when `useCurrentUser()` returns `isError: true` for a **non-401/403 reason** (network error, 5xx from `/users/me`), `notAuthorized` is `false`, neither branch above fires, and the function falls through here — rendering `<Outlet />` before portal entitlement has been confirmed.

This is precisely the "fail open" scenario the commit that introduced this component cited CWE-862 to prevent. For an unattended kiosk machine a transient backend hiccup on `/users/me` would periodically let the wallboard render and fire widget API calls without confirmed portal access.

Suggested fix — add an explicit `isError` branch before the fallthrough:

```tsx
// /users/me returned a non-auth error (5xx, network failure);
// don't render the outlet without confirmed portal entitlement.
if (isError) {
  return (
    <Box sx={{ height: "100dvh", width: "100%", display: "flex", flexDirection: "column" }}>
      <BareAuthLoader />
    </Box>
  );
}
```

---

apps/csm-portal/webapp/src/layouts/AuthGuard.test.tsx
```diff
@@ -404,3 +409,210 @@ describe("AuthGuard's AppLayout showCaseTabs wiring while auth itself hasn't res
...
+
+    // Still suspended: AuthGuard's own Suspense fallback (BareAuthLoader)
+    // shows, not a blank tree or a thrown error.
+    expect(screen.getByTestId("bare-auth-loader")).toBeInTheDocument();
+    expect(screen.queryByTestId("lazy-loaded-content")).not.toBeInTheDocument();
+
+    // Once the lazy import resolves, the real content takes over.
+    await waitFor(() => expect(screen.getByTestId("lazy-loaded-content")).toBeInTheDocument());
+  });
+});
```
Rashmika998, line 652: **[Suggestion]** The bare-mode tests cover loading, successful sign-in, and stale-redirect-key clearing, but not the case where `useCurrentUser()` returns `isError: true` with a **non-401/403** error (e.g., `ApiError(500)`). A test for that scenario would catch a regression if the `BareAuthorizedContent` gap flagged above is ever re-introduced:

```tsx
it("shows BareAuthLoader (not Outlet) when /users/me fails with a non-auth error in bare mode", async () => {
  asgardeoState.isSignedIn = true;
  currentUserState.isLoading = false;
  currentUserState.isError = true;
  currentUserState.error = new ApiError(500, "Internal Server Error");

  await act(async () => {
    renderBareAuthGuard();
  });

  expect(screen.queryByTestId("bare-route-content")).not.toBeInTheDocument();
  expect(screen.getByTestId("bare-auth-loader")).toBeInTheDocument();
});
```

---

apps/csm-portal/webapp/src/features/csm-dashboard/components/WallboardDashboard.tsx
```diff
@@ -0,0 +1,351 @@
...
+ * rather than repeated three times with the risk of one copy drifting
+ * from the other two. `sx` merges in on top of (and can override) the
+ * defaults below — the loaded state uses this to swap `minHeight` for a
+ * fixed `height` plus its own flex-column layout. */
+function WallboardPageFrame({
+  children,
+  sx,
+}: {
+  children: ReactNode;
+  sx?: Record<string, unknown>;
```
Rashmika998: **[Suggestion]** `sx?: Record<string, unknown>` loses MUI's `sx` type-safety — invalid style properties are silently accepted and valid ones lose autocomplete / exhaustiveness checking. Use `SxProps<Theme>` instead, and note that when composing `SxProps` values MUI's array syntax is safer than object spread (it handles the callback-style `sx` entry correctly):

```tsx
import type { SxProps, Theme } from "@mui/material/styles";

function WallboardPageFrame({
  children,
  sx,
}: {
  children: ReactNode;
  sx?: SxProps<Theme>;
}): JSX.Element {
  return (
    <Box
      sx={[
        { bgcolor: "#0f1420", minHeight: "100dvh", p: 2 },
        ...(Array.isArray(sx) ? sx : [sx]),
      ]}
    >
      {children}
    </Box>
  );
}
```

---

apps/csm-portal/webapp/src/features/csm-dashboard/components/WallboardDashboard.tsx
```diff
@@ -0,0 +1,351 @@
...
+    .filter((w) => w.shape !== "count")
+    .map((w) => `"${w.displayName}" (${w.shape})`);
+  useEffect(() => {
+    if (import.meta.env.DEV && droppedShapes.length > 0) {
+      console.warn(
+        `[WallboardDashboard] ${droppedShapes.length} non-count widget(s) on "${dashboardId}" ` +
+          `are not shown on the monitor dashboard: ${droppedShapes.join(", ")}`,
+      );
+    }
+    // Re-warn only when the set of dropped widgets actually changes, not on
```
Rashmika998: **[Suggestion]** The `eslint-disable-next-line react-hooks/exhaustive-deps` is needed because `droppedShapes` is a new array on every render even when the widget shapes haven't changed, making it an unstable dependency. A `useMemo` makes it stable and eliminates the suppression:

```tsx
const droppedShapes = useMemo(
  () =>
    (data?.widgets ?? [])
      .filter((w) => w.shape !== "count")
      .map((w) => `"${w.displayName}" (${w.shape})`),
  [data?.widgets],
);

useEffect(() => {
  if (import.meta.env.DEV && droppedShapes.length > 0) {
    console.warn(
      `[WallboardDashboard] ${droppedShapes.length} non-count widget(s) on "${dashboardId}" ` +
        `are not shown on the monitor dashboard: ${droppedShapes.join(", ")}`,
    );
  }
}, [dashboardId, droppedShapes]); // no suppression needed
```

---

Rashmika998 (commented):
## Summary

Solid port of an already-reviewed feature. The `AuthGuard bare` design is clean, `useWallboardTileData` extraction eliminates meaningful duplication, the self-throttling 60 s poll is well-motivated, and the options-object refactor for `useWidgetData` is a clear improvement. One access-control gap in `BareAuthorizedContent` should be closed before merge; the rest are suggestions and a nitpick.

---

apps/csm-portal/webapp/src/layouts/AuthGuard.tsx
Ananth-Abi: Good catch, and a real gap — fixed. Added an explicit isError branch that holds on BareAuthLoader instead of falling through to the outlet. Kept it a straight <BareAuthLoader /> return (it's already its own full-viewport frame, same as the isLoading branch right above it) rather than re-wrapping it in another frame Box. Deliberately diverges from AuthorizedAppShell's fail-open behavior on this one point — documented why in the function's own doc comment (unattended kiosk with no one to retry vs. a signed-in employee who can). Regression test added.

apps/csm-portal/webapp/src/layouts/AuthGuard.test.tsx
Ananth-Abi: Added, following the existing bare-mode tests' act + rerender pattern in this file for consistency with the other assertions around it.

apps/csm-portal/webapp/src/features/csm-dashboard/components/WallboardDashboard.tsx
Ananth-Abi: Fixed with your suggested array-composition approach. One adjustment: imported SxProps/Theme from @wso2/oxygen-ui rather than @mui/material/styles — @mui/material isn't a direct dependency in this repo (oxygen-ui wraps it and re-exports * from '@mui/material' through its own root), so the @mui/material/styles import wouldn't have resolved. Verified tsc is clean with the oxygen-ui import.

apps/csm-portal/webapp/src/features/csm-dashboard/components/WallboardDashboard.tsx
Ananth-Abi: Fixed exactly as suggested.

Rashmika998 (approved):

PR #1806, opened 2026-09-17
feat(entity-service): support excluding closure states and subscription types in project search

DaniduWijendra:
## Purpose
Adds the ability to exclude closure states and subscription types from a project search. This is the entity-service capability behind an "all customer projects" audience-targeting feature being built on top of this for the CSM portal's announcement create flow — that feature needs to express "every customer project except Restricted/Suspended, except Cloud Support/Cloud Evaluation Support," and neither exclusion existed on `SearchProjects` before this PR (only a single-value *include* filter existed for closure state; no filter at all existed for subscription/project type).

## Goals
- Let a caller exclude one or more closure states (e.g. Restricted, Suspended) from `POST /projects/search`.
- Let a caller exclude one or more subscription types (e.g. cloud_support, cloud_evaluation_support).
- Keep the Postgres data source's behavior explicit — reject the new fields rather than silently ignore them — since the only current consumer (announcements) is ServiceNow-only anyway.

## Approach
No UI in this repo layer; this is a backend-only, additive API change.

- **Postgres data source** (`project_service.go`): rejects a non-empty value for either new field with a `ValidationError`, mirroring `CreateCase`'s existing "only type case is supported" precedent for this data source.
- **ServiceNow data source** (`sn_project_service.go`): applied as a **post-fetch filter**, not a new upstream request parameter — the Choreo/Ballerina integration service behind ServiceNow project search isn't part of this repo, so there's no way to verify it would honor a new list-valued filter param. Instead, pages through every match (capped at the data source's existing 50-row page limit) and filters on `ClosureState`/`SubscriptionType`, both already reliably returned today. The caller's own `pagination.offset/limit` is applied to the filtered result afterward, and `total`/`hasMore` reflect the filtered count, not ServiceNow's raw `totalRecords`.
- **Safety bound fix** (addressing review feedback): the internal paging loop is capped at 200 pages as a guard against a wrong/always-true upstream `hasMore`. Originally, hitting that cap before the upstream's own total was exhausted returned whatever had been collected so far as if it were the complete result — silently under-reporting. Now returns a `503 ServiceUnavailableError` instead, so a caller never mistakes a truncated result for an authoritative one.

`openapi.yaml` updated with both new request fields.

## User stories
As a CS/Security engineer creating a customer announcement, I want to target "all customer projects except the ones that shouldn't receive it" without hand-picking every project from a 97-project list or running a ServiceNow Flow Designer dry-run first.

## Release note
`POST /projects/search` (ServiceNow data source) accepts two new optional filters: `excludeClosureStates` (array of closure-state strings) and `excludeSubscriptionTypes` (array of subscription-type enum values).

## Documentation
N/A — internal platform API consumed by other cs-tools services; documented in this PR's own `openapi.yaml` update, no customer-facing docs impact.

## Training
N/A — no training-content impact.

## Certification

## Marketing
N/A — internal backend capability, no marketing impact.

## Automation tests
- Unit tests
  - 10 new tests across `entity-service/internal/service/project_service_test.go` and `entity-service/internal/service/sn_project_exclude_filters_test.go`: invalid-value rejection, Postgres-path rejection, single-page exclusion filtering with correct `total`, pagination windowing over the *filtered* result (not the raw upstream page), multi-page pass-through when an exclude filter is set, and the truncation-guard regression test added in response to review.
  - Full existing suite (134 files, all packages) still passes; `go vet ./...` clean.
- Integration tests
  - None — this repo has no integration-test harness against a live ServiceNow instance; the ServiceNow paths are covered via a mocked HTTP client (`newTestSNClient`), matching the existing test convention for this package.

## Security checks
 - Followed secure coding standards in http://wso2.com/technical-reports/wso2-secure-engineering-guidelines? Yes
 - Ran FindSecurityBugs plugin and verified report? N/A — Go project, no Java/FindSecurityBugs tooling; ran this repo's own documented Go equivalent instead: `gosec -fmt=text ./...` (0 issues, 134 files scanned) and `govulncheck ./...` (no vulnerabilities found)
 - Confirmed that this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets? Yes

## Samples
N/A — no new samples.

## Related PRs
None yet. This is the first of several small PRs splitting out a larger local feature branch (audience targeting, security-label tagging on both the CSM and customer portals) — the rest will follow once this one is reviewed, so each piece stays independently reviewable.

## Migrations (if applicable)

## Test environment
Verified locally on macOS with Go 1.27.1 (`go build`, `go vet`, `go test ./...` against the full module). No live ServiceNow instance available in this environment — ServiceNow-path behavior is verified via the mocked-HTTP-client unit tests listed above, not a live integration run.

## Learning
N/A.

## Summary by CodeRabbit

* **New Features**
  * Added optional ServiceNow project search filters to exclude projects by closure state or subscription type.
  * Filtering occurs before pagination, ensuring totals, offsets, and limits reflect the filtered results.

* **Bug Fixes**
  * Added validation for unsupported exclusion values.
  * Requests using these filters with the Postgres data source now return a validation error.
  * Searches return a service-unavailable error instead of incomplete results if ServiceNow data cannot be fully retrieved.


DaniduWijendra:
@coderabbitai review

rksk (approved):

PR #1818, opened 2026-09-18
[Customer Portal][CSM Portal] Add the missing OPEN conversation state

Hesara2003:
## Purpose

Conversations in ServiceNow's "Open" state were invisible to both portals: they rendered with **no status at all**, and filtering by that state returned **every** conversation instead of the matching ones — a filter that looked like it worked and did nothing.

Reproduced against staging data:

| Request | Result |
|---|---|
| no filter | 1072 |
| `stateKeys: [1]` — **Open** | **1072** — every conversation |
| `stateKeys: [2]` — Active | 257 |
| `stateKeys: [3]` — Resolved | 6 |
| `stateKeys: [99]` — nonsense | **1072** — every conversation |

The same gap had a second, quieter symptom: **163 of 1073 conversations rendered with no status at all**, because an unmapped state key decodes to an empty enum before it ever reaches the portal.

## Root cause

`GET /projects/{id}/filters` passes ServiceNow's conversation-state choice list straight through, and that list carries six states including `{"id":"1","label":"Open"}`. This backend's translation table (`conversation_enum_mapping.go`) has only ever covered ids 2–6 — `ACTIVE`, `RESOLVED`, `CONVERTED`, `ABANDONED`, `CLOSED`. entity-service's own `snConversationStateKeyMap` stops at 2 as well, so no layer has ever known a state 1.

`conversationIDsToEnums` dropped the unmapped id, as its own comment described:

```go
// ... silently skipping any id with no known mapping.
```

That left `States` empty — and an empty `States` is **"no state filter"** to entity-service, not "no matches". So the search ran unfiltered and returned the whole project.

The `[99]` row above is the part that matters: this was never specific to Open. **Any** unrecognised state id widened the search to everything, with nothing in the response to indicate the filter had been ignored.

## Root cause, confirmed

ServiceNow's state key **1 is "Open"** — a chat created but never answered by the assistant; it becomes **Active** once the assistant replies. The two are genuinely distinct states, not synonyms.

No layer of this system mapped key 1. entity-service's `snConversationStateLabelMap` starts at 2, so an Open conversation decoded to an **empty enum** before the portal saw anything — which is why those conversations had no status, and why the original id was already gone by the time it reached this backend. The portal's own table started at 2 for the same reason, so the filter had nothing to send upstream.

## Approach

**1. The missing state is added end to end — entity-service and both portals.**

| Component | Change |
|---|---|
| entity-service | `ConversationStateOpen = "OPEN"`; `snConversationStateKeyMap` 1↔OPEN; `snConversationStateLabelMap` 1→OPEN; added to `validConversationState`; enum documented in `openapi.yaml` |
| customer-portal backend-v2 | `conversationStateIDs["OPEN"] = "1"`, `conversationStateLabels["OPEN"] = "Open"` |
| CSM backend | `validConversationStates` accepts OPEN; enum documented in `openapi.yaml` |
| CSM webapp | `BeConversationState` gains OPEN; rendered as its **own chip** rather than folded into Active, so an unclaimed conversation is distinguishable from one already being worked |

This fixes both symptoms at once, in both portals: the filter can translate Open, and those conversations render with a status instead of nothing.

OPEN is deliberately **read-only** — a conversation starts there and is never PATCHed back to it, so it stays out of the transition allow-list for `PATCH /conversations/{id}`.

**2. The filter fails closed.** `conversationIDsToEnums` reports an id it cannot translate rather than skipping it, `BuildEntitySearchConversationsRequest` returns an error, and the handler answers `400 Unsupported conversation state filter.` Kept as defence in depth: with (1) in place no current state hits it, but the next state ServiceNow adds will produce a clear error rather than a silently unfiltered search. A partially-mappable request is refused for the same reason.

**3. The filter metadata is left untouched.** An earlier revision trimmed the dropdown to mappable states, which would have hidden the 163 Open conversations entirely. Reverted — the right fix was to support the state, not to hide it.

## Verified against a live backend

Local `backend-v2` → local `entity-service`, both carrying this change:

| State | Count |
|---|---|
| **1 Open** | **163** — previously unfilterable and statusless |
| 2 Active | 257 |
| 3 Resolved | 6 |
| 4 Converted | 116 |
| 5 Abandoned | 522 |
| 6 Closed | 9 |
| **sum** | **1073** |
| **total, unfiltered** | **1073** |

The six states now reconcile to the total exactly. Before the change the five known states summed to 910 and the missing 163 were invisible.

Sampling those 163 directly: `state: {"id":"1","label":"Open"}` where it was previously `{}` — zero blanks.

`stateKeys: [99]` still returns `400`, confirming the fail-closed path.

## Deployment — entity-service first

entity-service converts the ServiceNow key to its own enum before either portal sees it. **With only a portal deployed nothing changes**: entity-service still emits an empty enum, so the state stays blank and the filter has nothing to match. Deploy entity-service first, then the portals.

## Scope note — the same shape exists elsewhere, but is not currently firing

`caseIDsToEnums`, `crIDsToEnums` and the call-request state mapper all skip unmapped ids the same way. I checked each map against the choice list the API actually returns:

| Filter | ids returned by `/filters` | covered by the map |
|---|---|---|
| conversations | 1, 2, 3, 4, 5, 6 | 2–6 — **gap at 1, fixed here** |
| cases | 1, 3, 6, 10, 18, 1003, 1006 | all 7 |
| call requests | 1–8 | all 8 |
| change requests | −2, −1, 0, 1, 2, 3, 4, 5 | all 8 |

Conversations was the only live defect. The other three are latent — complete today, and would fail the same way if a choice list ever gained an entry. Left alone here rather than widening a blocker fix; worth a follow-up. The fail-closed change means such a gap would now surface as a `400` in this one endpoint rather than silently wrong results.

## User stories

As a customer-portal user filtering chat history by state, I see only conversations in that state — and if the state cannot be applied, I am told so rather than shown everything.

## Release note

- [Customer Portal] Fix chat-history state filtering silently returning every conversation instead of applying the selected state.

## Documentation

N/A — the only contract change is a new `400` on an input that previously produced wrong results.

## Automation tests

`internal/dto/conversation_state_filter_test.go` (new):

- Open (id 1) translates to `OPEN` — the state that was missing;
- an OPEN conversation renders as `{id: 1, label: "Open"}` rather than no state at all;
- **every state the choice list offers is filterable**, all six ids — the invariant the original bug broke;
- an unknown id is refused rather than silently dropped;
- a partially-mappable request (`[2, 1]`) is refused rather than silently narrowed to `[ACTIVE]`;
- supported ids still translate to entity-service's enum vocabulary, and project scope still comes from the path;
- no state filter at all remains valid — "everything" is a legitimate answer when the caller asked for everything.

`internal/dto/type_alignment_test.go` updated for the new two-value signature.

Full suite green, `go vet` and `gofmt` clean, `gosec` 0 issues.

## Verified against a live backend

Local `backend-v2` against the shared upstream services, same project as the reproduction above:

| Request | Before | After |
|---|---|---|
| `stateKeys: [1]` | 1072 | **400** |
| `stateKeys: [99]` | 1072 | **400** |
| `stateKeys: [2]` | 257 | 257 |
| `stateKeys: [3]` | 6 | 6 |

## Security checks

 - Ran FindSecurityBugs plugin and verified report? N/A (Go service)

## Test environment

- macOS, Go 1.26. Local `backend-v2` against the shared upstream services.

## Summary by CodeRabbit

- **New Features**
  - Added support for the **Open** conversation state across conversation search, results, filters, and CSM Portal displays.
  - Open conversations now appear with a dedicated label and warning-style status indicator.
  - Search filters correctly include Open conversations instead of broadening results unexpectedly.

- **Bug Fixes**
  - Unsupported conversation-state filters now return a clear 400 error rather than running an unfiltered search.

- **Documentation**
  - Updated API definitions to include the Open conversation state.


SSSD-2001:
@coderabbitai review

rksk (approved):

PR #1819, opened 2026-09-18
[CSM Portal] add missing Open conversation state

rksk:
CSM's conversations/chats have a 6th state on the backing data source ("Open") that CSM never mapped. Any conversation in that state showed `state: null` in the CSM portal and could not be selected in the state filter, while the customer portal displayed it correctly (its backend passes the raw label through untouched rather than looking it up in a fixed map).

Add "Open" as a genuine 6th CSM conversation state end-to-end: entity-service map/enum, CSM backend filter allowlist, and CSM webapp type/label/filter option/chip. It renders as its own distinct chip (not folded into the existing "Active" group) so a CS engineer can immediately tell an unclaimed conversation apart from one already being worked.

Three commits, one per layer:
- `entity-service`: extend the state key/label maps and the search-filter allowlist with the new state; add the corresponding domain constant; extend the OpenAPI enum. Left it out of the PATCH transition allowlist since a conversation only ever originates in this state and is never transitioned back into it.
- CSM backend: extend the conversation search filter's state allowlist so the new value is no longer rejected before reaching entity-service; extend the OpenAPI enum.
- CSM webapp: extend the conversation state union type, the filter's option list, the display label, and give the new state its own chip color (existing `warning`/amber role, since `error`/`info`/`success`/`default` were already used by the other groups) instead of grouping it with Active.

As a CS engineer, I can see and filter for conversations that are in the new "Open" state, instead of them showing a blank status and being invisible to the state filter.

Fix: conversations in the "Open" state now display and filter correctly in the CSM portal (previously shown with a blank/null status).

N/A — no existing in-app help topic documents the conversation state list; nothing to update.

## Training
N/A — no training content references conversation states.

## Certification
N/A — no certification impact.

## Marketing
N/A — internal bug fix, no external-facing feature.

## Automation tests
 - Unit tests
   > Code coverage information

Added unit tests for the new state's chip grouping in the CSM webapp. No conversation-specific test file exists in entity-service or the CSM backend, so none were added there (existing coverage patterns for the other 5 states are likewise implicit, not per-state unit tests).
 - Integration tests

None added; existing `go test`/`pnpm test` suites pass unchanged plus the new webapp unit tests.

## Security checks
 - Ran FindSecurityBugs plugin and verified report? N/A — Go/TypeScript change, no Java code touched

## Samples
N/A

## Related PRs
None

## Migrations (if applicable)
N/A — no schema/data migration involved.

Verified with `go build`/`go vet`/`go test` (entity-service, CSM backend) and `pnpm build`/`pnpm lint`/`pnpm test` (CSM webapp) on macOS (Darwin), Go and Node toolchains per this repo's `go.mod`/`package.json`. Live verification against a dev instance of the backing data source is still pending (tracked in the originating task file, not yet run).

Traced the raw state value on the live dev data source to confirm it as a genuine distinct state (not a mislabeled duplicate of an existing one) before mapping it, rather than guessing from code alone.


rksk:
Handled via https://github.com/wso2-open-operations/cs-tools/pull/1818#event-31368019360

PR #1823, opened 2026-09-18
[Entity-Service] Fix comments not loading for non-CASE case-like types; CloseNotes/ResolutionNotes swap

Rashmika998:
## Purpose
Two real, live bugs found by comparing the ServiceNow and Postgres responses for the same work items via HAR captures, plus one upstream migration rename that needed following.

## Goals
Fix both bugs at the root, verify against live data, and lock them in with tests.

## Approach

**Comments silently empty for non-`CASE` case-like types.** `csm-portal-backend`'s case-scoped comment endpoint forwards to entity-service's generic `POST /comments/search` with `referenceType:"case"` injected into the payload. `comment_repo.go`'s `ReferenceTypeToWorkItemType` mapped `"case"` to a single `"CASE"` `work_item_type_enum` value — so a work item whose real type is `ENGAGEMENT`/`SERVICE_REQUEST`/`SECURITY_REPORT_ANALYSIS`/`ANNOUNCEMENT` returned **zero** comments through this path, even with real comment rows present. `case_repo.go`'s own `GetCaseByID`/`SearchCases` have matched all five case-like types since "Case-like work_item types" landed; `comment_repo.go` was simply never updated to match. Changed `ReferenceTypeToWorkItemType` to map to `[]string` and bind an array (`wi.type = ANY($n::work_item_type_enum[])`) in both `SearchComments` and `CreateComment`.

Verified against a live record (`work_item.type = 'SERVICE_REQUEST'`, 5 real comment rows) that previously returned `total: 0` through this path and now returns all 5, matching the ServiceNow response exactly.

**`closeNotes` silently swapped with `resolutionNotes`.** `GetCaseByID` scanned the real `close_notes` column (already correctly `COALESCE`d across all five case-like tables) into a variable named `resolutionNotes`, assigned to `CaseView.ResolutionNotes` instead of `CaseView.CloseNotes`. This schema has no separate `resolution_notes` column at all — only `close_notes`, on every case-like table — and the ServiceNow-backed path treats `CloseNotes`/`ResolutionNotes` as genuinely distinct fields sourced from two different upstream values. Fixed by scanning into `closeNotes` and assigning it to `cv.CloseNotes`; `cv.ResolutionNotes` now correctly stays `nil` on this data source rather than being double-filled from `close_notes`.

**Also following an upstream schema change**: `daily_usage_summary.usage_type` was renamed to `count_type` in an in-place edit to the already-merged migration `000054` — updated `instance_repo.go`'s query to match (would otherwise fail once the renamed migration is applied).

**Unrelated cleanup**: found and fixed a merge artifact blocking `go vet`/`go test` entirely — `directory_routes_test.go` had `TestPostgresOnlyRoutesAreAbsentWithoutAPool` declared twice in the same file (an older, less-complete copy alongside a newer one, with a stray misplaced doc comment). Not something this PR's changes caused, but needed fixing to get a clean build for verification.

## User stories
N/A — bug fixes to existing endpoints, no new user-facing behavior.

## Release note
Fixed case comments not appearing for engagements, service requests, security report analyses, and announcements (only plain cases worked before). Fixed a case's close notes always showing as empty on the Postgres data source.

## Documentation
`entity-service/CLAUDE.md` updated with both root causes and how they were found.

## Training
N/A

## Certification
N/A

## Marketing
N/A

## Automation tests
New tests locking in `ReferenceTypeToWorkItemType["case"]`'s full five-type coverage (and confirming the other reference types weren't accidentally widened). `go test ./...` passes (one known pre-existing unrelated failure: `TestSNCaseService_CreateCase_PublishesCaseCreated`).

## Security checks
- `gosec ./...`: 0 issues (177 files scanned)
- `govulncheck ./...`: no vulnerabilities found

## Samples
N/A

## Related PRs
None.

## Migrations
No new migrations. Follows an in-place rename in the already-merged `000054`.

## Test environment
Verified directly against live staging data: the comment fix was confirmed against a real work item with 5 real comment rows that were previously invisible through the affected path.

## Learning
N/A

## Summary by CodeRabbit

* **Bug Fixes**
  * Comments now correctly appear for all case-like work items, including engagements, service requests, security report analyses, and announcements.
  * New comments can be created on all supported case-like work item types.
  * Case details now display close notes in the correct field instead of resolution notes.
  * PostgreSQL-only endpoints are no longer exposed when PostgreSQL is unavailable.


Rashmika998:
Addressed both outside-diff-range findings from CodeRabbit's review — `workItemTypes` in `comment_repo.go` is now cast through `::text[]` before `::work_item_type_enum[]` in both `CreateComment` and `SearchComments`. This repository never registers `work_item_type_enum`/`_work_item_type_enum` with pgx, so binding a `[]string` directly to the enum array type has no encode plan — the same class of issue already fixed for every other enum array bind in this codebase (`time_card_repo.go`, `case_repo.go`, `escalation_repo.go`, `instance_repo.go`). Re-verified against the same live record used to confirm the original fix — still returns all 5 comments correctly with the safer cast in place.

cloby99 (approved):

PR #1827, opened 2026-09-18
[Entity-Service] Native GitHub issue ↔ change request sync

SasmithaDilshan:
Replaces the ServiceNow `[GitHub Integration]` flows with a native implementation in entity-service. Ported from the live instance rather than from documentation — `docs/github-cr-sync-spec.md` records what was actually there.

## The architecture, and why

ServiceNow **never called the GitHub Issues API** for these flows. It fired `repository_dispatch` events and a GitHub Actions workflow in each mapped repository did the work — commenting, labelling, closing the issue. This PR does the same thing, so the workflows already deployed in those repositories keep working.

```
Postgres trigger → github_outbound_queue → worker
  → POST /repos/{owner}/{repo}/dispatches
  → GitHub Actions workflow → the issue
```

Three event types, with the exact field names those workflows read:

| Event type | client_payload |
|---|---|
| `servicenow-note` | `issue_number`, `note_text`, `note_type`, `case_number`, `case_sys_id`, `sn_user` |
| `servicenow-case-update` | `action` (`closed`\|`assigned`), `github_issue_number`, `case_number`, `case_sys_id`, `resolution_notes`, `assigned_to`, `sn_user` |
| `servicenow-cr-update` | `github_issue_number`, `cr_number`, `cr_sys_id`, `cr_state`, `assigned_to`, `case_sys_id`, `planned_start`, `planned_end`, `action` |

`note_text` is sent **raw** — the workflow strips ServiceNow's `[code]` markers and HTML itself, and converting here would change what it receives.

**Inbound** is unchanged in shape: `POST /webhooks/github`, HMAC-verified, creating and updating change requests from labelled issues, mirroring comments, and enforcing the state guards — refusing to close below Review, reverting protected labels, explaining why in a comment.

## Two gates, so deploying is inert

Nothing reaches GitHub unless **both** hold: the account has an active row in `account_github_repo`, *and* the case has a `github_issue_number`. Both start empty, so deploying does nothing regardless of how many cases exist. That is what made it safe to apply against staging with 8,092 cases and 3,275 change requests on it.

## Verified end to end

Driven from real staging data against a real repository, through the real GitHub API, with the workflows running:

```
trigger → queue → dispatch (204) → workflow → comment on the issue
```

`scripts/github-integration-e2e.sh` covers both halves in 20 checks and restores every row it touches in a trap, including on failure.

## Defects found by running it, not by reading it

1. **The outbound worker could not drain the queue at all** — `ClaimDue` selected `q.git_reference`, a column the migration does not have. Every poll failed with SQLSTATE 42703. The outbound tests use a fake repository, so that SQL had never run.
2. **Replay protection was never wired in** — `ClaimDelivery`, `ReleaseDelivery` and `LinkDelivery` were implemented, on the interface, faked in tests, and never called. The handler's `ErrDeliverySeen` branch was unreachable.
3. **`client_payload` exceeded GitHub's ten-property limit** — every `servicenow-cr-update` returned `422: No more than 10 properties are allowed; 11 were supplied`.
4. **A permanent failure could not be recorded** — `Reschedule` used `$2` as both an integer and an untyped comparison operand, so Postgres rejected it with 42P08 and the row stayed `PENDING` forever.
5. **Planned-date changes never fired** — the trigger watched `planned_start_date`/`planned_end_date`, which are not columns on `change_request`; they are `start_on`/`end_on`. `newv -> col` returns NULL for a missing key rather than erroring, so both sides always compared equal. ServiceNow watched those fields, so a quarter of its CR updates were silently absent.
6. **A foreign key aborted the caller's transaction** — the queue referenced `change_request(id)`, so closing a case failed *inside the trigger*, taking the portal's UPDATE down with it.

An earlier commit on this branch also deleted `trg_github_outbound_assignment` by a bad splice; it is restored, and the validation now asserts the four triggers **exist** rather than that migrations apply.

## Confirmed against the live instance

`github.dispatch.config` is keyed by **account** with a per-account credential, and no repository is shared between accounts — so `account_github_repo` and its `UNIQUE (owner, repository)` match the real data.

## Not a gap in this PR, but a cutover blocker next to it

Filing an issue from a case already works: `CreateGithubIssueDialog` in the portal
calls `POST /cases/{id}/github-issues` on csm-portal-backend, which calls the same
route on entity-service (`routes.go:601`), which proxies to ServiceNow.

There is exactly one implementation of `CaseGithubIssueService` —
`NewServiceNowCaseGithubIssueService` — and it posts to the ServiceNow integration
service. **There is no Postgres implementation**, so at cutover this endpoint has
nothing behind it and cases stop getting linked. The sync in this PR then has
nothing to sync, because `case.github_issue_number` is gate 2.

That needs its own PR before cutover: a Postgres-backed `CaseGithubIssueService`
that creates the issue via the GitHub API and writes `github_issue_number`. Note
`repoOverride` in the existing contract — an agent can file to a repository other
than the account's default, so the case will also need to store owner/repository
rather than deriving them from `account_github_repo`.

## Repo-side cutover work

The workflows in each mapped repository need two changes, because they are ServiceNow-coupled:
- link paths (`/now/cwf/agent/record/...`) become CSM portal routes
- `sn_cr_notifier.yml`'s ServiceNow Table API call must be removed or repointed

Both were done and proven on a scratch repository before this PR.

## Unrelated, but worth knowing

- `Github Integration API`'s comment webhook in ServiceNow has an **inverted signature check** — valid requests get 401, unsigned requests are accepted and processed. Live today.
- The Postgres `case_service.go` publishes **no events**; all seven `case.*` publishes live in `sn_case_service.go`. Anything consuming those goes silent at cutover.
- `TestSNCaseService_CreateCase_PublishesCaseCreated` fails on `dev-app-csm-portal` itself, untouched by this branch.

## Deployment

`docs/github-integration-deployment.md`. Two things that bite: `GITHUB_TOKEN` needs **Contents: write** (dispatch is gated on that, not Issues), and the gateway must forward the webhook body byte-for-byte or every delivery 401s.

## Summary by CodeRabbit

* **New Features**
  * Added GitHub integration for synchronizing labeled issues with change requests.
  * Added a signed GitHub webhook endpoint supporting issue and comment updates.
  * Added outbound updates for change-request creation, status changes, assignments, comments, and case closures.
  * Added configurable GitHub Enterprise support, label mappings, retry handling, and duplicate-delivery protection.
  * Added repository routing and GitHub-linked change-request numbering.
* **Documentation**
  * Added deployment guidance and end-to-end integration testing instructions.


cloby99 (dismissed):

cloby99 (approved):

SasmithaDilshan:
Tracked retrospectively as wso2-enterprise/digiops-cs#3137, under wso2-enterprise/digiops-cs#3136 (epic wso2-enterprise/digiops-cs#2365).

PR #1834, opened 2026-09-18
[CSM-PORTAL-WEBAPP]: live send progress + failed-project visibility for announcement create

DaniduWijendra:
## Summary
- Both announcement create flows ("All customer projects" / "Specific projects" and the EOL/product-version flow) fan out one real `POST /cases` call per target project. For a large audience, the button previously just said "Creating…" with no visibility into progress until the whole batch finished.
- Adds a live progress card (`AnnouncementSendProgress`) showing "N/total" plus succeeded/failed counts, updating in real time as each project's own create call settles (via a new optional `onSettle` callback on `settleWithConcurrencyLimit`).
- A failed project is shown as a chip labeled by its **short key** (e.g. `CUPPTSUB`), not its raw id — a bare count or a UUID doesn't tell the sender which customer project actually needs a retry.
- If at least one case-create call fails (partial or total), the form now **stays on the page** instead of auto-navigating back to the list, so the failed-project chips stay visible for the sender to act on rather than only living in a toast that disappears the moment the page changes. A tag-attach-only failure (case created fine, the optional "Security announcement" label failed to attach) still navigates away, since the announcement itself did go out in that case.
- Getting a project's key for the "Specific projects" picker required a small additive change to the shared `AsyncProjectMultiSelect` component (a new optional `onSelectedProjectsChange` callback, backward compatible with its other existing call sites).

## Test plan
- [x] `tsc -b` clean
- [x] `eslint` clean (only 2 pre-existing warnings in unrelated files)
- [x] Full `csm-announcements` + `AsyncProjectMultiSelect` consumer test suites pass (109 tests), including new tests asserting:
  - the progress card updates live as each project's create call settles (staggered/deferred promises)
  - a failed project's id is visible immediately, before the batch finishes
  - the page stays on screen (no navigate) after a partial or total create failure, for both flows
  - a failed project shows its short key, never its raw id, both in the progress card and the error banner
- [x] Manually verified against the real staging backend (`localhost:3001` dev server pointed at staging) — confirmed a project shown as "succeeded" in the progress card genuinely received the announcement (visible on the support portal), and a project shown as "failed" (using a temporary local-only test hook, removed before this PR) correctly shows its key and keeps the page open for retry.

## Summary by CodeRabbit

- **New Features**
  - Added live progress tracking for announcement sends, including completion counts, success/failure totals, progress bars, and affected project labels.
  - Project short keys now appear in progress details and error messages.
  - Added targeted retries that resend announcements only to failed projects.

- **Bug Fixes**
  - Partial failures keep forms open for review and retry.
  - Announcement details remain locked during retries.
  - Improved failure messaging and mandatory exclusion controls.
  - Cancel is disabled while sending, and stale progress clears when the audience changes.

- **Tests**
  - Added coverage for progress, failures, labels, retries, and settlement behavior.


DaniduWijendra:
@coderabbitai review

DaniduWijendra:
@coderabbitai review

DaniduWijendra:
Superseded by #1874 (merged), which reconciled the announcement draft/approval work against dev-app-csm-portal. This PR's own commits target the create forms' immediate-send flow, which no longer exists — #1852/#1853 (also reconciled into #1874) replaced it with a save-a-draft/submit-for-approval design, moving the actual fan-out send into the pending-request dialog's Publish action instead.

What was reused from this PR (adapted to the new location, not applied as-is) landed in #1874:
- The mandatory project-exclusion checkbox styling in AudienceScopeControls.
- The `AnnouncementSendProgress` card (succeeded/failed tally, failed-project chips, correct "sent with failures" wording), wired into the dialog's Publish flow instead of the create forms.

Not carried over: failed-project chips in the dialog are labeled by raw project id, not a resolved short key like this PR's original ("CUPPTSUB") — the dialog's frozen `resolvedProjectIds` snapshot has no key data at all, so that needs a new id→key batch-lookup path that doesn't exist yet in this flow. Tracked as a follow-up, not yet its own PR.

Closing this one since its own diff can't be applied to the current code.

PR #1836, opened 2026-09-19
[CSM Portal] Post an internal work note on every fix-ETA update

rksk:
## Purpose
A CSM portal "Set fix ETA" update (`PATCH /cases/{id}` with any of `bestCaseFixEta`/`mostLikelyFixEta`/`worstCaseFixEta`) silently updated ServiceNow with no trace in the case's own comment/activity history, unless the caller also opted into `addPublicComment` (a separate, customer-visible comment posted entirely inside ServiceNow's own scoped-app script). This left CS engineers unable to tell from a case's activity feed that an internal fix-ETA update had happened at all, or when.

## Goals
Every fix-ETA update posts an internal `work_note` comment recording the committed date(s), independent of whether the customer-visible comment path is also used.

## Approach
`snCaseService.UpdateCase` (`entity-service/internal/service/sn_case_service.go`) now calls a new `postFixEtaWorkNote` helper right after the fix-ETA fields are echoed back from ServiceNow's PATCH response. It reuses the existing `CreateCaseComment` method with `Type: CommentTypeWorkNote`, and builds its content from the SN-echoed committed values (not the raw request), so the note reflects what was actually stored. Only the fields that were part of the PATCH are mentioned. A failure to post the note is logged and swallowed — it does not fail the `UpdateCase` call, since the fix-ETA fields have already updated successfully in ServiceNow by that point.

No change to any ServiceNow scoped-app script, the Ballerina layer, or the CSM webapp — this is a Go-only addition to the entity-service, reusing an already-existing comment-creation path.

## User stories
As a CS engineer, when I or a teammate updates a case's fix-ETA estimates from the CSM portal, I want to see that change reflected in the case's own history, not just in the current field values.

## Release note
Fix-ETA updates from the CSM portal now post an internal work note recording the committed estimate(s), so the update is visible in the case's activity history even when it isn't shared with the customer.

## Documentation
N/A — internal-only behavior change, no new user-facing UI or field.

## Training
N/A

## Certification
N/A — no impact on certification exams.

## Marketing
N/A

## Automation tests
- Unit tests
  Added `TestSNCaseService_UpdateCase_PostsFixEtaWorkNote` in `sn_case_service_test.go`, covering: a work note is posted after a fix-ETA PATCH, it targets the correct case as an internal (`work_notes`) comment, its content reflects the SN-echoed committed values, and it omits any field not part of that PATCH.
- Integration tests
  N/A — no integration test harness for this service beyond the existing mocked-SN-client unit tests.

## Security checks
- Followed secure coding standards in http://wso2.com/technical-reports/wso2-secure-engineering-guidelines: yes
- Ran FindSecurityBugs plugin and verified report: N/A (Go service; no FindSecurityBugs tooling in this module)
- Confirmed that this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets: yes

## Samples
N/A

## Related PRs
None

## Migrations (if applicable)
N/A

## Test environment
`go build ./...`, `go vet ./...`, `go test ./...` — all passing locally (Go toolchain per `go.mod`).

## Learning
N/A


rksk:
Moving this to the CSM BE (apps/csm-portal/backend), which is common to both data sources (ServiceNow and Postgres), rather than entity-service's SN-specific code path. Superseded by a follow-up PR.

PR #1839, opened 2026-09-19
[Entity-Service]: add announcement_requests entity (Phase 2 draft/approval)

DaniduWijendra:
## Summary

First PR of Phase 2 (draft / dry-run-gated approval / publish) for the CSM portal announcement feature. Adds a new Postgres-only `announcement_requests` entity — no ServiceNow equivalent, same architectural shape as `sla_clocks`/`scheduled_task_run` (migration-gated, registered only when a DB pool exists).

- **State machine**: `draft → pending_approval → approved → published`, enforced entirely in the Go service layer (no generic state-machine abstraction exists in this codebase — mirrors `case_service.go`'s own ad hoc guard convention).
- **The existing "Dry run" mechanism is the approval-review artifact** — no new preview-rendering surface. `Submit` is rejected unless a dry run has already been recorded for the request (`dryRunCaseId` set).
- **No approver role or email-sending logic** — approval stays a human process over email, entirely outside this service. `Approve` just records that whoever's working the request says it's been approved.
- **Audience is frozen at `Submit`, not `Approve`** — the resolved project-id snapshot describes exactly what the dry-run link and the approval email describe; `Publish` sends to that frozen list, never a live re-resolution.
- **Editing behaves differently per current state**: plain edit while `draft`; edit-and-revert-to-`draft` while `pending_approval` (the content is out for real review, so a silent change isn't safe); edit-in-place with no state/audience change while `approved` (a human already said yes over email — a deliberate, explicitly-accepted trade-off).
- **Every state-transition write is atomically conditioned on the expected state inside the `UPDATE`'s own `WHERE` clause**, not just checked beforehand in the service layer — closes a real race where two concurrent transitions on the same row could otherwise have the slower one silently clobber the faster one's result. Caught this via a deliberate self-review pass, not by an external reviewer.
- **No persisted per-project delivery ledger** — publishing still uses the existing ad hoc fan-out (`POST /cases` per project) unchanged; this entity only gates *when* that's allowed to run and records who/when. A durable batch/delivery ledger is explicitly Phase 3, not this.

## Not in this PR

- csm-portal-backend thin forwarders (next PR)
- webapp UI changes (subsequent PRs) — create forms, pending-list dialog, registry tab

## Test plan

- [x] `go build ./...`, `go vet ./...` clean
- [x] `go test ./...` — full suite passes, including the `TestEveryRegisteredRouteIsPublished` OpenAPI-contract test
- [x] `gosec -fmt=text ./...` — 0 issues
[...]

PR #1840, opened 2026-09-19
[CSM-PORTAL-BACKEND]: add announcement_requests forwarders (Phase 2)

DaniduWijendra:
## Summary

Second PR of Phase 2 (draft / dry-run-gated approval / publish). Depends on entity-service's new `announcement_requests` endpoints ([#1839](https://github.com/wso2-open-operations/cs-tools/pull/1839), not yet merged) — this PR's real runtime behavior needs that one merged and deployed first, though it builds and tests cleanly against a mocked entity client on its own.

- **Thin forwarders** for create/get/search/update/dry-run/approve/publish, mirroring the existing `TaskSlaHandler` passthrough convention.
- **`createdBy`/`actorId` are always the authenticated caller** — never a client-supplied value. `Approve`/`Publish` don't even read a request body, since the only input those transitions need is who's performing them.
- **`Submit` is the one endpoint with real logic**: it loads the request's own stored `kind`/`audienceDefinition`, then resolves the authoritative project-id snapshot server-side — paging through the same entity-service search endpoints a live audience resolution already uses, with the mandatory excluded-project-key denylist injected via the existing `injectExcludeProjectKeys` for the customer "all projects" scope. Bounded the same way the webapp's own paged-resolution hooks are (fails loudly past 200 pages rather than silently truncating an over-large audience into the frozen snapshot).
- **`UpdateAnnouncementRequest` uses `mapUpstreamError`** (not the Generic variant), per this backend's own PATCH convention — the payload just submitted is what's potentially being rejected (e.g. an audience change on an approved request).

## A real bug caught by my own tests, not by review

Using `json:"-"` on the actor-id field to keep a client-supplied value from being read on the way *in* also silently dropped the real, server-set value on the way back *out* when re-marshaling the request for the upstream call — that tag suppresses both directions, not just unmarshal. My first pass of `CreateAnnouncementRequest`/`UpdateAnnouncementRequest`/`RecordAnnouncementRequestDryRun` all had this bug; every one of them was sending an empty `createdBy`/`actorId` upstream despite "working" in the sense of returning 200/201. The three tests asserting the actual authenticated value made it into the forwarded body caught it immediately. Fixed by using a separate inbound (no actor field at all) / outbound (actor field with the real value) struct pair instead of one struct trying to do both jobs.

## Test plan

- [x] `go build ./...`, `go vet ./...` clean
- [x] `go test ./...` — full suite passes
- [x] `gosec -fmt=text ./...` — 4 pre-existing findings in `internal/dashboard`/`internal/sftpgo`, confirmed unrelated to this change (untouched files, same findings present on `main`)
- [x] `govulncheck ./...` — no vulnerabilities
- [x] New `announcement_requests_test.go`: auth required on every endpoint; actor-id/createdBy forced to the authenticated caller across Create/Update/RecordDryRun (the exact case that caught the bug above); Approve/Publish take no body at all; Submit's three audience-resolution paths (customer specific/all, eol) including multi-page accumulation, the mandatory-exclusion injection on every page, an empty-audience rejection, an unknown-kind rejection, the pagination safety-bound actually firing instead of looping forever, and the full upstream-error-mapping matrix
- [ ] End-to-end verification blocked on #1839 being deployed first — the actual entity-service endpoints this forwards to don't exist yet on any deployed environment.

## Summary by CodeRabbit

- **New Features**
  - Added an announcement request workflow with draft, approval, and publishing stages.
  - Added capabilities to create, search, view, edit, dry-run, submit, approve, and publish announcement requests.
  - Added automatic audience resolution for customer-specific, all-customer, and end-of-life announcements.
  - Added API documentation for the new endpoints, request payloads, and responses.

- **Tests**
  - Added comprehensive coverage for authentication, validation, workflow transitions, audience resolution, and error handling.


DaniduWijendra:
@coderabbitai review

DaniduWijendra:
Superseded by #1881 (merged), which reconciled this PR's content against dev-app-csm-portal — closing in favor of that.

